PR #24, opened 2026-07-28
graft init: pick what gets written instead of every detected agent

anirudhkumar-nanonets:
## The problem

Running `graft init` once in a repo produced ten files nobody asked for:

```
?? .cursor/rules/graft.mdc      ?? AGENTS.md
?? .cursor/mcp.json             ?? GEMINI.md
?? .gemini/settings.json        ?? opencode.json
?? .github/copilot-instructions.md
?? .mcp.json
 M .claude/helpers/graft-hooks.cjs
 M .claude/helpers/graft-statusline.cjs
```

This wasn't a missing feature — `--agents`, `--all-agents`, `--no-agents`, `--no-mcp`
and `--no-hooks` all already existed. The problem was the **default**: with no
`--agents`, `runHostsInit` fell back to `detectHosts`, and detection keys off
directories in `$HOME` (`~/.cursor`, `~/.gemini`, `~/.codex`, `~/.config/opencode`).
Try three coding CLIs once and plain `graft init` behaves exactly like `--all-agents`,
in every repo, forever after.

### The part that doesn't show up in `git status`

Selecting the `agents` host also wrote three **user-level** files:

| Path | What it did |
|---|---|
| `~/.codex/config.toml` | registered `[mcp_servers.graft]` |
| `~/.codex/hooks/graft/graft-hooks.cjs` | post-edit shim, 0755 |
| `~/.codex/hooks.json` | `PostToolUse` matching `Write\|Edit\|MultiEdit` |

A one-repo command was arming a hook that fires on every edit in every project on the
machine, and loading an MCP server into every future Codex session. `--no-hooks`
suppressed two of them; **nothing** suppressed the `config.toml` write.

## The change

`graft init` now asks. On a terminal it lists every known agent, flags which ones were
detected, shows the exact files each would write, and wires only what you select
(Claude Code pre-selected):

```
graft init — select what to wire into this repo:

› [x] claude                   .claude/settings.json, .claude/helpers/graft-statusline.cjs, … +2 more
  [ ] agents                   AGENTS.md · + 3 in ~/.codex/ (machine-wide)
  [ ] adal                     .adal/skills/graft/SKILL.md · no MCP
  [ ] cursor                   .cursor/rules/graft.mdc, .cursor/mcp.json
  [ ] gemini                   GEMINI.md, .gemini/settings.json
  [ ] copilot (not detected)   .github/copilot-instructions.md · no MCP
  [ ] kiro (not detected)      .kiro/steering/graft.md, .kiro/settings/mcp.json
  [ ] windsurf (not detected)  .windsurf/rules/graft.md · no MCP

↑↓ move · space toggle · a all · enter confirm · esc cancel
```

Undetected hosts are listed too, dimmed, so they're discoverable without knowing
`--list-agents` exists.

### New flags

| Flag | Effect |
|---|---|
| `--dry-run` | print every path `init` would touch — repo section, then a separate machine-wide section — and exit without writing |
| `--yes`, `-y` | skip the prompt and wire every detected agent (the pre-0.8 default, now opt-in) |
| `--no-global` | skip every write outside the repo; still wires `AGENTS.md` |

`--dry-run` on this repo reproduces the original complaint before writing a byte:

```
would write — this repo:
  .claude/settings.json    graft statusline + hook blocks
  .mcp.json                mcpServers.graft
  AGENTS.md                fenced graft section
  opencode.json            mcp.graft
  .cursor/rules/graft.mdc  graft-owned file
  …11 repo writes total

would write — your machine, affects ALL repos:
  ~/.codex/config.toml                  [mcp_servers.graft]
  ~/.codex/hooks/graft/graft-hooks.cjs  post-edit hook shim
  ~/.codex/hooks.json                   PostToolUse: Write|Edit|MultiEdit

suppress the out-of-repo writes with --no-global
nothing was written (--dry-run)
```

## ⚠️ Breaking: non-interactive callers

`graft init` in CI, a Dockerfile, or any piped shell now writes **nothing** and prints
the command to run instead. That's the intended fix, but it needs a migration:

```bash
graft init --yes            # wire every detected agent (pre-0.8 behaviour)
graft init --agents claude  # or name them explicitly
```

`scripts/postinstall.mjs` only prints a nudge, so `npm install` is unaffected.

## Why the diff touches 12 files

Only **two** files change behaviour. A picker and a `--dry-run` both have to answer
*"what would you write?"* **before** anything is written, and path knowledge was
tangled inside the functions that do the writing — each writer computed its own
`join(...)` at the moment it wrote, so there was nothing to ask.

| Layer | Files | Role |
|---|---|---|
| Describe | `src/hosts/plan.ts` | **new.** `PlannedWrite` (path + `scope` + what changes) and `planInit()`, which composes the whole plan without touching disk |
| Write | `mcp-config.ts`, `codex-hooks.ts`, `claude/init.ts`, `hosts/init.ts` | each split into a pure `*Targets()` half and a writing half that consumes it. **Output identical** |
| Decide | `src/cli-picker.ts`, `src/cli.ts` | the picker, and the flag → selection → write resolution order |
| Prove & tell | 3 test files, `README.md`, `CHANGELOG.md` | +38 tests; docs, including the `~/.codex/` writes which had never been documented |

The one new idea is the field the old code never had:

```ts
export type WriteScope = 'repo' | 'global';   // 'global' = outside the repo
```

Every user-facing improvement here is downstream of naming it — the amber
`machine-wide` label, the separate `--dry-run` section, and `--no-global`, which is a
single `.filter()` and so covers any future out-of-repo target automatically.

### The test that keeps it honest

A plan that silently disagrees with the writes is worse than no plan — it would
confidently show the wrong list. So the two are pinned to each other in both
directions:

```ts
test('every path a real runHostsInit writes was in the plan, and vice versa', () => {
  const planned = new Set(selectedWrites(planInit(repo, { home }), ids).map(w => w.path));
  const r = runHostsInit(repo, { home, agents: ids });
  const actual = new Set([...r.written.map(w => w.path),
                          ...r.mcp.map(m => m.path),
                          ...r.hooks.map(h => h.path)]);
  assert.deepEqual([...actual].sort(), [...planned].sort());
});
```

Add a write without adding it to the plan and this fails; add it to the plan without
wiring the write and it fails too.

## Notes on scope

- **`runHostsInit`'s detection fallback stays.** Calling it with no `agents` still
  writes for detected hosts — it's a tested library entry point. The CLI simply no
  longer uses that path; it always passes an explicit list. Only the CLI boundary's
  default moved.
- **No `package.json` bump.** The `## 0.8.0` CHANGELOG section lands with the feature;
  the version bump goes in a separate `release:` commit, matching how 0.7.0 shipped.
- The picker is `node:readline` + `setRawMode` — no new dependency.

## Testing

`npx tsc --noEmit` clean; **436 pass / 0 fail** (398 on `main`).

- `test/hosts-plan.test.ts` (new, 9) — the anti-drift assertion above; exactly three
  `global` writes under `~/.codex`; they vanish when Codex isn't installed while
  `AGENTS.md` is still planned; all five Claude writes are repo-scoped.
- `test/cli-picker.test.ts` (new, 23) — `renderPicker`/`reducePicker` are pure, so the
  whole TUI is tested with no TTY. Includes a regression for a real hang: a terminal
  batches bytes, so `down space enter` arrived as one 5-byte chunk and matched nothing.
  `keysOf()` walks the chunk and consumes CSI sequences whole, which also stops every
  `↑` from reading as a bare ESC (abort).
- `test/hosts-init.test.ts` (+6) — end-to-end runs of the real CLI with `HOME` pointed
  at a scratch dir, so the suite asserts on `~/.codex` without touching yours. One
  asserts it stays empty under `--no-global`.

Verified by hand: non-TTY writes nothing · `--dry-run` lists 11 repo + 3 machine-wide
and writes nothing · `--yes` reproduces pre-0.8 behaviour · `--no-global` writes only
`AGENTS.md` and leaves `~/.codex` empty · the picker driven on a real pty wired
claude + agents while leaving detected-but-unselected Cursor alone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,36 @@
 # Changelog
 
+## 0.8.0
+
+### Changed
+
+- **`graft init` now asks which agents to wire, instead of writing files for every
+  agent it detects.** Detection keyed off directories in `$HOME`, so anyone who had
+  tried several coding CLIs got instruction files and MCP configs for all of them —
+  plain `graft init` effectively behaved like `--all-agents`. On a terminal it now
+  shows every known agent, which ones were detected, and the exact files each would
+  write, and wires only what you select (Claude Code pre-selected).
+
+  **Migration —** `graft init` in CI, a Dockerfile, or any non-interactive shell now
+  writes **nothing** and prints the command to run instead. Add `--yes` for the old
+  behaviour, or `--agents <ids>` to be explicit:
+
+  ```bash
+  graft init --yes                  # wire every detected agent (pre-0.8 default)
+  graft init --agents claude        # or name them
+  ```
+
+### Added
+
+- **`graft init --dry-run`** — print every path `init` would touch, then exit without
+  writing. Out-of-repo writes get their own section.
+- **`graft init --no-global`** — skip every write outside the repo. Selecting the
+  `agents` host writes to `~/.codex/config.toml`, `~/.codex/hooks.json`, and
+  `~/.codex/hooks/graft/`; those are user-level and apply to every repo you open with
+  Codex, and previously nothing suppressed the `config.toml` write (`--no-hooks` only
+  covered the other two). These are now labelled `machine-wide` in the picker.
+- **`graft init --yes`** — wire every detected agent without prompting.
+
 ## 0.7.0
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -61,7 +61,9 @@ npm install -g @nanonets/graft   # install the CLI, once
 graft init                       # build the graph + wire it into Claude Code
 ```
 
-That is the whole setup. `graft init` builds `graft/` from your code and drops a statusline and hooks into `.claude/`, so from the next session on Graft rides along in Claude Code: it pulls the matching nodes into each prompt and rebuilds the graph in the background after every turn. No daemon, no re-indexing to remember, nothing to run or maintain by default — the graph is just files.
+That is the whole setup. `graft init` asks which of your coding agents to wire up, builds `graft/` from your code, and drops a statusline and hooks into `.claude/`, so from the next session on Graft rides along in Claude Code: it pulls the matching nodes into each prompt and rebuilds the graph in the background after every turn. No daemon, no re-indexing to remember, nothing to run or maintain by default — the graph is just files.
+
+Nothing is written until you pick. Run `graft init --dry-run` to see every file it would touch first, or `graft init --agents claude` to skip the prompt and wire Claude Code alone.
 
 `graft build` adds `graft/` to your `.gitignore` automatically — the graph is a local, regenerable cache (like `node_modules`), not something you commit. What you share is the wiring `init` dropped into `.claude/`; each teammate runs `graft build` to generate their own graph:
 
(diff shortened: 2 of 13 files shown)


anirudhkumar-nanonets:
## Self-review follow-up — 4 fixes in ce6e125

Reviewed the branch and found one real regression this PR introduced, plus three smaller things.

**1. The graph build was gated on Claude Code being selected** *(the regression)*

```ts
if (wantClaude) {
  const res = runInit(repo, { build: opts.build, cliPath });   // ← the only build call
```

On `main` this was reachable only by explicitly typing `--agents cursor`. With the picker it's **one spacebar press on the default path** — so a Cursor-only user unchecking `claude` got `.cursor/rules/graft.mdc` telling Cursor to consult a graph that was never built:

```console
$ graft init "$R" --agents cursor    # before
$ ls -a "$R"
.  ..  .cursor                       # no graft/
```

Extracted `buildGraphIfMissing()` out of `runInit` and run it on both paths. Verified with the built CLI:

```console
$ graft init "$R" --agents cursor    # after
✓ cursor: …/.cursor/rules/graft.mdc (created)
✓ built the graph (graft build)
```

**2. Unused readline interface fighting raw mode** — `createInterface({ input: stdin })` was never used except `rl.close()`. It defaults `terminal` to `input.isTTY`, so on a real terminal readline takes over raw mode alongside the picker's own handler, and the `finally` restored raw mode *before* calling `rl.close()`. Deleted both lines.

**3. Leftover `end` listener** — `finish()` removed the `data` listener but not `once('end')`, so after a confirmed pick a later stdin close would flip `aborted` and repaint the picker over init's `✓ wrote …` output. Both listeners now come off together.

**4. `--no-global` claimed credit for nothing** — `--agents cursor --no-global` printed `· skipped out-of-repo writes` although cursor has no out-of-repo target. Now gated on the plan actually containing a `global` write.

### Tests

**441 pass / 0 fail** (was 436), `tsc --noEmit` clean. +5 tests:

- `buildGraphIfMissing` guards: `build: false`, missing `cliPath`, existing `wiring.json`.
- CLI regression for #1 — a non-claude selection must still reach the build decision. Fails on the old code, where neither build message was printed at all.
- CLI regression for #4 — `--no-global` stays quiet for a repo-only selection.

Re-verified by hand: the picker driven on a real pty still wires claude + agents and leaves detected-but-unselected Cursor alone; `--no-global` still leaves `~/.codex` empty.

PR #32, opened 2026-07-30
Seed a git worktree's graph from its parent checkout

anirudhkumar-nanonets:
`graft/` is gitignored, so `git worktree add` never checks it out — and the graph is the only thing the MCP tools read. In a fresh worktree every tool answered `no matching nodes` / `no graph found` for the whole session, and `INDEX.md` and the cards were missing too, so `grep` and the repo map came up empty.

A query in a worktree now copies the parent checkout's `graft/` in, then treats the difference between the two checkouts as ordinary drift. The worktree's `.git` is a file naming its parent, so there is nothing to configure.

## Why copy rather than rebuild cold

- **The Tier-2 meaning layer survives.** `buildGraph` folds a prior `wiring.json` in by `body_hash`, so unchanged code keeps summaries that cost money. A cold rebuild discards all of them.
- **The repair is small.** `probeDrift` confirms suspect files by content hash, so a fresh checkout's new mtimes cost one hashing pass; the drift that survives is exactly the diff between the two checkouts.
- **$0 and offline**, like everything else on the query path.

## Scope

- `src/graph/seed.ts` — `mainWorktreeRoot()` (filesystem-only; rejects submodules, bare repos and `--separate-git-dir`) and `seedGraph()` (copy, with `wiring.json` installed last via rename so a half-copy leaves no graph).
- `src/graph/refresh.ts` — seeds at the no-graph branch under the existing sync lock, then falls through to the normal drift repair.
- `src/graph/build.ts` — seeds before reading its priors, so `graft build` in a worktree is incremental too; reports `seededFrom`.
- `src/cli.ts` — one line when a build started from a copied graph.

Reads the parent, never writes to it. No-ops unless a built parent checkout is genuinely on disk, so a fresh clone, CI, or a cloned (rather than worktree'd) cloud session behaves exactly as before. `GRAFT_NO_SEED=1` turns it off.

## Verification

- 11 new tests in `test/graph-seed.test.ts` drive real `git init` / `git worktree add` rather than hand-written `.git` files, and cover both layouts (temp-dir and repo-nested `.claude/worktrees/`), the submodule/bare/garbage rejections, the off switch, and that the parent's graph is byte-identical afterwards. Full suite 503/503.
- Checked live in a Claude Code worktree-isolated agent (`.claude/worktrees/agent-…`, a real linked worktree): `graft/` absent beforehand; after one query, `[graft] refreshed the graph — copied the graph from the main checkout (…)`, 920 nodes, 30 cards, `INDEX.md`, parent untouched.

## Known gap, not addressed here

The MCP server takes its root from the cwd it was launched with (`src/cli.ts` `startMcpServer`), so a subagent that relocates into a worktree mid-session keeps querying the parent's graph — it answers with the parent's line numbers instead of erroring. Separate root-resolution issue; this PR only fixes the missing-cache one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,24 @@
 - **`GRAFT_REFRESH=hash`** — confirm every file by hashing its contents instead of trusting
   size and mtime, for tooling that rewrites files while preserving both.
 
+### Fixed
+
+- **A git worktree is no longer blind.** `graft/` is gitignored, so `git worktree add`
+  never checks it out — and the graph is the only thing the MCP tools read. Every tool in
+  a fresh worktree answered `no matching nodes` / `no graph found` for the whole session,
+  and `INDEX.md` and the cards were missing too, so `grep` and the repo map came up empty.
+
+  A query in a worktree now copies the parent checkout's `graft/` in and then treats the
+  difference between the two checkouts as ordinary drift. The worktree's `.git` is a file
+  naming its parent, so there is nothing to configure; the copy is $0 and offline, and the
+  Tier-2 meaning layer survives it (a cold rebuild would have thrown away every summary
+  you paid for and re-parsed the repo). `graft build` in a worktree starts from the same
+  copy, so it is incremental too.
+
+  Reads the parent, never writes to it. No-ops unless there is genuinely a built parent
+  checkout on disk — a fresh clone, CI, or a cloned (rather than worktree'd) cloud session
+  behaves exactly as before. `GRAFT_NO_SEED=1` turns it off.
+
 ## 0.8.0
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,12 +56,14 @@
   a fresh worktree answered `no matching nodes` / `no graph found` for the whole session,
   and `INDEX.md` and the cards were missing too, so `grep` and the repo map came up empty.
 
-  A query in a worktree now copies the parent checkout's `graft/` in and then treats the
-  difference between the two checkouts as ordinary drift. The worktree's `.git` is a file
-  naming its parent, so there is nothing to configure; the copy is $0 and offline, and the
-  Tier-2 meaning layer survives it (a cold rebuild would have thrown away every summary
-  you paid for and re-parsed the repo). `graft build` in a worktree starts from the same
-  copy, so it is incremental too.
+  A query in a worktree now copies the parent checkout's graph and query sidecars in, then
+  treats the difference between the two checkouts as ordinary drift. The worktree's `.git`
+  is a file naming its parent, so there is nothing to configure; the copy is $0 and
+  offline, and the Tier-2 meaning layer survives it (a cold rebuild would have thrown away
+  every summary you paid for and re-parsed the repo). `graft build` in a worktree starts
+  from the same copy, so it is incremental too — and it is what writes the worktree's
+  cards and `INDEX.md`, generated from *this* checkout's code rather than copied from the
+  parent's branch. A query still writes only what a query reads.
 
   Reads the parent, never writes to it. No-ops unless there is genuinely a built parent
   checkout on disk — a fresh clone, CI, or a cloned (rather than worktree'd) cloud session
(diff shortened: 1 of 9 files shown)


anirudhkumar-nanonets:
### Code review

Found 2 issues:

1. Seeding copies the markdown passive surface on the query path, and the follow-up repair never refreshes it. `SKIP` omits `INDEX.md` and the cards, so `copyTree` brings them across; the seed call in `buildGraph` is gated only on `opts.reuse`, not on `opts.graphOnly`, and `writeCards`/`writeIndex` stay behind `if (!opts.graphOnly)`. Net effect: a query in a worktree writes cards that describe the *parent's* branch and nothing rewrites them until an explicit `graft build`, so `grep` and the repo map can answer from the wrong checkout with no staleness signal. This contradicts the invariant stated in the same file being modified — `src/graph/refresh.ts` header: "Writes only what a query reads (`graphOnly`): the graph, the `ask` sidecar, the freshness record. Not the markdown cards, not `INDEX.md`, not `.gitignore`." Either scope the query-path copy to the graph plus its sidecars and leave the markdown to the `Stop` hook's build, or document this as a deliberate exception to that rule.

https://github.com/NanoNets/Graft/blob/428d80b5a2302b4c8972ab5e538c738469e1b770/src/graph/seed.ts#L107-L121

https://github.com/NanoNets/Graft/blob/428d80b5a2302b4c8972ab5e538c738469e1b770/src/graph/build.ts#L151-L154

2. `/^gitdir:[ \t]*(.+?)[ \t]*$/m` is the polynomial-ReDoS shape this repo just removed elsewhere: `.` also matches space and tab, so `(.+?)` and `[ \t]*$` overlap ambiguously. CodeQL flagged the same class on `src/context/node-file.ts` in [#26](https://github.com/NanoNets/Graft/pull/26#discussion_r3675601733), and commit `44a191e` replaced it with a linear-time loop while explicitly rejecting "the input is not attacker-controlled" as a reason to leave the ambiguity in place. `.github/workflows/codeql.yml` runs on PRs, so this will likely be flagged on merge. Match the existing precedent: capture greedily without the trailing-whitespace alternation and trim in code.

https://github.com/NanoNets/Graft/blob/428d80b5a2302b4c8972ab5e538c738469e1b770/src/graph/seed.ts#L74-L76

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anirudhkumar-nanonets:
Both review findings addressed in fd2a376, plus the sub-threshold one.

1. **Query path no longer writes markdown.** `SKIP` (a blocklist) is now `seedable` (an allowlist): the copy brings `.graph/wiring.json`, the ask sidecar, the fingerprint and the extraction memo, and nothing else. Cards and `INDEX.md` never travel — they would be the parent's branch, and the query path would never correct them. An explicit `graft build` generates them from this checkout's own graph, and the seeded memo makes that build free: verified in a real worktree, `parsed: 0 of 127 files (127 replayed from cache)`.

2. **No regex on the `.git` pointer.** Replaced with `split`/`startsWith`/`trim`, mirroring `stripTrailingSlashes` from 44a191e. New test covers CRLF, tabs, padded and empty pointers, and extra lines in the file.

3. Also fixed, though it scored below the reporting threshold: losing the seed lock race returned a bare `CLEAN`, i.e. the silent "no matching nodes" this PR removes. `seedUnderLock` now reports `busy` and the gate says so.

Live re-check in a fresh worktree — after one query, `graft/` contains `.cache/` and `.graph/` only, zero `.md`; after `graft build`, `INDEX.md` and the card tree appear. Suite 504/504.

anirudhkumar-nanonets:
## Follow-up, not in this PR

**The MCP server's root is fixed at launch, so a worktree session can answer from the wrong checkout.** `startMcpServer` takes its root from the cwd the server was spawned with, so an agent that relocates into a worktree mid-session keeps querying the parent's graph — plausible-looking hits carrying the parent's line numbers, with no error to signal it. Observed in a Claude Code worktree-isolated agent: the MCP tool cited `ensureFreshGraph` at `L143` (the parent's working tree) while the worktree's own checkout has it at `L117`, which is what the CLI correctly reported.

Separate root-resolution issue; this PR only fixes the missing-cache one.

https://github.com/NanoNets/Graft/blob/fd2a3765c2ca05ccefbc37a6c2d353502720be43/src/cli.ts#L343-L345

PR #41, opened 2026-08-01
fix: unique node ids, UTF-16 read coherence, clearer ambiguous-callers messaging

qoole:
Three engine correctness fixes, independent of each other's motivation but shipped together as they were found together (surfaced while testing new-language extraction at scale; all reproduce on TypeScript/Python today).

## 1. Unique node ids

Same-file same-name redefinitions — branch-guarded defs, Python `@overload` stacks, duplicate methods — mint colliding `path#name` ids. `wiring.json` preserves the duplicates, but every consumer builds last-write-wins `Map`s over them: enrich can attach a stale summary to the wrong symbol, `graft check` drift detection goes quiet on the shadowed copy, callers/impact attribution is arbitrary, and ask-index docs collapse. Real-world scale: psf/requests has 9 colliding groups (20 shadowed nodes) today.

Fix: ids are minted collision-free per file via a small loop-until-free helper (`path#name`, `path#name~2`, … in document order; first occurrence keeps the bare id). A prerequisite refactor carries a method's owner as a `NodeV1.owner` field instead of re-parsing id strings in resolution (behavior-identical, verified). Qualified symbol queries strip ordinals so duplicates stay findable. A standing all-ids-unique test gate prevents regression.

Evidence: semantic before/after diff on psf/requests — 20 duplicate copies → 0, exactly 20 previously-swallowed `contains` edges recovered 1:1, zero edge losses, zero node changes. spf13/afero (no duplicates) — normalized edge multiset EQUAL, proving the change is a no-op where nothing was broken. Cold-vs-incremental builds remain byte-identical on duplicate-bearing fixtures.

## 2. UTF-16 read coherence (hash-what-you-parse)

Repo-source readers each did their own `readFileSync(path, "utf8")` — tier-1 extraction, fingerprint, check, the tier-2 concept pipeline, ask's snippet slicing, and grep. A UTF-16LE source file decodes to NUL-interleaved mojibake, differently consumed at each site. All readers now route through one BOM-aware `readSourceFile` (UTF-16LE decoded; unsupported UTF-16BE yields a clean skip — an empty entry instead of a mojibake read; UTF-8 byte-identical to before), so what is hashed is always what is parsed/summarized/searched.

## 3. Ambiguity in the callers zero-hit note

When several definitions share a queried name, `graft callers` printed "no indexed callers" per match with no hint that ambiguity (cross-file matches are deliberately dropped rather than guessed) is the cause. The note now states the candidate count, at all four surfaces (CLI human + JSON, MCP, workspace federation).

Also included: the dot-dir/skip-dir predicate duplicated across `fs.ts` and three `scopes.ts` sites is consolidated into one shared `shouldSkipDir` (no behavior change).

25 new tests; suite green; `NodeV1.owner` is additive so the graph format version is unchanged, and the extractor stamp self-invalidates caches — no migration needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,32 @@
   matches. Pass a real prefix (`--in server/src/gpu`), a full file path
   (`--in src/a.ts`), or use `grep`'s pattern to match on content.
 
+- **Duplicate-named definitions no longer silently collide onto one graph node
+  id.** A branch-guarded redeclaration, a reopened class, or any other same-name
+  definition within a file used to mint the exact same node id as an earlier
+  definition, so the second one silently overwrote the first in every id-keyed
+  lookup (`callers`, `ask`, MCP tools). Every definition now mints a unique id
+  (`~2`, `~3`, ... on a document-order duplicate), and a qualified query
+  (`Class.method`) now matches every duplicate, not just the first.
+
+- **UTF-16LE source is now decoded consistently everywhere graft reads repo
+  source.** `graft build`'s parse, `check`'s and `fingerprint`'s drift hashes,
+  the context summarizer's input, `ask --source`'s span slicer, and `grep` each
+  read files with their own `readFileSync(file, "utf8")` — hashing what the
+  parser actually sees wasn't guaranteed, and a UTF-16LE file (the common
+  encoding Windows tooling writes) got silently mojibake'd by some readers and
+  not others. All of them now share one `readSourceFile`, so a file decodes
+  identically no matter which command reads it. UTF-16BE, unsupported by
+  Node's built-in decoders, is a clean skip (an empty entry) rather than a
+  mojibake read.
+
+- **`graft callers`'s zero-hit note now says when the query name itself is
+  ambiguous.** When a symbol name is defined more than once, name resolution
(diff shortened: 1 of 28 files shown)


anirudhkumar-nanonets:

qoole:
Rebased onto latest main and re-ran the suite — 563/563 green, tsc clean.

Two notes from the conflict resolution:
- The UTF-16 read-routing composes with the new posix-path work: drift hashing in `check` now keys on `relPosix` paths while reading content through `readSourceFile`, so both fixes apply at every site.
- The `shouldSkipDir` consolidation shrank to fit #51's walker: with scope discovery now derived from the visible file set, the remaining hand-rolled skip checks were `skippedPath`'s segment predicate and `walkFilesystem` (ingest/fs.ts) plus `discoverWorkspaceChildren`'s git-child filter (scopes.ts) — those three now share the one predicate, and the tests in `test/ingest-fs.test.ts` pin that walkDir, scope discovery, workspace-glob resolution, and git-child discovery agree on the same skip set.

qoole:
Thanks for the merge and for #56 — routing the unsupported-encoding case into drift detection is a nice completion of the contract.

PR #42, opened 2026-08-01
feat: --include-dir override for the built-in skip list

qoole:
## What

`graft build --include-dir <name>` (repeatable) removes a directory name from the built-in skip list (`build`, `dist`, `out`, `target`, `vendor`, `coverage`, …) for this repo. The choice is persisted in graft's per-repo state, so every subsequent walk agrees without re-passing flags: extraction, the freshness probe (statusline/hooks), monorepo scope discovery, the tier-2 concept pipeline and its drift check, Go-module discovery, and workspace child builds (the flag propagates to each child's own state). Values are validated: directory names only — dot-directories are not overridable.

## Why

The skip list is ecosystem-shaped (JS build outputs), but some repos keep real, hand-written source in those names — most commonly `build/` holding build *scripts* rather than build *output* (e.g. psake and PSFramework each keep half a dozen live scripts there; graft currently drops them silently). An explicit, persisted opt-in recovers them.

Git-awareness (skip only when the dir is gitignored) was considered and deliberately rejected: graft's freshness probe documents a no-git invariant with a ~3ms budget, and gitignore-conditional walks would make the committed graph depend on the user's global excludes file. The explicit override is deterministic and costs nothing on the hot path.

17 new tests covering the flag, validation, persistence across no-flag rebuilds, freshness-probe coherence, Go-module discovery under an included dir, and workspace child propagation. No behavior change when the flag has never been used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 ## Unreleased
 
+### Added
+
+- **`graft build --include-dir <name>`** — an explicit, persisted override for
+  `SKIP_DIRS` (repeatable: `--include-dir build --include-dir tools`). Some
+  ecosystems keep genuine hand-written source under a directory name graft
+  otherwise treats as build output (e.g. a `build/` that isn't generated).
+  The override is persisted per repo: set it once and every later no-flag
+  `graft build`, plus the hooks/refresh path (which never sees CLI flags at
+  all), include it identically. It lifts only graft's own skip list — in a
+  Git repository, Git's ignore rules stay authoritative, so a directory that
+  is both skip-listed and gitignored needs un-ignoring (or `git add -f`) too,
+  the same contract indexing already applies to tracked-but-ignored files.
+  Dot-directories are never overridable. Reaches the wiring graph, the Tier-2
+  markdown/concept pipeline, Go module discovery, and workspace child builds
+  alike, and is validated up front (a bare directory name only — no paths, no
+  dot-prefixes).
+
 ### Fixed
 
 - **`graft map` no longer promotes unrelated methods into hubs and hotspots.**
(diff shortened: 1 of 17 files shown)


Aizen-sama214:
Thanks for this! We’ve merged #51, which changed file enumeration to respect .gitignore. Your --include-dir feature is still useful for overriding built-in skips. Could you rebase onto main and adapt it to the new walker? Happy to help with conflicts.

qoole:
Rebased onto latest main and adapted to the #51 walker — 557/557 green.

Semantics, as adapted: `--include-dir` lifts only graft's **own** skip list, matching your framing. In a Git repo the override reaches the git-visible enumeration (`skippedPath` now honors it per segment), so a committed-but-skip-listed directory like a vendored `vendor/` gets indexed — but Git's ignore rules stay authoritative: an ignored directory remains excluded even when named, the same contract #51 applies to tracked-vs-ignored files. Both behaviors are pinned in `test/ingest-fs.test.ts`.

Structural notes:
- `buildGraph`'s single enumeration now threads the persisted override, so scope discovery and Go-module resolution inherit it from the same file set — the previous separate `readGoModules` threading collapsed into a doc note.
- The Tier-2 context pipeline and `discoverWorkspaceChildren` read the same persisted config, so a no-flag rebuild and the hooks/refresh path behave identically to the invocation that set it.
- Heads-up: this PR and #41 both introduce the `shouldSkipDir` helper in `src/ingest/fs.ts` (they're deliberately independent). Whichever merges second will have a small conflict there — happy to restack the survivor on demand.

qoole:
Closing this out — the series landed on main via the 0.9.0 release (3455201…6d0f227, authorship intact — much appreciated), including your follow-up moving the persisted config into git-ignored `.graft/config.json`, which is a strict improvement over parking it in the cache dir. Nothing left here that main doesn't already have.

PR #46, opened 2026-08-04
feat: add C# language support

reinhardtb:
## Summary

Adds C# support to Graft's Tier-1 depth graph on the current extractor architecture. `.cs` files are routed through the dedicated C# AST extractor rather than the generic breadth tier.

### What's implemented

**Language detection**: `.cs` and `.CS` files map to the `csharp` language type and the depth extractor owns the extension.

**Construct extraction**:
- `class_declaration` → `class`
- `struct_declaration` → `struct`
- `interface_declaration` → `interface`
- `record_declaration` → `class`
- `enum_declaration` → `enum`
- `method_declaration`, `constructor_declaration`, and `property_declaration` → `method`
- `using_directive` import edges, including static and aliased imports
- `invocation_expression` call edges with member receiver resolution
- Explicit `public` visibility, with interface members treated as public

**C# heritage**:
- Interface base lists emit `extends` for every inherited interface.
- Structs and resolved interface-only class bases emit `implements`.
- Resolved class bases emit `extends`.
- C# base-list ambiguity is corrected during graph resolution using the target declaration kind.

**Bindings and calls**:
- Field declarations bind both bare field access and `this.field` access.
- Typed parameters retain their receiver types.
- `var value = new Type()` binds to the constructed type.
- Bare instance method calls are resolved as implicit `this` calls.

**Dependencies**: `tree-sitter-c-sharp@0.21.3`, compatible with `tree-sitter@0.21.x`.

### Rebase status

Rebased onto the current `trailhq/main` (`63421bc`), which landed the `EXTENSIONS` refactor that replaced the old `Language` union / `languageOf()` this PR was originally built on, and moved `.cs` into the breadth tier. This PR is now the depth-tier upgrade on top of that architecture: `.cs` is removed from `GENERIC_LANGS` (and its breadth query deleted), the depth register owns the extension, and every registry (`KINDS_BY_LANG`, `CALL_TYPES`, `GRAMMARS`) and the bindings/resolve passes carry the new language. The only rejection the previous round of feedback did not cover — the `extract.ts` hunks — is resolved by this rebase; current Swift/Java/PHP/generic tests are preserved and green.

### Verification

Verified with the repo's own commands:

- `env -u ELECTRON_RUN_AS_NODE npm run build`
- `env -u ELECTRON_RUN_AS_NODE npm test`
- **1163 tests passed, 0 failed** (run with bounded concurrency; the full-parallel default exhausts the local 12-core sandbox, so CI's provisioned runner is the authoritative run)

### Test coverage

`test/graph-csharp.test.ts` covers:
1. Depth-tier ownership and language routing
2. Classes, interfaces, structs, enums, records, methods, properties, and visibility
3. Interface inheritance and class/struct implementation relationships
4. Bare fields, `this.field`, typed parameters, and `var` receiver resolution
5. Implicit-`this` method calls
6. `using`, static `using`, and aliased imports


anirudhkumar-nanonets:

grahas:
+1 on this, C# support would be awesome

shhdwi:
Reviewed on top of current `main` — holding for an author rebase. main now supports C# via the **breadth tier** (`generic.ts` + `c_sharp.scm`), and it landed the `EXTENSIONS` refactor that replaced the old `Language` union / `languageOf()` this PR is built on. Result: `extract.ts` conflicts across ~11 hunks and the new depth `.cs` would collide with the existing breadth `.cs`. This would be a welcome depth-tier upgrade (like PHP #64 was over breadth), but it needs a rebase onto the current extractor architecture + removing `.cs` from `GENERIC_LANGS` — more than a mechanical merge. Happy to re-review once rebased.

reinhardtb:
Apologies, this has slipped under the radar. Will rebase during the course of the day.

reinhardtb:
Done - please review

reinhardtb:
Rebased onto the current `trailhq/main` (`63421bc`). The rework follows the feedback:

- `.cs` is now a depth-tier extension: it was removed from `GENERIC_LANGS` (and the `c_sharp.scm` breadth query deleted), so exactly one tier owns it — `genericLangOf("x.cs")` returns `null` again.
- The C# depth extractor is ported onto the `EXTENSIONS` architecture: `Language` union, `EXTENSIONS`, `KINDS_BY_LANG`, `CALL_TYPES`, `GRAMMARS`, plus the Swift/Java/PHP-era branches in `walk`/`describe`/`heritageEdges`/`calleeName`/`isImport`/`importSpecifier` all carry the new language without touching current main's Swift/Java/PHP behavior.
- Bindings, heritage relation correction in `resolve.ts`, `types.ts` comments, dependency (`tree-sitter-c-sharp@0.21.3`) and the `test/graph-csharp.test.ts` suite are all preserved.

Verify locally: `npm run build` and `npm test` (1163 tests, 0 failed here; CI is the authoritative full-parallel run).

Issue #36, opened 2026-07-30
Two small accuracy fixes: build banner under-reports indexed languages; README overstates what `init` writes for Claude Code


dudarenok-maker:
Two unrelated one-liners found while evaluating 0.8.2. Happy to split into separate issues if you'd rather.

## 1. The build banner omits languages that were actually indexed

A repo of 4 `.ts` files plus 1 `.mjs`:

```console
$ graft build
✓ wiring: 11 nodes (5 file, 5 function, 1 interface), 9 edges, 5 cards [typescript]
  parsed: 5 of 5 files
```

The banner reports `[typescript]`, but the `.mjs` file is indexed and its symbols are queryable:

```console
$ graft grep mjsOnlySymbol
"mjsOnlySymbol" — 1 hits in 1 symbols across 1 files (searched 5 indexed files)

mjsOnlySymbol · function · scripts\tool.mjs:L1-L3 · 0 in-edges
```

Same on a larger repo: the banner said `[python, tsx, typescript]` while 155 cards were generated under a `scripts/` tree of 82 `.mjs` files.

The banner is the natural place to check coverage, so under-reporting sends people hunting for a problem that isn't there. It also makes genuinely-absent languages hard to distinguish from ones that simply aren't listed — in my case Dart really was unindexed (0 cards), and I initially assumed `.mjs` was too.

## 2. README: `init` no longer writes a CLAUDE.md section

The README describes the Claude Code wiring as adding a marker-fenced section to `CLAUDE.md` or the equivalent instruction file. As of 0.8.2 it doesn't — it writes `.claude/skills/graft/SKILL.md`, and the fenced-section behaviour applies to `GEMINI.md` and `.github/copilot-instructions.md`:

```console
$ graft init --dry-run -y --no-build
would write — this repo:
  .claude\settings.json                 graft statusline + hook blocks
  .claude\helpers\graft-statusline.cjs  statusline shim
  .claude\helpers\graft-hooks.cjs       hooks shim
  .claude\skills\graft\SKILL.md         graft skill
  .mcp.json                             mcpServers.graft
  GEMINI.md                             fenced graft section
  .gemini\settings.json                 mcpServers.graft
  .github\copilot-instructions.md       fenced graft section
```

[...]

Issue #35, opened 2026-07-30
`map` clusters one entry per file instead of per directory


dudarenok-maker:
`graft map` is documented as "directory clusters, per-directory hubs, and global hotspots", but it emits one row per *file*, each reporting `1 files`.

## Repro

5 files across 3 directories (`src/` ×3, `lib/` ×1, `scripts/` ×1):

```console
$ graft map
repo map — 5 files · 6 symbols · 9 edges · typescript

src\consumer.ts     1 files · 2 symbols
lib\helper.ts       1 files · 1 symbols
scripts\tool.mjs    1 files · 1 symbols
src\direct.ts       1 files · 1 symbols
src\gate.ts         1 files · 1 symbols   hubs: isActive (src\gate.ts, 1←)
```

## Expected

Three rows — `src` (3 files), `lib` (1 file), `scripts` (1 file).

## Actual

Five rows, one per file, each reporting `1 files`.

## At scale this makes the command unusable

On a 1,958-file repo the same output gives one single-file "directory" per file and a footer reading `… +1942 more directories not shown (raise max-dirs to see more)` — so the token-budgeted orientation view spends its entire budget describing 16 arbitrary files instead of the repo's shape.

## Possibly a separate issue

Because hubs/hotspots rank by bare symbol name, the top entries on that repo were `filter` (517←), `get` (441←, from a test file), `trim`, `post` and `max`. Distinct same-named methods across unrelated classes appear to be merged into a single node, so the ranking reflects name frequency rather than coupling. Happy to split this out if you'd prefer it tracked separately.

## Environment

graft 0.8.2 · Node 24.15.0 · Windows 11.

anirudhkumar-nanonets:
Clustering half fixed in #44 — same root cause as #33 (`dirKey` splits on `/`, so a Windows path was one segment). Your five-file repro is now a test asserting three rows.

Leaving this open for the hub-ranking half. One correction: distinct same-named methods aren't merged. When a member call's receiver type is unknown, edge resolution wires it to a *unique* repo-wide match of that name — so one uniquely-named method absorbs every same-named call, builtins included. Reproduces here: `Bindings.set` is the only user symbol named `set` and shows 9 in-edges, most of them real `Map.set()` calls.

Issue #34, opened 2026-07-30
`callers` misses consumers that reference a function without calling it directly


dudarenok-maker:
`callers` finds direct call sites but misses modules that import a function and pass it around as a value — for example the common injectable-default pattern `opts._fn ?? fn`. The omission is silent, which is risky for the blast-radius use case the command is meant for.

## Repro

```ts
// src/gate.ts
export function isActive(): boolean {
  return true;
}
```

```ts
// src/direct.ts — calls it directly
import { isActive } from './gate.js';

export function callsItDirectly(): boolean {
  return isActive();
}
```

```ts
// src/consumer.ts — imports it, uses it as a value
import { isActive } from './gate.js';

interface Opts {
  _isActive?: typeof isActive;
}

export function run(opts: Opts = {}): boolean {
  const check = opts._isActive ?? isActive;
  return check();
}
```

```console
$ graft build && graft callers isActive
isActive · function · src\gate.ts:L1-L3
  calls ← callsItDirectly (src\direct.ts:L3-L5)
```

## Expected

Both `src/direct.ts` and `src/consumer.ts` listed — `consumer.ts` imports `isActive` and invokes it on the default path.

## Actual

Only `direct.ts`. `consumer.ts` is absent, with no indication that anything was skipped.

## Why it matters

Injectable defaults for testing (`opts._dep ?? realDep`) are a common pattern, and a codebase that uses them gets a blast radius that looks authoritative but is short. I hit this on a real repo: a gate function reported one consumer (its own test file) while a production module importing it that way was missing.

The README notes that method resolution depends on type annotations and constructor assignments, but this is a plain top-level function import, so it reads as in-scope for `callers`.

Surfacing these as a weaker edge class — `references ←` alongside `calls ←` — would be enough to avoid the silent miss, without needing full value-flow analysis.

## Environment

graft 0.8.2 · Node 24.15.0 · Windows 11.

PR #58, opened 2026-08-06
feat: PowerShell language support (.ps1/.psm1)

qoole:
> **Builds on #40** (tree-sitter 0.25) — this branch is stacked on that PR's branch, so its two commits appear here until #40 merges; the PowerShell work is the five commits from `chore: add PowerShell tree-sitter grammar` onward. Opening as a draft until #40 lands, then I'll rebase onto main and mark it ready.

Tier-1 graph extraction for PowerShell via tree-sitter-powershell (airbus-cert; prebuilds for all six platform/arch combos — the reason #40's 0.25 runtime bump is the prerequisite).

## What

- **Definitions**: functions (incl. `filter`/`workflow`), classes, methods & constructors, enums. `exported=true` for top-level defs and class members; functions nested inside functions/methods are scope-local → false. (PowerShell has no file-local export syntax; `Export-ModuleMember`/manifest interpretation is cross-file and deferred — see limitations.)
- **Call edges**: command invocations (scope-/module-qualifiers stripped: `script:Foo`, `Mod\Get-Foo`), member calls `$obj.M()`, static calls `[Type]::M()`, call-operator statics `& Get-Foo`. Dynamic invocation (`& $var`) produces no edge.
- **Import edges**: dot-sourcing (`. ./x.ps1`, quoted paths, backslash-normalized), `Import-Module` with relative paths (module *names* stay external specifiers, like Go stdlib imports), `using module` (extracted from its command shape — the grammar has no dedicated node).
- **Case-insensitive resolution**: PowerShell is case-insensitive; call/heritage/typed-member lookups case-fold **within a PowerShell-only domain** (a PS `FOO` can never link to a Python `foo`), original casing preserved everywhere, lookup order exact-local → folded-local → exact-global → folded-global (local scope wins, matching PS semantics). Folded matches surface as `confidence: "inferred"`.
- **Typed-receiver bindings**: typed params (incl. decorated `[Parameter(...)][Widget]$w`), typed properties, cast declarations, `[Widget]::new()`, `New-Object` (positional and `-TypeName`), `$this` → enclosing class.
- **Builtin-cmdlet denylist**: calls to the core cmdlet/Pester-DSL surface (`Get-ChildItem`, `Describe`, …) never bind to a same-named in-repo definition (usually a Pester mock) — dropped-edge-is-safe, the same posture the resolver already takes for unresolvable member calls.
- **Ingest**: UTF-16LE `.ps1` (Windows PowerShell 5.1 `Out-File` default) decoded at the shared `readSourceFile` site (hash-what-you-parse holds); Pester `*.Tests.ps1` classified as test files and de-ranked accordingly; `.psd1` deliberately not graph-parsed (manifests yield no definitions).
- Deep grammar import (`tree-sitter-powershell/bindings/node/index.js`) avoids a DEP0151 warning on every CLI run (package has extensionless `main` + no `exports` map).

## Fit with recent main

This was originally built against pre-0.9.0 main and re-anchored after the recent landings; the seams were adapted rather than ported blind:

- **Language-vs-parser reporting (#36)**: PowerShell slots into the `EXTENSIONS` table as two rows (`.ps1`/`.psm1` → grammar `powershell`, label `powershell`), so the banner/map coverage report is correct by construction.
- **Imported-function references (#34)**: the "don't descend into import statements" rule gets one PS carve-out — a dot-source can wrap a whole script block (`. { ... }`), whose body must still be walked for its own calls. Pinned by test.
- **Strict member-call resolution (#35)**: PS member calls inherit the drop-don't-guess rule unchanged; the PS case-folding applies inside `resolveTypedMember`'s owner-qualified lookup, never as a bare-name fallback.
- **Gitignore-aware walker / posix paths / UTF-16 read coherence / encoding-drift**: PS files flow through the same enumeration, path storage, and `readSourceFile` contracts as every other language — no PS-specific read or walk path exists.

## Evidence

- Suite: **618/618 green** on this branch (35 PowerShell-specific tests: full-pipeline fixtures modeled on the Go tests, bindings, case-collision isolation, malformed-input tolerance, UTF-16, CRLF, >32 KB files, parser alternation, test-path classification pinning).
- Real-world battery (run on the pre-rebase branch; spot-revalidated post-rebase): 8 repos (Pester, ComputerManagementDsc, dbatools, ImportExcel, posh-git, PSFramework, Plaster, psake) — every build clean, double-build **byte-identical determinism 8/8**, incremental rebuild correct 8/8 (dbatools: 1,707 files, full 34 s / incremental 2.5 s).
- Function recall vs an exhaustive regex census: 88–98% per repo, with **every** miss attributed to upstream grammar parse errors, none to extraction logic.
- Resolution quality on dbatools: the most-called function (`Stop-Function`, 2,857 call sites, shadowed by 10 Pester mocks) resolves 726 incoming call edges to its production definition.
- Import wiring on posh-git: **24/30** import edges resolve to in-repo files via literal `$PSScriptRoot` resolution (was 0/30); the 6 unresolved are genuinely dynamic (runtime variables) and stay raw by design. Re-verified on this exact branch post-rebase, including byte-identical cold-rebuild determinism.
- Parse health (upstream grammar, not this integration): 41–94% of files parse error-free per repo; the dominant causes bisect to two airbus-cert grammar gaps (unquoted comma-separated bareword arguments; bare trailing scriptblock arguments). Extraction degrades gracefully — error-containing files still yield correct nodes and edges. Happy to file those upstream grammar issues.

## Known limitations (deliberate)

- `Export-ModuleMember` / manifest `FunctionsToExport` not interpreted (uniform `exported=true` instead) — the common unquoted multi-name idiom doesn't parse in the current grammar anyway.
- `$PSScriptRoot`-interpolated, `Join-Path`-computed, and wildcard import arguments degrade to raw specifiers.
- `.psd1` manifest relationships (RootModule/NestedModules) not graphed.
- Monorepo scope markers unchanged (PS has no fixed-name manifest file).
- Class base lists conflate base class and interfaces as `extends` (PS cannot define interfaces; targets stay external names).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,10 @@ updates:
     schedule:
       interval: weekly
     open-pull-requests-limit: 10
+    ignore:
+      # Upgrading this independently of the pinned tree-sitter runtime breaks
+      # peer resolution; see the overrides block in package.json.
+      - dependency-name: tree-sitter-typescript
 
   - package-ecosystem: github-actions
     directory: /
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
-          node-version: '20'
+          node-version: '22'
           cache: npm
 
       - name: Install dependencies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Added
+
+- **PowerShell code graphs.** `.ps1` and `.psm1` files now contribute functions,
+  classes, methods, enums, calls, imports, and case-insensitive receiver bindings.
+  Dot-sourcing, `Import-Module`, and `using module` resolve local modules, while
+  UTF-16LE source from Windows PowerShell is decoded during graph ingestion.
+
 ### Changed
 
 - **Bump the tree-sitter runtime to 0.25, and the Python and Go grammars to match.**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,8 +101,10 @@
   `tree-sitter-typescript@0.23.2` is still the latest release and declares a
   `peerOptional tree-sitter@^0.21.0`, so an npm `overrides` entry pins its
   nested `tree-sitter` dependency to the root one and lets the install resolve
-  cleanly. No behavior change; the grammar's native binding loads and parses
-  correctly under tree-sitter 0.25.1, and the full suite is green.
+  cleanly. Syntax-node comparisons now use Tree-sitter's stable node IDs rather
+  than JavaScript object identity, whose behavior differs between runtime
+  versions. The minimum Node.js version is now 22.12, matching Commander 15 and
+  ensuring Windows installs can use Tree-sitter 0.25's prebuilt native module.
 
 [#33]: https://github.com/NanoNets/Graft/issues/33
 [#35]: https://github.com/NanoNets/Graft/issues/35
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,14 @@
   matches. Pass a real prefix (`--in server/src/gpu`), a full file path
   (`--in src/a.ts`), or use `grep`'s pattern to match on content.
 
+- **Bump the tree-sitter runtime to 0.25, and the Python and Go grammars to match.**
+  Groundwork for adding new language grammars that require the 0.25 runtime.
+  `tree-sitter-typescript@0.23.2` is still the latest release and declares a
+  `peerOptional tree-sitter@^0.21.0`, so an npm `overrides` entry pins its
+  nested `tree-sitter` dependency to the root one and lets the install resolve
+  cleanly. No behavior change; the grammar's native binding loads and parses
+  correctly under tree-sitter 0.25.1, and the full suite is green.
+
 [#33]: https://github.com/NanoNets/Graft/issues/33
 [#35]: https://github.com/NanoNets/Graft/issues/35
 [#36]: https://github.com/NanoNets/Graft/issues/36
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@nanonets/graft",
-  "version": "0.8.2",
+  "version": "0.9.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@nanonets/graft",
-      "version": "0.8.2",
+      "version": "0.9.0",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,7 @@
         "typescript": "^6.0.3"
       },
       "engines": {
-        "node": ">=20"
+        "node": ">=22.12.0"
       }
     },
     "node_modules/@anthropic-ai/sdk": {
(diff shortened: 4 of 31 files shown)


qoole:
Added a domain-isolation fix found while extending the language work: PowerShell method nodes were entering the shared member-owner index, so a typed member call from another language could in principle wire cross-language to a PS class method. PS methods (and PS heritage edges) now live only in their own indexes, mirroring how PS bare names were already isolated. Suite is 621/621 on the branch.

qoole:
Which tier do you want PowerShell in — if you want it at all? I'd rather rebase once, toward whatever you'd actually merge. Default plan is the breadth row unless you say otherwise.

The constraint that decides it: every published `tree-sitter-powershell` (0.25.9 through 0.26.4) declares `peerDependencies: { "tree-sitter": "^0.25.0" }`, and main is still on `"tree-sitter": "^0.21.1"`. That's why this was stacked on #40, and it means the depth version of this branch can't be rebased onto current main at all until the runtime moves. Kotlin could be promoted natively in #130 only because `tree-sitter-kotlin` still peers on `^0.21.0`; PowerShell has no version that does.

The breadth row has no such problem. `tree-sitter-powershell.wasm` ships in the `tree-sitter-wasm` 1.1.4 bundle the lockfile already pins, built from the same airbus-cert grammar this branch uses natively — the bundle declares `tree-sitter-powershell: ^0.26.4`, the identical range. So it's a `GENERIC_LANGS` row, a hand-written `src/graph/queries/powershell.scm`, a `SNIPPETS` case in `test/generic-extract.test.ts`, and the README Broad list plus its "Twenty-two languages" count — the Nix #159 / Lua #187 shape, with no `package.json` churn. Two things I checked before proposing it: the `.scm` has to be written from scratch (airbus-cert ships no tags query), and `classifyKind`'s walker fallback matches `DEF_SUFFIX` plus a class/struct/enum-ish prefix, neither of which `function_statement` hits — so a bare row with no query would mint classes, enums and `class_method_definition` methods, zero functions, and no call edges at all, since the fallback path is symbols-only. I also haven't warmed that wasm under `web-tree-sitter` 0.25.10 yet; the bundle builds with `tree-sitter-cli` 0.26.x, this grammar has an external scanner, and #139 shows a bundled grammar in the same 1.1.4 can crash, so I'd smoke-test warm + parse before opening anything.

Worth stating plainly rather than letting it turn up mid-review. A breadth row carries the definitions (functions incl. `filter`/`workflow`, classes, methods, constructors, enums) and bare-name command call edges — but those resolve only where case and qualifier match exactly, since the breadth resolver is exact-match and `loadQuery` strips `#gsub!` before compiling. The resolver work in this branch is what moves those numbers: scope-qualifier stripping (`script:Foo`, `Mod\Get-Foo`), case-folding inside a PS-only domain, dot-source / `Import-Module` / `$PSScriptRoot` import edges, typed receivers, and the builtin-cmdlet denylist with its test-aware tie-break. On dbatools, `Stop-Function` has 2,857 call sites and 10 Pester mocks redeclaring the name; with the tie-break it resolves 726 incoming edges to the production definition, and without it the never-guess ambiguity rule drops all of them. On posh-git, imports are 24/30 with `$PSScriptRoot` resolution and 0/30 without. Both tiers also inherit the same grammar ceiling — 41–94% of files parse error-free per repo, from two airbus-cert gaps (unquoted comma-separated bareword arguments, bare trailing scriptblock arguments) — so neither option fixes that. UTF-16LE `.ps1` decode already works on main through `readSourceFile`, so that part costs nothing either way.

Independent of the answer: main's `isTestPath` doesn't classify Pester's `*.Tests.ps1` — the `\.test\.[a-z]+$` alternative misses it, since the next character is `s`. One-line regex change plus a test; happy to send it standalone.

This branch predates `generic.ts` entirely, and the breadth tier gets the symbols-and-calls half of it in four files instead of thirty-one — hard to argue with. Breadth now with a promote-to-depth follow-up once the runtime moves is the path PHP (#64, then #157) and Kotlin (#130) took, so nothing gets burned. And if PowerShell isn't wanted at all — nothing in the tracker asks for it — say so and I'll close this instead.

PR #59, opened 2026-08-06
feat: Rust language support (.rs)

qoole:
> **Builds on #58** (PowerShell support), which builds on #40 (tree-sitter 0.25) — this branch is stacked on #58's branch, so ancestor commits appear here until those PRs merge; the Rust work is the ten commits from `chore: add Rust tree-sitter grammar` onward. Opening as a draft; I'll rebase and mark ready as the stack beneath it lands.

Tier-1 graph extraction for Rust via tree-sitter-rust 0.24.0 (the tree-sitter org's own grammar; prebuilds for all six platform/arch combos — #40's 0.25 runtime is the prerequisite, and the same npm `overrides` pattern handles its peer range).

## What

- **Definitions**: functions, structs, enums, traits (→ `interface`), type aliases, unions, methods with impl-block owners (generics-stripped: `impl Cache<T>` owns as `Cache`), and `macro_rules!` definitions. Inline modules contribute scope segments (`file.rs#imp.open`), so cfg-gated sibling mods stay distinct and resolvable.
- **Visibility**: `pub`/`pub(crate)`/`pub(super)` → exported; `pub(self)` and plain items → not; trait-impl methods inherit the trait contract (exported); `#[macro_export]` macros exported; everything inside `#[cfg(test)]` mods unexported (`#[cfg(not(test))]` is correctly NOT matched).
- **Calls**: bare calls, method calls with typed-receiver inference (typed `let`, constructor-let `T::new()`, typed params, `self` → enclosing impl), `Type::method` / `Self::method` static calls, turbofish unwrapped, tuple-index calls guarded. Scoped calls (`crate::worker::run()`) resolve through the module tree and are **restricted to the resolved file's candidates — never stripped to a bare name**.
- **Macros**: invocations and definitions live in their own namespace — names carry the `!` (`log_it!`), so a macro can never wire to a same-named function (Rust's namespaces are distinct). Std/prelude macros are skipped at extraction; path-qualified forms (`std::println!`) are recognized too.
- **Imports**: `use` paths (all forms: scoped, grouped per-leaf, renames, wildcards) and `mod` declarations resolve through the real module tree — `crate::`/`super::`/`self::`, both `x.rs` and `x/mod.rs` layouts, the parser-file rule (`src/parser.rs`'s children live in `src/parser/`), inline-mod `super` consumption (a `use super::*` inside `#[cfg(test)] mod tests` correctly refers to its own file), and integration-test crate roots (`tests/*.rs` anchor `crate::` to themselves, not the library).
- **Workspaces**: each `Cargo.toml`'s `[package]` name (section-aware parse; `[[bin]]` names ignored) maps cross-crate imports (`foo_bar::…` ↔ `crates/foo-bar/`), including dependency `{ package = "…" }` aliases; duplicate package names are treated as ambiguous and dropped rather than guessed, and crate matching is deterministic regardless of enumeration order.
- **Domain isolation**: Rust names resolve in a Rust-only domain — separate global-name and method-owner indexes — so a Rust symbol never links to a same-named symbol from another language, in either direction.
- **Ambiguity posture**: a type with several trait impls of the same method name (`From<A>`/`From<B>` both giving `Frame.from`; `Display`/`Debug` both giving `fmt`) drops the member call as ambiguous instead of guessing first-in-file. Drop-don't-guess throughout, matching #35's resolution philosophy.

## Evidence

- Suite: **all green** on this branch (77 Rust-specific assertions across extraction, bindings, resolution, and the review-derived regression pins below; plus CRLF, >32 KB, malformed-input tolerance, and double-build determinism guards mirroring the PowerShell suite).
- Real-world battery (two production Rust codebases, 350 files total, one a multi-crate polyglot workspace with a TS frontend): builds clean, **byte-identical double-build determinism on both**, **99.5%** and **91.8%** of internal-path imports wired (the remainder are dropped, never misdirected), 81 imports wired cross-crate through the workspace map, and **zero cross-language edges in either direction** in the polyglot repo.
- The branch went through two independent adversarial review passes before this PR; every wrong-edge finding became a failing-test-first fix (crate-root anchoring corner cases, inline-mod call paths, workspace crate-name shadowing of `mod` declarations, `#[path]` attributes, the cross-language method-owner leak). The reviews' dropped-edge findings are disclosed below rather than patched around.

## Known limitations (deliberate; all fail toward dropped edges, never wrong ones)

- Macro token-tree bodies are opaque — `assert_eq!(compute(), 3)` contributes no `compute` call edge (no expansion).
- UFCS calls (`<T as Tr>::method()`) drop.
- `#[path = "…"]` mod declarations drop (resolving conventionally would wire the wrong file).
- Cross-file `impl Trait for Type` heritage is dropped (trait-default-method resolution needs the type defined in the same file).
- `impl` blocks inside function bodies mint unowned functions.
- Files nested deeper than one level under `tests/` don't anchor `crate::` (statically ambiguous — each test binary is its own crate).
- `pub use` re-export chains and proc-macro expansion are unsupported.
- Grammar 0.24.0 lacks Rust 1.82's `&raw const/mut` (~2–4% of files in modern codebases parse with errors; extraction degrades gracefully and still yields those files' other definitions). Happy to file that upstream.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,10 @@ updates:
     schedule:
       interval: weekly
     open-pull-requests-limit: 10
+    ignore:
+      # Upgrading this independently of the pinned tree-sitter runtime breaks
+      # peer resolution; see the overrides block in package.json.
+      - dependency-name: tree-sitter-typescript
 
   - package-ecosystem: github-actions
     directory: /
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
-          node-version: '20'
+          node-version: '22'
           cache: npm
 
       - name: Install dependencies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Added
+
+- **PowerShell code graphs.** `.ps1` and `.psm1` files now contribute functions,
+  classes, methods, enums, calls, imports, and case-insensitive receiver bindings.
+  Dot-sourcing, `Import-Module`, and `using module` resolve local modules, while
+  UTF-16LE source from Windows PowerShell is decoded during graph ingestion.
+
 ### Changed
 
 - **Bump the tree-sitter runtime to 0.25, and the Python and Go grammars to match.**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,8 +101,10 @@
   `tree-sitter-typescript@0.23.2` is still the latest release and declares a
   `peerOptional tree-sitter@^0.21.0`, so an npm `overrides` entry pins its
   nested `tree-sitter` dependency to the root one and lets the install resolve
-  cleanly. No behavior change; the grammar's native binding loads and parses
-  correctly under tree-sitter 0.25.1, and the full suite is green.
+  cleanly. Syntax-node comparisons now use Tree-sitter's stable node IDs rather
+  than JavaScript object identity, whose behavior differs between runtime
+  versions. The minimum Node.js version is now 22.12, matching Commander 15 and
+  ensuring Windows installs can use Tree-sitter 0.25's prebuilt native module.
 
 [#33]: https://github.com/NanoNets/Graft/issues/33
 [#35]: https://github.com/NanoNets/Graft/issues/35
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,14 @@
   matches. Pass a real prefix (`--in server/src/gpu`), a full file path
   (`--in src/a.ts`), or use `grep`'s pattern to match on content.
 
+- **Bump the tree-sitter runtime to 0.25, and the Python and Go grammars to match.**
+  Groundwork for adding new language grammars that require the 0.25 runtime.
+  `tree-sitter-typescript@0.23.2` is still the latest release and declares a
+  `peerOptional tree-sitter@^0.21.0`, so an npm `overrides` entry pins its
+  nested `tree-sitter` dependency to the root one and lets the install resolve
+  cleanly. No behavior change; the grammar's native binding loads and parses
+  correctly under tree-sitter 0.25.1, and the full suite is green.
+
 [#33]: https://github.com/NanoNets/Graft/issues/33
 [#35]: https://github.com/NanoNets/Graft/issues/35
 [#36]: https://github.com/NanoNets/Graft/issues/36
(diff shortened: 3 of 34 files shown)


qoole:
One question, answerable in a line: do you want Rust promoted to the depth tier the way Java (#83), PHP (#64) and Kotlin (#130) were, or is `--lsp` + rust-analyzer the intended answer for Rust? I'll act on either today rather than leave this sitting in the queue looking like reviewable work.

Context first: main is 0.13.0, this branch's merge base is `fda5b7f` (0.9.0), and it's stacked #40 → #58 → #59 — so Files changed shows the PowerShell work and the runtime bump alongside the Rust commits. Don't read the diff as-is. Breadth already covers Rust: the `GENERIC_LANGS` row, `src/graph/queries/rust.scm`, and @shhdwi's #96 wiring `use` → import edges, on WASM with no node-gyp build. #96 also resolves `crate::` and disambiguates `x.rs` from `x/mod.rs`, which I'd otherwise have claimed here. That's the coverage half of this PR, done.

The objection I should raise rather than wait for: `--lsp` already gives Rust receiver typing, and rust-analyzer is the first row in `LSP_SERVERS`. So the depth-tier claim is narrower than the PR body makes it sound — the same `recvType` edges in a plain `graft build`, deterministic and with no toolchain on `PATH`, plus three things breadth can't express: impl-block owner stamping, so `impl Config { fn parse() }` is owner-qualified instead of a bare `parse` indistinguishable from a free function (`generic.ts` mints `cap.name.text` verbatim); `super::`/`self::` and cross-crate paths from workspace `Cargo.toml` `[package]` names, including dependency renames; and macro namespacing, so `log_it!` can't wire to a function `log_it`. To be precise about what's actually lost today: an untyped `cache.get()` reaches the resolver as the bare name `get` and settles same-file-unique, else unique-global, else drops — never a fan-out. It's recall that's missing, not precision.

Costs. It's downstream of whatever you decide on #40 — native `tree-sitter-rust` against main's `^0.21.1` — and #40 itself needs widening before it installs, so this is moot until that's settled. The branch as written also mints none of the `mod`/`const`/`static` nodes `rust.scm` emits, and needs a `Cargo.toml` where main infers crate roots from `lib.rs`/`main.rs` paths; both mine to fix. And it lands on your tests: `test/generic-extract.test.ts` uses `.rs` as its worked example in five places, so a promotion means re-pointing those at another breadth language, which I'd do in the same PR. On the registry row I'd follow the PHP shape rather than churn #96 — drop the row as #157 did after #64, and leave `extractUses`/`resolveRustUse` in place, exactly as `extractPhpUses`/`resolvePhpUse` still sit on main.

For scale: 24 Rust-specific tests on the branch across `test/graph-rust.test.ts` and `test/graph-rust-resolve.test.ts` (CRLF, >32 KB chunk boundaries, byte-determinism, duplicate workspace package names, two cross-language isolation cases). A suite total measured against a 0.9.0 base would be meaningless, so I'll quote a real number only once it's rebased.

If the answer is that the breadth row plus `--lsp` is enough for Rust, that's a good answer and I'll close this.

PR #65, opened 2026-08-07
Gate the pull channel on match strength; rule-built crux for keyless builds

anirudhkumar-nanonets:
Two changes to what `graft ask` returns.

**`4ff0873` — gate the pull channel on match strength**

`isWeakMatch()` in `src/ask/ask.ts` suppresses the source pack when a lexical
result is below both floors (`coverageStrong < STRONG_FLOOR`, `coverage <
HIGH_FLOOR`). A weak pack now returns 3 pointers and no inlined source instead
of a full pack with a savings footer. `--full` bypasses the gate.

**`abdb043` — rule-built crux for keyless builds**

`ruleCrux()` builds a <=8-line crux deterministically, with no LLM and no API
key: signature plus the first meaningful lines for a function, a member
signature list for a class/interface/struct/trait/enum. `inlineSource()` falls
back to it before `sliceSpan`. Measured on one long function: 2848 -> 1071
chars (-62%). Adds `--crux` to build the LLM crux layer without concept nodes,
and `resolveBuildLayers()` in `src/cli-meta.ts` so layer selection is testable.

Tests: 13 new (7 in `test/ask.test.ts`, 6 in `test/cli-meta.test.ts`).

---

**On the benchmark numbers in README.md**

This branch does not change README.md. The SWE-bench table on main (commit
`00c9cc4`) covers 20 instances, and they were not all measured on this code:

- 11 instances ran on this branch's build (`graft-fix12-abdb043.tgz`, v0.9.0)
- 9 instances ran 2026-08-04 on npm 0.8.2, two days before these commits
  existed, under the older harness that still had web access enabled

Merging this makes the 11 the current behaviour. The 9 would need re-running to
put the whole table on one version.

### DIFF
--- a/src/ask/ask.ts
+++ b/src/ask/ask.ts
@@ -30,7 +30,7 @@ import {
 import { normalizePathPrefix } from "../util/paths.js";
 import { resolveSymbol } from "../graph/traverse.js";
 import type { EdgeV1, GraphV1, NodeV1, Relation } from "../graph/types.js";
-import { rankScopesAndFuse } from "./fuse.js";
+import { rankScopesAndFuse, HIGH_FLOOR, STRONG_FLOOR } from "./fuse.js";
 import { personalizedPageRank } from "./graphrank.js";
 import { readSourceFile } from "../util/source.js";
 import { counts, tokenize, type AskIndex, type AskIndexDoc } from "./index-file.js";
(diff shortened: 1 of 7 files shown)


shhdwi:
Reviewed on top of current `main` — holding for an author rebase; the conflicts are small textually but semantically hard (106 commits since the merge base):
- `ask.ts` output shaping diverged: this PR reworks an `askSavingsFooter` + weak-pack suppression, but main **replaced the footer with a leading `askSavingsLine`** — the footer no longer exists, so the suppression needs re-expressing against the new design.
- The strength signal partly landed already (`coverageStrong`/`STRONG_FLOOR`/`HIGH_FLOOR` are computed + exported on main via fuse.ts), so `isWeakMatch` would sit on top of existing infra — but the actual suppression gate is **not** yet on main, so it's not superseded, just needs rebuilding on the new signal.
- `cli.ts`'s build action was heavily rewritten on main (telemetry wrapping, `command: Command` param, new flags, `patchBuildConfig` vs the PR's `writeBuildConfig`); the `--crux` / `resolveBuildLayers` work must be re-woven in.
- `ruleCrux` (keyless rule-built crux) is genuinely new and not superseded.

Both ideas are still wanted — please rebase and re-express them against the new savings-line + telemetry-wrapped build action. Happy to re-review.

PR #67, opened 2026-08-07
Add C/C++ to the Tier-1 symbol extractor; make unsupported languages fail loudly

### LINKED ISSUE #66: Add C/C++ to the tier-1 symbol extractor (and make unsupported languages fail loudly)
**Version:** graft 0.9.0 (via `npx @nanonets/graft`)
**Repo under test:** a C++17 game engine — 350 first-party `.cpp`/`.h` files, ~100k lines, plus
vendored SDL2/ENet.

## Summary

`build --deep` produced a genuinely useful concept map of this C++ codebase, but the **tier-1 symbol
graph indexed none of it**, so `callers`, `grep` and `skeleton` are unusable here. The concept pass
and the symbol pass disagree about which languages they support, and that asymmetry currently makes
the tool report **confident false negatives** rather than "unsupported".

Two requests, in priority order:

1. **Make the language gap visible** (small, high value — see "The misleading part" below).
2. **Add C/C++ to the extractor** (larger).

## What was measured

A full clean `build --deep` over 1671 files produced:

```
✓ wiring: 865 nodes (696 function, 120 method, 33 file, 16 class), 1929 edges, 33 cards [javascript, python]
```

- **33 files** in `graft/.graph/wiring.json` — 840 `.py` symbols and 25 `.js`, and **zero C++**.
  Those 33 are `tools/*.py` plus a handful of vendored SDL2 build scripts.
- Meanwhile the `--deep` concept pass produced **861 cards, 659 of which cite `.cpp`/`.h` sources**,
  with accurate summaries. So the LLM half handles C++ fine; only the parser half doesn't.

The cause looks straightforward: `package.json` bundles `tree-sitter-go`, `tree-sitter-python` and
`tree-sitter-typescript` only, and `dist/graph/extract.js` dispatches on a fixed extension set —
`.ts .tsx .js .jsx .mjs .cjs .py .go`. A `.cpp`/`.h` file never reaches a parser. By contrast
`dist/context/build.js` does list `.cpp`/`.hpp`/`.cc`, which is why the two passes diverge.

## The misleading part (worth fixing even if C++ support never lands)

On an unsupported language the failures don't read as "unsupported" — they read as authoritative
answers about the code:

```
$ graft grep snapEntityToFloor
no hits for "snapEntityToFloor" in 33 indexed files. The pattern may be too specific — retry
graft grep with a bare symbol name or short substring ... All indexed code was searched; use raw
grep -rn only for genuinely unindexed files (docs, configs, brand-new files)
```

`snapEntityToFloor` is defined and called many times in this repo. "All indexed code was searched"
is true and deeply misleading — the advice explicitly steers the caller *away* from `grep -rn`,
which is the only thing that would have worked. Likewise:

```
$ graft callers snapEntityToFloor
✗ no symbol "snapEntityToFloor" in the graph — check spelling or run graft build

$ graft skeleton src/game/enemy_ai.cpp
no definitions indexed for this file
```

"check spelling or run graft build" sends you to re-run a 20-minute build that cannot fix it.

This matters more than usual for agentic use. The MCP server's own instructions say *"Prefer these
tools over grep/read — one call usually replaces several file reads."* An agent that follows that on
a C++ repo will conclude a symbol doesn't exist. Suggested minimum fix:

- Have `build` report skipped-by-language counts, e.g.
  `skipped 350 files (no parser: .cpp, .h) — symbol tools will not cover them`.
- Have `grep`/`callers`/`skeleton` say *"this file's language has no parser"* when the target's
  extension isn't in `EXTENSIONS`, instead of "no hits" / "no symbol" / "no definitions".
- Scope the MCP "prefer these over grep" instruction to the languages actually indexed.

## On adding C/C++

`tree-sitter-cpp` exists and is maintained, so the grammar itself is easy. The extractor looks
hand-written per language rather than query-driven (`describeGo`, TS-specific branches in `describe`,
`isFunctionBoundary` enumerating JS/TS node types), so C++ needs its own `describeCpp` for
`function_definition`, `class_specifier`, `struct_specifier`, `namespace_definition` and
`template_declaration`, plus unwrapping `function_declarator` to reach the name.

The harder part is edges, and it may be worth scoping out of a first pass. Edge resolution here is
built on **import bindings** (`collectImportedSymbols`, `collectTsImportBindings`,
`withoutShadowedImports`). C++ has no import bindings — it has `#include`, namespaces, overloads and
templates — so name-only resolution would produce false edges between every same-named method in the
codebase.

**A definitions-only tier would already be worth shipping.** It makes `skeleton` and `grep` work on
C/C++, and it can honestly report that `callers` has no edge data for the language rather than
answering wrongly. Even partial support beats the current silent skip.

## Why it's worth it

C and C++ are a large share of the repos where a prebuilt symbol graph pays off most: big, old,
header-heavy trees where "who calls this" is genuinely expensive to answer. Right now graft's
`--deep` output is useful on such a repo while its free, fast tier silently isn't — and the free
tier is the part users reach for first.

Happy to test a branch against this codebase (350 C++ files, ~100k lines) and report back.

edrethardo:
Fixes #66 — both requests: make the language gap visible, and add C/C++ to the tier-1 extractor.

## 1. C/C++ in the Tier-1 symbol graph (definitions-only)

`.c` `.h` `.cpp` `.cc` `.cxx` `.hpp` `.hh` `.hxx` now parse with `tree-sitter-cpp`, under a single `c/c++` label (the cpp grammar parses C, and a `.h` can't be attributed to either language from its name alone).

**Extracted:** free functions; inline methods; out-of-class definitions (`float Physics::step(...)` → `method` named `step`, owner `Physics`, id `physics.cpp#Physics.step`); classes, structs, enums — found through `namespace` and `template` wrappers; destructors and operators keep their spelled names (`~Entity`, `operator==`). Prototypes and forward declarations are **not** definitions and are skipped, so a header's `void update(float);` never shadows the one real definition.

**Deliberately definitions-only, as the issue suggested.** C++ has no import bindings — only `#include`, namespaces, overloads, templates — so the name-only edge resolution the other languages build on would fabricate a `calls` edge between every same-named method in the repo. The extractor emits no call/reference/import edges for C/C++ (structural `contains` edges only), and `callers` on a C/C++ symbol answers honestly:

```
snapEntityToFloor · function · src/game/entity_ground.h:L28-L90
  no edge data — this symbol is in a definitions-only language (C/C++): call edges are
  not extracted, so the graph cannot know its callers. Find its uses with graft grep
  "snapEntityToFloor" or raw grep -rn
```

## 2. Unsupported languages fail loudly

On a repo whose code has no parser, the old messages were confident false negatives — "no hits … All indexed code was searched", "no symbol — check spelling", "no definitions indexed" — and explicitly steered the caller *away* from raw grep, the only tool that would have worked. Now:

- **`graft build`** prints `skipped: N files (no parser: .rs, …) — symbol tools will not cover them` (single-repo and per-workspace-child), and records the gap in the graph as `meta.unindexed` (omitted when empty; old graphs unaffected).
- **Zero-hit `grep`** replaces the "use raw grep -rn only for genuinely unindexed files" tail with the gap: *"…but N source files (.ext) have no parser and are NOT in the symbol graph — if the target lives there, raw grep -rn / reading the file is the right tool"*.
- **Unknown-symbol `callers`** (CLI, MCP, and workspace-federated) appends the same note instead of blaming spelling alone.
- **`skeleton`** on an unsupported-language file says *"this file's language has no parser (.rs)"* instead of "no definitions indexed for this file".
- **MCP `instructions`** carry a coverage caveat scoping the "prefer these tools over grep/read" claim to the languages actually indexed; the static host/skill instruction templates name the skipped-language case too.

The classification of "code the symbol pass should have covered" is `CODE_EXTENSIONS` — the concept pass's own list — so the two passes can no longer silently disagree about what counts as code (`src/graph/coverage.ts`).

## Validation on the codebase from the issue

Same C++ game engine (350 first-party `.cpp`/`.h` + vendored SDL2/ENet, 1,671 files):

| | 0.9.0 | this branch |
|---|---|---|
| indexed files | 33 | **1,623** |
| nodes | 865 | **16,943** (10,647 function, 2,151 method, 2,018 struct, 439 enum, 65 class) |
| build banner | `[javascript, python]` | `[c/c++, javascript, python]` + `skipped: 48 files (no parser: .sh, .java, .swift)` |
| `grep snapEntityToFloor` | "no hits … All indexed code was searched" | **48 hits in 15 symbols across 12 files** |
| `callers snapEntityToFloor` | "no symbol — check spelling" | resolves both definitions + the honest no-edge-data note |
| `skeleton <file>.cpp` | "no definitions indexed" | full signature listing |

Cold build 24.5s, incremental rebuild 1.6s; every query (grep/skeleton/callers/map/ask) sub-second on the 17k-node graph. A 17 MB generated `parser.c` extracts without issue through the existing chunked-parse path.

## Tests

- `test/graph-cpp.test.ts` — extraction shapes: functions, inline/out-of-class/qualified methods with owners, structs/enums through namespace+template wrappers, prototype/forward-decl skipping, destructor/operator names, and the no-edges invariant.
- `test/graph-coverage.test.ts` — build skip-counting + `meta.unindexed` round-trip, the honest grep/callers/skeleton/MCP messages, workspace merging, and the "fully covered repo reports nothing" case.
- `test/graph-languages.test.ts` — the #36 one-table invariant extended over the C-family extensions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,37 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- **C/C++ in the Tier-1 symbol graph (definitions-only)** — `.c`, `.h`, `.cpp`,
+  `.cc`, `.cxx`, `.hpp`, `.hh`, `.hxx` are parsed with `tree-sitter-cpp`, so
+  `skeleton`, `grep` and `ask` now cover C/C++ repos: free functions, inline
+  and out-of-class methods (`Physics::step` → method with owner `Physics`),
+  classes, structs and enums, found through namespaces and template wrappers.
+  Prototypes and forward declarations are not definitions and are skipped.
+  Deliberately definitions-only: C++ has no import bindings (only `#include`,
+  namespaces, overloads, templates), so name-only call resolution would
+  fabricate edges between every same-named method in the repo — `callers` on a
+  C/C++ symbol says "no edge data … definitions-only" instead of answering
+  wrongly, and points at `graft grep` ([#66]).
+
+### Fixed
+
+- **Unsupported languages fail loudly instead of masquerading as answers.**
+  On a repo whose code has no Tier-1 parser, `grep`/`callers`/`skeleton`
+  returned confident false negatives ("no hits … all indexed code was
+  searched", "no symbol — check spelling", "no definitions indexed") that
+  steered callers away from raw grep, the only tool that would have worked.
+  Now: `graft build` prints `skipped: N files (no parser: …) — symbol tools
+  will not cover them` and records the gap in the graph (`meta.unindexed`);
+  zero-hit grep, unknown-symbol callers (single-repo and workspace) and
+  skeleton name the no-parser gap and point at raw grep/read; and the MCP
+  `instructions` scope the "prefer these tools over grep/read" claim to the
+  languages actually indexed ([#66]).
+
+[#66]: https://github.com/NanoNets/Graft/issues/66
+
 ## 0.9.0
 
 ### Added
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,20 @@
 {
   "name": "@nanonets/graft",
+<<<<<<< HEAD
   "version": "0.9.1",
+=======
+  "version": "0.9.0",
+>>>>>>> 8ef2fa4 (Add C/C++ to the Tier-1 extractor; make unsupported languages fail loudly)
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@nanonets/graft",
+<<<<<<< HEAD
       "version": "0.9.1",
+=======
+      "version": "0.9.0",
+>>>>>>> 8ef2fa4 (Add C/C++ to the Tier-1 extractor; make unsupported languages fail loudly)
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
@@ -16,6 +24,7 @@
         "gray-matter": "^4.0.3",
         "openai": "^6.45.0",
         "tree-sitter": "^0.21.1",
+        "tree-sitter-cpp": "^0.23.4",
         "tree-sitter-go": "^0.23.4",
         "tree-sitter-python": "^0.21.0",
         "tree-sitter-typescript": "^0.23.2"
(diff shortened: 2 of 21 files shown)


edrethardo:
Closing this — 0.10/0.11's breadth tier (`src/graph/generic.ts`) covers the C/C++ indexing this PR was for, via WASM grammars + tags.scm, and `resolveCInclude` covers the `#include` → file edges. No point reviewing a parallel implementation of something already shipped.

Two pieces here aren't in that tier, in case they're ever of interest:

- **Receiver typing for the C family.** `generic.ts` notes that `recvType` needs a per-language binding pass. A small one over declared syntax only — locals, parameters, class fields, plus `this->` and `Class::method` — is enough to resolve `obj.method()` / `ptr->method()` through the existing owner-qualified index, no LSP required. On a 350-file C++ engine it's the difference between one `callers` answer per name and one per receiver type.
- **Saying so when a language has no parser.** `graft build` reporting `skipped: N files (no parser: .x)`, plus grep/callers/skeleton phrasing an empty result as a coverage gap rather than a fact about the code. That was the more important half of #66 for agent use: an agent told "no hits, all indexed code was searched" concludes the symbol doesn't exist.

Both are on my fork if they're ever wanted; happy to open a focused PR for either. Thanks for building the thing.

Issue #68, opened 2026-08-07
Inferred call edges for C/C++ (follow-up to #66): unique-name + owner-qualified resolution, confidence-labeled


edrethardo:
Follow-up to #66 / #67, which added C/C++ as a **definitions-only** tier: `skeleton`/`grep`/`ask` work, but `callers` honestly reports "no edge data". This issue proposes graduating C/C++ to **inferred call edges**, using the resolution machinery graft already has.

## Why this is smaller than it sounds

`resolve.ts` already implements exactly the conservative resolver an import-binding-less language needs, and it's language-agnostic:

- `resolveName`: same-file match → `extracted`; **unique** cross-file match → `inferred`; ambiguous → **dropped, never guessed**.
- `resolveTypedMember`: member calls resolve only through an owner-qualified index (`Owner.method`) with a receiver type, walking `extends` chains, dropping on ambiguity.
- `EdgeV1.confidence: "extracted" | "inferred"` already models the trust level.

C/C++ was excluded from edge *capture* in #67 because naive capture would have fed name-only member calls into this machinery with no receiver types. The missing piece is C++-aware call-site capture, not a new resolver.

## Measured viability

On the C++ game engine from #66 (350 first-party `.cpp`/`.h` + vendored SDL2/ENet; 11,983 C/C++ function+method definitions after #67):

- **93% of distinct names are unique repo-wide** (9,418 of 10,091); 95% among first-party files.
- The ambiguous names the gate refuses are exactly where guessing would be worst: `main`×114, `init`×32, `loop`×30, `operator=`×29, `quit`×29.

So a unique-name gate yields real caller coverage for the overwhelming majority of symbols while structurally refusing the dangerous cases.

## Proposed capture rules (C/C++ `call_expression`)

1. **Bare calls** `foo(...)` → `{name, viaMember: false}` — resolved by `resolveName` (same-file → extracted, unique cross-file → inferred, else dropped).
2. **Qualified calls** `Physics::step(...)` → `{name: "step", viaMember: true, recvType: "Physics"}` — resolved via the existing `ownerMethod` index (#67 already populates `owner` for out-of-class definitions).
3. **`this->method(...)`** → `recvType` = enclosing class/struct (including the receiver class of an out-of-class definition body).
4. **`obj.method(...)` / `ptr->method(...)` with an untracked receiver** → captured `viaMember` **without** `recvType` → dropped by the existing `if (!e.recvType) continue`. No type inference, no guessing.
[...]

PR #69, opened 2026-08-09
feat: add C/C++ language support

kapelner:
## Summary

Adds C/C++ language support to graft's Tier-1 extraction, following the same pattern established for the other supported languages (Go, Python, TS). One `"cpp"` grammar (`tree-sitter-cpp`) parses the whole C/C++ family — `.c/.h/.cpp/.hpp/.cc/.cxx/.hh/.hxx` — uniformly, the same approach clangd and most polyglot tooling take. No separate `.c`-only language for v1 (documented limitation below).

- **Classes/structs/enums/templates** extracted directly (`class_specifier`/`struct_specifier`/`enum_specifier` all carry a `name` field; `template_declaration` spans are attributed back to the templated class/function so the card doesn't cut off the `template<typename T>` line).
- **Functions/methods** need a dedicated `describeCpp()`: `function_definition` carries no `name` field at all — it's buried in a declarator chain that can be a plain identifier, an out-of-class `Class::method` qualifier, a destructor (`~Class`), or an operator overload (`operator==`). A header-declared prototype and its `.cpp` out-of-line definition resolve to exactly **one** node, not two or zero — the single most important case to get right for a header/source-split codebase.
- **Visibility is stateful**: an `access_specifier` token (`public:`/`private:`/`protected:`) applies to every subsequent class member until the next one (default `private` for `class`, `public` for `struct`) — unlike every other supported language's per-node exported check, which needed new plumbing (`ctx.cppAccess` threaded through the walk).
- **Heritage** (`: public Base`) always emits `extends` — C++ has no `interface` keyword, so no "first = extends, rest = implements" heuristic is needed (simpler than the analogous C#-style precedent this was modeled on).
- **Calls**: member (`.`/`->`), qualified (`Class::static_method`, `ns::fn`) resolved via a direct `recvType` supply (bypassing the bindings-lookup path, since the qualifier already IS the type name).
- **`#include` edges** capture both `"local.h"` and `<system>` forms.

### Known limitations (documented in code + CHANGELOG)
- Macros that expand to declarations aren't understood — same class of limitation as any AST-only tool.
- Template specializations are treated as ordinary functions/methods by name (no specialization identity modeled).
- A bareword `#include "local.h"` (no `./` prefix) only resolves to an in-repo file when it happens to match a real repo-relative path — C's directory-relative-without-prefix include convention isn't specially handled by `resolve.ts`'s existing import resolver.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `npm run build` clean
- [x] `test/graph-cpp.test.ts` — 7 tests covering classes/structs/enums/templates, the header/source-split single-node guarantee, access-specifier visibility, heritage, member/qualified calls, `#include` edges, and contains edges
- [x] Full existing test suite green, no regressions
- [x] Smoke-tested via `graft build` against a real header/source-split C++ fixture


shhdwi:
Reviewed on top of current `main` — holding for an author rebase. main already has C/C++ via the **breadth tier** plus the `extractIncludes` include edges (#68). This depth extractor conflicts in `extract.ts`/`resolve.ts`/`hooks.ts` and, beyond the textual conflicts, needs two design decisions: (1) remove the breadth `.c/.h/.cpp` rows from `generic.ts` so depth doesn't collide, and (2) its `preferDefinition` prototype/definition dedup conflicts with main's new unique-match-only same-file resolve rule. A valuable depth upgrade, but author-level rebase work. Happy to re-review once rebased.

PR #83, opened 2026-08-12
feat: promote Java to the full-fidelity tier (hand-written extractor)

dbianco:
## Promote Java from the broad tier to full fidelity

Rebased on `main`, so this now sits on top of the generic breadth tier (#84). The framing has changed with it: this is no longer "add Java" — the breadth tier already parses it. It's a proposal that **Java is worth a hand-written extractor**, and the case for that is measurable.

The two tiers compose without any change to dispatch: `build.ts` prefers a hand-written extractor and falls back to generic only when there is none, so `.java` reaches this path and `queries/java.scm` is simply not consulted for it. I've deliberately **left `java.scm` in place** — removing another contributor's file is the maintainers' call, not this PR's.

### What the tiers actually produce

Same repo, same commit, same machine — [`google/gson`](https://github.com/google/gson) at `--depth 1`, 264 `.java` files. The only difference is whether `.java` is claimed by `EXTENSIONS` (this branch) or falls through to `genericLangOf` (`main`):

| | broad tier (`main`) | this branch |
|---|---|---|
| nodes | 4,042 | 4,446 |
| **`contains`** | **0** | 4,182 |
| **`imports`** | **0** | 2,602 (1,021 resolved in-repo) |
| **`extends`** | **0** | 234 |
| **`implements`** | **0** | 196 |
| `calls` | 2,716 | 4,995 |
| total edges | 2,716 | 12,209 |
| records as `struct` | 0 | 33 |
| enums as `enum` | 0 | 26 |

The headline isn't the call count. It's that on the broad tier **every edge is a `calls` edge** — there is no containment, no import graph, and no type hierarchy at all. `graft ask "what implements this interface"` has nothing to answer from; neither does "what does this file depend on".

That isn't a criticism of the breadth tier, which is doing what it says: symbols plus name-resolved calls, one grammar per language, at a fraction of the effort. It's the argument for why a language you actually work in benefits from the expensive path.

### What the hand-written extractor adds

Three things the generic tier structurally cannot do, because `tags.scm` has no scope model:

**1. Typed receivers.** `store.save(x)` resolves through the declared type of the field `private final Store store`, so the edge lands on `Store.save` and not on every `save` in the repo. Fields and parameters are bound per scope; dependency-injected collaborators — the edges worth having in a service codebase — are exactly the typed-final-field case.

**2. Package-path imports.** A Java import names a *type* and states no source root, so resolution matches the FQN against a path suffix (`com.acme.Foo` → `**/com/acme/Foo.java`). Root-agnostic, no build-file parsing, so it stays deterministic and dependency-free. A static-member import retries its enclosing type; a wildcard, or a suffix shared by two files, stays unresolved rather than inventing an edge.

**3. Overload disambiguation.** Method and constructor nodes record `arity` and `variadic`; call edges record the call site's `argCount`. Without it, a convenience overload delegating to the fuller one resolves to *itself* — see the [comment below](#issuecomment-5268369387) for how that defect surfaced and was fixed.

### Java-specific things that bite

Worth flagging for review, since each one is invisible to the TS/Python/Go fixtures:

- `method_invocation` has **no `function` field** — it splits the callee into `object` + `name`. The shared `calleeName()` returns null for every Java call site without an early branch.
- Java has **no free functions**, so an implicit-`this` call (`decorate(x)`) is a method call. Resolved against the function index it vanishes — and it's the commonest intra-class edge in the language.
- `new Foo()` targets a **type**, not a function.
- A record maps to `struct` rather than `class`, so a DTO and a service stay distinguishable where DTOs are most of the type surface.

### Known limits (tested, not papered over)

- A `var` local states no type, so a member call through one stays unresolved. Cross-file return-type inference is out of scope for a deterministic single-file pass. Fields and parameters are explicitly typed even in `var`-heavy code and carry most of the graph; the constructor edge still lands, so the dependency stays visible.
- Same-arity overloads differing only by parameter *type* (`render(String)` / `render(int)`) can't be separated by count. Narrowing declines to choose rather than guessing.

Both are pinned by tests so a future change to either is deliberate.

### Not covered

Annotations aren't extracted, so frameworks that express routing or persistence through them (JPA `@Entity`/`@Column`, and similar) keep that topology invisible. Happy to follow up separately if it's of interest.

### Tests

11 Java cases in `test/graph-java.test.ts` — node kinds, modifier-based visibility, the three call-resolution paths, heritage, package-path and static-member imports, ambiguous-suffix and `var` behavior, and three overload cases. `.java` added to the language-label matrix. README moves Java from Broad to Full-fidelity, since that's what this branch makes true.

Full suite: **618 passing** (this branch's cases plus upstream's new ones). Every new test was verified to fail when the behavior it covers is reverted, so none of them are vacuous.

---

If the preference is to keep Java on the breadth tier and hold the line on hand-written extractors, that's a reasonable call and I won't argue it further — the measurements above are the whole of my case. I'd also happily split the overload fix into its own PR if you'd rather review the language support and the resolver change separately.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
         "openai": "^6.45.0",
         "tree-sitter": "^0.21.1",
         "tree-sitter-go": "^0.23.4",
+        "tree-sitter-java": "^0.23.5",
         "tree-sitter-python": "^0.21.0",
         "tree-sitter-typescript": "^0.23.2"
       },
@@ -859,6 +860,25 @@
       "license": "MIT",
       "dependencies": {
         "node-addon-api": "^8.2.1",
+        "node-gyp-build": "^4.8.2"
+      },
+      "peerDependencies": {
+        "tree-sitter": "^0.21.1"
+      },
+      "peerDependenciesMeta": {
+        "tree-sitter": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/tree-sitter-java": {
+      "version": "0.23.5",
+      "resolved": "https://registry.npmjs.org/tree-sitter-java/-/tree-sitter-java-0.23.5.tgz",
+      "integrity": "sha512-Yju7oQ0Xx7GcUT01mUglPP+bYfvqjNCGdxqigTnew9nLGoII42PNVP3bHrYeMxswiCRM0yubWmN5qk+zsg0zMA==",
+      "hasInstallScript": true,
+      "license": "MIT",
+      "dependencies": {
+        "node-addon-api": "^8.2.2",
         "node-gyp-build": "^4.8.2"
       },
       "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "openai": "^6.45.0",
     "tree-sitter": "^0.21.1",
     "tree-sitter-go": "^0.23.4",
+    "tree-sitter-java": "^0.23.5",
     "tree-sitter-python": "^0.21.0",
     "tree-sitter-typescript": "^0.23.2"
   },
@@ -72,5 +73,13 @@
     "esbuild": "^0.28.1",
     "tsx": "^4.22.4",
     "typescript": "^6.0.3"
+  },
+  "allowScripts": {
+    "tree-sitter@0.21.1": true,
+    "tree-sitter-typescript@0.23.2": true,
+    "tree-sitter-python@0.21.0": true,
+    "tree-sitter-go@0.23.4": true,
+    "tree-sitter-javascript@0.23.1": true,
+    "tree-sitter-java@0.23.5": true
   }
 }
--- a/src/graph/bindings.ts
+++ b/src/graph/bindings.ts
@@ -43,6 +43,9 @@ const FN_VALUE_TYPES = new Set(["arrow_function", "function", "function_expressi
  * same scope key extract.ts's walk will look it up with), or null if `node`
  * isn't a definition. */
 export function defName(node: Parser.SyntaxNode, lang: Language): string | null {
+  if (lang === "java") {
+    return JAVA_DEF_TYPES.has(node.type) ? (node.childForFieldName("name")?.text ?? null) : null;
+  }
   if (lang === "go") {
     if (node.type === "method_declaration") {
       const name = node.childForFieldName("name")?.text;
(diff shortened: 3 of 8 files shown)


dbianco:
Pushed a second commit that fixes a defect in the first one. Flagging it explicitly rather than letting it hide in the diff, since the first commit shipped a confidently-wrong edge.

### The defect

Overload resolution matched on name + owner only. With two same-named methods on one class, the existing same-file tiebreak picked the first candidate — so the commonest shape in Java, a convenience overload delegating to the fuller one:

```java
public String join(String left, String right) {
    return join(left, right, "-");   // the 3-arg overload
}
```

produced a `calls` edge from the 2-arg method **to itself**, marked `extracted` — the highest confidence level. A self-loop in a call graph, asserted as fact.

### Why the fixtures missed it

None of the other languages here have an overload *set* of implementations: Python and Go have no overloading, and TypeScript's overload signatures resolve to a single body. So there was no existing fixture shape that could have caught this, and the Java tests I wrote in the first commit all passed.

It surfaced only from running the branch against a real Java codebase and noticing a `callers` result that pointed at the method I was already looking at. Worth knowing if you're reviewing the first commit in isolation — it looks correct, and the unit tests agree with it.

### The fix

Method and constructor nodes now record `arity` (declared parameter count) and `variadic`; Java call edges record the call site's `argCount`. The owner-qualified resolver narrows an overload set to the candidates that count could reach, *before* the existing same-file / ambiguous logic runs.

Conservative in four directions so nothing that already resolved can regress:

- a variadic candidate accepts `arity - 1` upward and is never excluded by count
- a candidate with no recorded arity is kept — older graphs and other languages, where absence of data is not evidence of mismatch
- if narrowing leaves nothing, the **original** set is returned, so prior behavior stands rather than dropping a real edge
- only Java emits either field, so every other language takes the identity path

Both new fields are optional additions to `NodeV1`, so graphs built before this still load and resolve exactly as they did.

### Known limit

Same-arity overloads differing only by parameter *type* — `render(String)` vs `render(int)` — cannot be separated by count. Narrowing declines to choose and leaves the previous behavior rather than guessing at types. There's a test pinning this, so a future type-aware pass is a deliberate change rather than an accident.

### Tests

3 new cases: delegation resolves to the other overload rather than itself; a variadic candidate survives an argument-count filter; a same-arity pair stays unseparated. Each was verified to fail when the fix is reverted. Full suite: 606 passing.

Happy to split this into its own PR if you'd rather review the language support and the resolver change separately — say the word and I'll restructure.

dbianco:
Follow-up on the overload fix above, because the defect turned out to have a downstream effect I didn't expect — and it's an argument for graph correctness generally, not just for this PR.

I'd run a question through an agent against a real Java codebase **before** the arity fix, and asked it where a piece of functionality was implemented and what called it. The answer was substantially correct, but two of its cited line numbers were wrong — off by four and five lines.

Those two citations were the **delegating overload pair**. Every other citation in the same answer — controllers, routes, wiring, a helper on another class — was exact, copied faithfully from the graph.

Re-running the same question on the same codebase after the fix, with nothing else changed: **every citation exact, including the overload pair.** A route reference that had been off by one also came out right.

The reading I'd offer is that a wrong edge doesn't sit inert. `callers` on the affected symbol returned a self-loop — a method calling itself — which is visibly nonsense, so the consumer stopped trusting the graph *at that point*, went and derived the numbers by hand from source, and got them wrong. The graph's own spans for those two methods were correct the whole time; they just weren't used.

So the cost of that edge wasn't one bad edge. It was the agent silently falling back to a worse method exactly where the graph looked unreliable, and being confidently wrong there — in the one place a reviewer would be least likely to double-check, since everything around it was right.

Worth noting for calibration: the fix improved *precision*, not cost. Same number of tool calls before and after. The value was in the answers being right, not in getting there faster.

I can't share the codebase, but the effect should reproduce on any Java project with a convenience overload delegating to a fuller one — which is most of them. The `join(left, right)` / `join(left, right, separator)` fixture in `test/graph-java.test.ts` is the minimal shape of it.

shhdwi:
Merged. Thank you, this is excellent work. 🙏

You didn't just add a language, you made the case for one and backed it with a same-repo, same-commit measurement. The `google/gson` table reproduced precisely on our end (12,209 edges, 0% orphan symbols, records to `struct`, enums to `enum`).

You also clearly internalized the philosophy this project cares about most: resolve precisely or drop, never guess. Typed receivers through declared field/param types, imports staying unresolved on an ambiguous suffix, `narrowByArity` declining on same-arity overloads, variadics never filtered. The spot-checks confirmed it (2,203 cross-file method calls, all owner-qualified and correct).

Small things that showed real care: flagging the Java-specific traps the TS/Python/Go fixtures can't catch, being honest about the limits instead of papering over them, verifying each new test fails when reverted, and leaving `java.scm` in place because removing another contributor's file wasn't yours to decide. 619 tests green.

The annotation-extraction follow-up you offered would be very welcome as a separate PR if you're up for it. Really appreciate the rigor here.

PR #96, opened 2026-08-13
Breadth tier: file-level import edges for C/C++ (#include), Rust (use), and PHP (use)

shhdwi:
## What

The breadth tier captured definitions, calls, and references — but not the preprocessor / module imports. A codebase's dependency graph (its `#include` / `use` graph) was invisible. This adds it for the two languages where it resolves to a *file* cleanly and precision-first:

### C/C++ `#include "x.h"`
- **generic.ts** walks `preproc_include`; **local** includes become `imports` raw edges, **system** `<stdio.h>` includes are skipped (high volume, no in-repo target).
- **resolve.ts** `resolveCInclude`: relative-to-includer first (certain), else a **unique path-suffix** match (an `-I`-reached header). Ambiguous/not-found stays a string — never guessed.

### Rust `use crate::…`
- **generic.ts** walks `use_declaration`; in-crate `crate::` paths become `imports` raw edges (std/super/self/external crates and globs skipped).
- **resolve.ts** `resolveRustUse` settles the path against the file's **own crate root** (the `lib.rs`/`main.rs` dir — workspace-safe) by **longest module prefix**: `crate::a::b::C` → module `a/b`; `crate::lexical` → `lexical.rs` or `lexical/mod.rs`. Two guardrails found via a real multi-crate run: the crate-root (`lib.rs`) is a target only for a single-segment `use crate::Item`; and a file with **no owning crate root** (an integration test under `tests/`, whose `crate::` is the *test binary*, not the lib) is never resolved across the workspace.

## Validation (safe, static — no code executed)

Crater-style breadth, done safely: graft only *parses* source, so cloning + `graft build` over real ecosystem crates is inert (unlike `cargo build`/`test`).

**Rust, 5 crates** (anyhow, clap, regex, ripgrep, serde_json): **every library `src/` `use crate::` edge resolves, 0 cross-crate false edges**, invariants pass. The only unresolved are integration-test `crate::` paths — correct Rust semantics (a test binary's own root).

**Cross-language matrix, 11 repos / 8 languages** — all pass `--strict` invariants, all builds byte-identical on rebuild:

| repo | lang | tier | invariants |
|---|---|---|---|
| gson, JSON-java | Java | depth | ✓ |
| hiredis | C | breadth + include | ✓ |
| fmt | C++ | breadth + include | ✓ |
| serde_json, clap, ripgrep | Rust | breadth + use | ✓ |
| FastRoute | PHP | breadth + references | ✓ |
| cobra | Go | depth | ✓ |
| flask | Python | depth | ✓ |
| sinatra | Ruby | breadth | ✓ |

hiredis: 82 include edges, 79 resolved (96%). serde_json: 38 use edges, 100% resolved.

## Scope / honesty

Symbol-orphan is unchanged by design — imports are *file*-level. The win is dependency navigation ("what includes/uses this"). Deliberately C/C++ + Rust only: those resolve to a file. PHP/other `use` name namespaces (PSR-4) — lower precision, separate work.

**625 tests** (+2: C include resolve/skip/ambiguous-drop; Rust longest-prefix/single-segment/skip/unresolved/test-binary-drop).

### DIFF
--- a/src/graph/generic.ts
+++ b/src/graph/generic.ts
@@ -216,9 +216,45 @@ export function extractGeneric(rel: string, source: string, langName: string): E
     walkExtract(tree.rootNode as TsNode, mkDef); // no tags.scm → symbols only
   }
   // The preprocessor is invisible to tags.scm, but in C/C++ a local `#include "x.h"`
-  // IS the dependency graph — capture it as a file→file import.
+  // IS the dependency graph — capture it as a file→file import. Likewise a Rust
+  // `use crate::…` is an in-crate module dependency.
   if (langName === "c" || langName === "cpp") extractIncludes(tree.rootNode as TsNode, rel, rawEdges);
+  else if (langName === "rust") extractUses(tree.rootNode as TsNode, rel, rawEdges);
   return { nodes, rawEdges };
+}
+
+/** Rust `use crate::a::b::Item` → a file→module `imports` raw edge whose specifier is the
+ * crate-relative module path (`a/b`). Only in-crate imports are captured; `std::`,
+ * `super::`, `self::`, external crates, and globs are skipped — resolve.ts settles the
+ * path against the file's crate root, and drops it when it can't. */
+function extractUses(root: TsNode, rel: string, rawEdges: RawEdge[]): void {
+  const visit = (n: TsNode): void => {
+    if (n.type === "use_declaration") {
+      const spec = rustUseModule(n.text);
+      if (spec !== null) rawEdges.push({ source: rel, relation: "imports", specifier: spec ? `crate/${spec}` : "crate", file: rel });
+    }
+    for (let i = 0; i < (n.namedChildCount ?? 0); i++) {
+      const c = n.namedChild?.(i);
+      if (c) visit(c);
+    }
+  };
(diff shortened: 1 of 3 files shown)


shhdwi:
Added **PHP `use`** (third commit) — completes the breadth-import trio.

- `generic.ts` walks `namespace_use_declaration` (groups `{A, B}` expand; `as` aliases stripped; `use function`/`use const` skipped — those name a symbol, not a PSR-4 class file).
- `resolve.ts` `resolvePhpUse` settles the FQN to the class file by the **longest namespace-tail suffix** (`App/Models/User.php` → `Models/User.php` → `User.php`), since the PSR-4 root is unknown; ambiguous tail or vendor class stays a string.

**Validated:** PHP-Parser **738/753 resolved (98%)** — the rest are vendor/global (`PHPUnit\…`, `InvalidArgumentException`); FastRoute **100% of in-repo `use FastRoute\*`** (all unresolved are external vendor). Invariants pass. **626 tests.**

PR now covers **C/C++ `#include`, Rust `use`, PHP `use`** — file-level dependency edges, precision-first (drop-rather-than-guess), each validated on real ecosystem repos.

PR #124, opened 2026-08-16
fix(init): honor --no-mcp and --no-hooks for Claude Code

### LINKED ISSUE #118: graft init --no-mcp does not work for Claude Code
Using the command `graft init --no-mcp` or `--no-hooks`  configures MCP server or installs hooks nevertheless.
The parameters are not even used in `runInit` for Claude Code:

```
if (wantClaude) {
    const res = runInit(repo, {
        build: opts.build,
        cliPath
    });
    ...
}
```

Would be nice to wire the parameters for Claude Code Installation, to. 
Reason: Using Claude Code with a DeepSeek LLM otherwise produces much more LLM cache missed when using MCP instead of calling graft via CLI.

michaelxer:
Fixes #118.

## Problem
`graft init --no-mcp` (and `--no-hooks`) only reached other-agent wiring. Claude Code still wrote `.mcp.json` and the hook shim/settings, even though the README says `--no-mcp` skips Claude Code MCP registration.

That's the path that burns extra LLM cache when Claude Code is pointed at a DeepSeek endpoint.

## Fix
- `runInit` honors `mcp` / `hooks` the same way `runHostsInit` already did
- `--dry-run` / the picker drop those Claude paths
- a later wiring refresh replays the stamp so an upgrade does not put them back
- default `graft init` is unchanged

## How I tested
```
node --import tsx --test test/claude-init.test.ts test/claude-settings-merge.test.ts test/hosts-init.test.ts test/hosts-plan.test.ts test/upkeep.test.ts
```
80 passed.

Local repro after the change:
```
graft init <tmp> --no-build --agents claude --no-mcp --no-hooks
```
writes statusline + skill, no `.mcp.json`, no `graft-hooks.cjs`. Default `--agents claude` still writes both.

Did not run a live Claude Code session against DeepSeek.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- **`graft init --no-mcp` / `--no-hooks` were ignored for Claude Code.** Those
+  flags only reached `runHostsInit` (Cursor/Codex/…). `runInit` always wrote
+  `.mcp.json` and the Claude hook shim/settings — including the user-level
+  copies under `~/.claude`. They now skip the matching Claude Code writes too,
+  `--dry-run` / the picker drop those paths, and a later wiring refresh
+  replays the stamp so an upgrade does not put them back.
+
 ## 0.16.0
 
 ### Added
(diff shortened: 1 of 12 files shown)


Christian-B-Koller:
Appreciate your work 👍

PR #130, opened 2026-08-17
feat(graph): full-fidelity Kotlin extraction

shovelmn12:
## Problem
Kotlin .kt/.kts files were handled by the generic breadth tier — a single language-agnostic tags.scm query over a WASM grammar. That tier is deliberately flat: it captures definitions and bare call names, but has no scope-aware resolution, so Kotlin symbols, calls, heritage, and imports got none of the treatment TypeScript, Python, Go, and Java receive on the depth tier. Concretely: class/interface/enum/annotation were indistinguishable, member calls resolved only as bare names (or not at all), :-clause supertypes produced no extends edges, and visibility was unmodeled.
There was no drop-in fix: tree-sitter-kotlin has no per-kind node types and exposes no name/body fields, so the depth tier's standard shape (one mapped node type with named fields) could not be used off the shelf.

## Fix
Add a hand-written Kotlin extractor (describeKotlin) that reads the real tree-sitter-kotlin grammar, plus the routing to run .kt/.kts on the depth tier:
- Definitions — class_declaration is rekinded off its own keyword into class / interface / enum / annotation; object/companion object become classes; secondary constructors and member functions become methods; typealias becomes a type; top-level val/var become variables. Since the grammar exposes no name/body fields, names are read off the first unnamed type_identifier/simple_identifier child and header ends off the class_body/function_body/statements children.
- Calls — edges via call_expression; member calls through navigation_expression with this/super receiver handling (plus the implicit-this form), routed through owner-qualified resolution.
- Heritage & imports — extends edges from the : clause (the head type of each delegation_specifier), import edges from import_header.
- Visibility — internal/private/protected map to the exported flag (public is implicit, so exported by default).
- Adds tree-sitter-kotlin as a dependency (native grammar, no prebuilds — compiled via node-gyp on install); removes Kotlin from the generic breadth registry.

## Verification
- Ran extraction on a 162-file Spring Boot Kotlin repo: 1066 nodes / 2401 edges, 0 extraction errors.
- Updated test/graph-languages.test.ts and test/supported-extensions.test.ts; CHANGELOG entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## 0.12.0
+
+### Added
+
+- **Kotlin moves to full-fidelity extraction.** `.kt` and `.kts` files are now
+  parsed by a hand-written tree-sitter extractor — the same tier as TypeScript,
+  Python, Go, and Java — instead of the generic breadth grammar, so Kotlin
+  symbols, call edges, heritage, and imports resolve with scope awareness. The
+  kind mapping now matches tree-sitter-kotlin's real node types (the earlier
+  attempt reused Java's, which do not exist in the Kotlin grammar and emitted no
+  symbols at all): `class_declaration` is re-read off its own keyword into
+  class / interface / enum / annotation, `object` and `companion object` become
+  classes, secondary constructors and member functions become methods,
+  `typealias` becomes a type, and top-level `val`/`var` become variables.
+- **Kotlin edges.** Calls resolve through `call_expression` (member calls via
+  `navigation_expression`, with `this`/`super` receivers), the `:` heritage
+  clause yields `extends` edges, `import_header` yields import edges, and
+  `internal`/`private`/`protected` visibility maps to the exported flag.
+
 ## 0.11.0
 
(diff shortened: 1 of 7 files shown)


shhdwi:
Verified locally on top of current `main`. Rebased 5 conflicts, all additive: version→0.12.1, CHANGELOG (kept both Added/Changed), `extract.ts` (Kotlin heritage `kotlinTypeDecl` + the PHP attribute-edges line coexist), `supported-extensions.test.ts` (`.php`+`.kt`+`.kts` in depth), and the lockfile regenerated with `tree-sitter-kotlin@0.3.8`.

Depth-tier (full-fidelity) Kotlin: on a Sample.kt with a class/interface/override + top-level funcs, `build` produced 8 nodes, no garbage; `skeleton` shows correct signatures + `: Greeter` heritage; `extends Welcomer→Greeter` resolves; `callers makeWelcome → main` resolves. One new standard native dep (`tree-sitter-kotlin`, already in `allowScripts`), no committed binaries. **Full suite 882/882.** Merging.

Issue #127, opened 2026-08-17
deep build swallows per-file LLM failures (incl. 429) and exits successfully


Rajahn:
During `graft build --deep` against an 884-file Java repo behind a gateway with a hard token quota, once the quota was exhausted every remaining per-file summary call failed with HTTP 429:

- The run logged **1,617 `✗ ... 429 ... exceed quota` lines**, one per failing file.
- It then printed the normal success footer (`graft/ is git-ignored (added automatically) ...`) and **exited 0**.
- `graft check` afterwards reports `OK — the graph is in sync with the code`, with no indication that the meaning tier is mostly missing.

There is also no retry/backoff on 429, so a transient rate limit and a permanent quota exhaustion behave identically: every affected file is silently dropped.

The consequence: the user only discovers the degraded graph days later via poor `ask` results, with nothing in the exit code, the log footer, or `graft check` connecting cause and effect.

Request: persistent errors like quota exhaustion should fail the build loudly (non-zero exit + clear message) instead of degrading file by file, and some 429 backoff/retry would avoid dropping files on transient limits.

Environment: graft 0.10.1 (`@nanonets/graft`), Node 22, OpenAI-compatible corporate gateway, model deepseek-v3.2, full-repo `--deep` (~3M tokens required vs 500k quota).

shhdwi:
The two worst parts here are covered by @christian-jorge's #111: retry/backoff on 429 (honoring the endpoint's Retry-After, tunable via `GRAFT_MAX_RETRIES`) so a transient rate limit no longer behaves like permanent quota exhaustion, and the fix for a failed round erasing summaries/the concept layer. The remaining piece your report nails is observability: even with retries, a run that exhausts quota should not exit 0, and `graft check` should not say "in sync" when the meaning tier is largely missing. That surfacing (a non-zero exit or a clear degraded-tier warning, plus a check signal) is the still-open part. Keeping open for it.

PR #132, opened 2026-08-17
Checkpoint crux to disk during --deep, atomically (#128)

### LINKED ISSUE #128: deep tier: crux results persisted only at build end; an interrupted run loses all of them
During `graft build --deep`, per-file summaries are flushed to `graft/.cache/summaries.json` as they complete, but crux/meaning results appear to reach `graft/.graph/wiring.json` only when the build finishes.

Observed on an 884-file repo: a `--deep -j 4` run was killed (SIGTERM) after ~2 hours, with the crux pass at roughly 606/879 files:

- `summaries.json` survived — those summaries replayed from cache on the next run, as designed.
- `wiring.json` mtime was unchanged since the structural build: **zero crux results survived**, even though every crux call up to that point had succeeded and been paid for. The next run re-issued all of them from scratch.

A deep pass on a large repo realistically takes 1–3 hours, so any interruption (timeout, laptop sleep, CI restart) discards the entire pass.

Request: give the crux pass the same durability the summaries cache already has — periodic checkpointing so an interrupted `--deep` can resume instead of restarting.

Environment: graft 0.10.1, Node 22, OpenAI-compatible gateway, 884 Java files.

shhdwi:
Fixes #128.

## Problem
Per-file summaries checkpoint to the cache as they complete, but crux/meaning reached `wiring.json` only at build **end** (one `writeGraph`). A `--deep` run killed partway (SIGTERM / CI timeout / laptop sleep — realistically 1-3h on a large repo) discarded **every crux it had already computed and paid for**, and the next run re-issued all of them. Observed by @Rajahn: an 884-file run killed at ~606/879 lost the entire crux pass.

## Fix
Reuses the existing mechanism where `wiring.json` *is* the Tier-2 crux cache (folded back in on the next build, keyed by `body_hash`):
- **enrich.ts** — the crux pass calls an optional `checkpoint` from its per-file completion handler, throttled to ~15s (seam `GRAFT_CRUX_CHECKPOINT_MS`).
- **build.ts** — assemble the graph *before* the meaning pass and pass `checkpoint: () => writeGraph(graph, outDir)`. crux/summary mutate nodes in place and never change the node/edge set, so `meta` stays valid; the opt-in LSP pass still runs before the final write.
- **write.ts** — `writeGraph` is now **atomic** (temp + `rename`). A periodic flush is exactly what gets killed mid-write, and a partial `writeFileSync` would leave a truncated, unparseable graph.

Net: an interrupted `--deep` run keeps its crux; the next run replays it by `body_hash` instead of repaying it.

## Tests
**688 pass** (+2, via a fake summarizer, no real model): partial crux is flushed mid-pass (checkpoint), and a rerun replays the completed files instead of re-issuing them. Real-repo build sanity-checked (valid graph, no `.tmp` leak). `graph-refresh`'s read-only-graph case updated: an atomic rename can replace a read-only *file*, so the unwritable target is now the *directory* (the realistic case, and the atomicity is worth it).

This is the durability gap #111 does not cover; #127/#129 (retry/backoff, content-fallback parsing) are addressed there.

### DIFF
--- a/src/graph/build.ts
+++ b/src/graph/build.ts
@@ -283,22 +283,14 @@ export async function buildGraph(
 
   const edges = resolveEdges(nodes, rawEdges, { goModules: readGoModules(root, repoFiles) });
 
-  // graph.json is its own Tier-2 cache: fold in the prior meaning layer so an
-  // unchanged body is never re-summarized (and a Tier-1-only run never wipes it).
-  const prior = readGraph(wiringPath(outDir));
-  const priorById = new Map((prior?.nodes ?? []).map((n) => [n.id, n]));
-  const meaning = await enrichGraph(nodes, priorById, sources, {
-    summarizer: opts.summarizer,
-    concurrency: opts.concurrency,
-    onProgress: ({ index, total, node }) =>
-      opts.onProgress?.({ phase: "enrich", index, total, file: node }),
-  });
-  errors.push(...meaning.errors);
-
   // Guard 5 (minimum-substance): node counts aren't known until nodes are
   // assembled, so the merge-tiny-scopes-into-root guard runs here.
   const scopes = applyMinSubstanceGuard(discoveredScopes, nodes);
 
+  // Assemble the graph BEFORE the meaning pass, so the crux pass can checkpoint it to
+  // disk periodically (#128): crux/summary mutate node objects in place and never
+  // change the node/edge SET, so `meta` stays valid; the opt-in LSP pass below is the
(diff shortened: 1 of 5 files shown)


anirudhkumar-nanonets:
Superseded by #151, which cherry-picked this fix (checkpoint the crux to disk during the deep pass, atomically) and shipped it alongside the blast-radius command. Thanks @shhdwi — the commit is on main as part of 42c4991.

PR #135, opened 2026-08-18
feat: add Emacs Lisp language support

timotheosh:
Registers .el files as an indexed language and provides a tree-sitter
    query that captures defun, defmacro, defvar, defconst, defcustom,
    defvar-local, defvar-keymap, defface definitions, and call-site
    references.


shhdwi:
Re-checked on top of current `main` — the over-generation issue from the earlier review is **still present** in the current head, so holding.

The reference query is unchanged: `(list . (symbol) @name) @reference.call` tags every list head as a call. Probe (`sample.el`): a `(defun prefix …)` that is never called, plus a `(let ((prefix "/usr") …) …)` inside `build-path`. Result — the query mints call references for parameter names and `let`-binding names; most die at resolution, but **`graft callers prefix` returns a real, spurious `calls ← build-path` edge** (the exact collision flagged before). Genuine calls do work (`callers helper → greet`).

Also still ships **no test** (only `generic.ts` + `elisp.scm` changed).

To land: gate the reference match so binding heads (arg lists, `let`/`let*`/`cond` bindings) aren't treated as calls, and add a `SNIPPETS` case in `generic-extract.test.ts` (including a negative assertion that a let-bound name isn't called). Happy to re-review once that's in.

timotheosh:
Pushed a fix. The over-generation was real — `(list . (symbol) @name) @reference.call` can't tell a call site from a binding head, since elisp uses the same `list` node for both.

Added a `@reference.call.ignore` capture to gate it: a query can mark a token at the same offset a `@reference.call` match would use, and generic.ts drops that match before it becomes an edge (same mechanism already used for `defNameAt` to stop a definition's own name token from self-looping). elisp.scm now gates three shapes off from being calls — `defun`/`defmacro` parameter names, `let`/`let*` binding heads, and bare `cond` clause tests.

Verified against your `prefix`/`build-path` probe: `graft callers prefix` now returns no edges, and `callers helper → greet` still resolves.

Added a regression test (`generic-extract.test.ts`) that reproduces the exact collision plus the `cond` `t`-clause case, with both a negative assertion (no spurious caller) and a positive one (real calls still resolve) — so this can't silently regress.

timotheosh:
The two failing checks (`test (ubuntu-latest)`, `test (windows-latest)`) are pre-existing on `main` — not caused by this change. `main` @ `ee1ef03` (the exact commit this branch merged in) fails CI identically: https://github.com/NanoNets/Graft/actions/runs/32722784307

- `test (ubuntu-latest)`: `viz --tabs: a bad tab name fails loudly...` (`test/viz-export.test.ts`) fails with `✗ no context graph at .../graft — run \`graft build --deep\` first` instead of the expected validation message — looks like a test-isolation issue unrelated to elisp.
- `test (windows-latest)`: `npm ci` itself fails before any tests run — `tree-sitter-kotlin` has no prebuilt Windows binary and `node-gyp` can't find a usable Visual Studio toolchain on the runner.

Neither touches the elisp code path. Leaving both out of scope here since they need fixing against `main` independently.

PR #136, opened 2026-08-18
test: pin Node 24 WASM compatibility

### LINKED ISSUE #122: graft build OOMs with "Fatal process out of memory: Zone" on large monorepos (~2,900+ files); --max-old-space-size has no effect
**Env:** Windows 11, Node v24.16.0, graft 0.10.1 (latest published), 32 GB RAM.
**Repo:** public [`block/buzz`](https://github.com/block/buzz) (Rust + TS/Tauri desktop + Flutter mobile monorepo), 3,294 parseable files.

**Command:** `graft build .` (also via `graft init`).
**Result:** aborts with `# Fatal process out of memory: Zone` (`v8::base::FatalOOM`), exit 3, no `graft/` produced. Reproducible.

**Where it dies:** consistently at ~file 2895–2908 of 3294, within the `mobile/` Dart tree — but ~2,600 non-mobile files parse cleanly first, and the largest files involved are ordinary (~230 KB test files). So this reads as **aggregate memory growth across the run**, not one pathological file.

**Tried:**
- `NODE_OPTIONS=--max-old-space-size=12288` → still OOMs.
- `-e .rs` → still exits 3.
- **Works — scoping to a subtree:** `graft build crates` (348 files) → 12,932 nodes / 18,739 edges. `graft build desktop` (2,417 files) → 20,070 nodes / 43,223 edges. Both succeed on the same machine, same version. The pipeline is fine; it just can't hold a full large-monorepo parse in memory.

That 2,417-file success next to a failure at ~2,900 is consistent with a per-file leak rather than a size cliff on any individual file.

**Possibly already diagnosed:** draft PR #108 states "the breadth tier allocated a `Parser` and a `Tree` per file and freed neither. Both live in the WASM heap, where V8 GC cannot reach them." That matches this failure mode exactly, and would also explain why `--max-old-space-size` has no effect — the V8 old-space limit doesn't govern the WASM heap.

**Questions / requests:**
1. Is #108 the same root cause? If so, this is a real-world repro to validate against — the leak is currently in the latest published release (0.10.1), so full-repo builds on large monorepos are broken for end users today.
2. Regardless of #108: consider **skip-and-continue** when a file would exceed a memory budget, rather than aborting the entire run — losing a 3,294-file build at file 2,900 is expensive.
3. A documented **chunk-and-merge** path for large repos would also help; scoping per subtree works today but produces several disconnected graphs instead of one, so cross-subtree edges are lost.

nsdevaraj:
## Summary

- add an isolated Node 24 regression probe that warms and exercises every breadth-tier WASM grammar
- add a focused Node 24 CI job without duplicating the full OS test matrix
- document that the `tree-sitter-wasm` bundle now avoids the V8 Turboshaft `Zone` OOM

## Context

The published 0.10.1 grammar bundle aborts under Node 24 with `Fatal process out of memory: Zone`; `--max-old-space-size` cannot help because the failure is in V8's WASM optimizing compiler. The broader `tree-sitter-wasm` bundle already merged on `main` makes the minimized workload and a full 459-file cold build pass. This PR pins that runtime behavior so a dependency change cannot silently reintroduce the crash.

The probe runs in a child process because this failure aborts V8 rather than throwing a catchable JavaScript exception.

## Verification

- `node --import tsx --test test/generic-node24.test.ts` — 1 passed on Node 24.19.0
- `npm run build` — passed
- `npm test` — 687 passed
- full cold build of the original Rust/Swift/TypeScript repro — 459/459 files parsed

Fixes #122

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,3 +41,24 @@ jobs:
 
       - name: Test
         run: npm test
+
+  node24-wasm:
+    name: Node 24 WASM regression
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+        with:
+          persist-credentials: false
+
+      - name: Setup Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version: '24'
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Exercise every breadth grammar
+        run: node --import tsx --test test/generic-node24.test.ts
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ### Fixed
 
+- **Node 24 no longer aborts breadth-tier builds with `Fatal process out of
+  memory: Zone`.** The broader `tree-sitter-wasm` grammar bundle avoids the V8
+  Turboshaft failure triggered by the previous bundle. CI now exercises every
+  breadth grammar under Node 24 to keep that runtime compatibility pinned
+  ([#122]).
 - **`npm install -g @nanonets/graft@latest` could silently do nothing.** The
   generated shims (`.claude/helpers/graft-*.cjs`, and Codex's
   `~/.codex/hooks/graft/`) locate the installed package at runtime from four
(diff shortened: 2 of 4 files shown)


shhdwi:
Verified locally on top of current `main` (clean merge; build clean). Diff is test/CI/docs only (4 files, +59/-0): an additive `node24-wasm` CI job on Node 24 with pinned action SHAs, plus `test/generic-node24*` which warms and exercises every breadth grammar 64x in a child process with no V8 Zone OOM. No runtime code changed; existing matrix untouched. New test runs (not skips) on Node >=24 and passes. Merging.

Issue #117, opened 2026-08-15
Retrieval measurement: `ask` misses files its own graph holds one edge away (186 PRs, 4 languages)


gersonsebastianx:
# Retrieval measurement: `ask` misses files its own graph holds one edge away (186 PRs, 4 languages)

Hi — I wanted to evaluate graft properly before adopting it, so I built a
retrieval harness and ran it against 186 merged PRs across four repos. Sharing
the results because one of them looks like a concrete, cheap improvement rather
than a complaint. Everything below is reproducible and cost $0 to produce:
no judge model, no API key.

For transparency: I built and ran this with the assistance of a coding agent.
The method, the fairness corrections and the discarded results in §3 were all
worked through and verified rather than taken on trust — the harness is public
so you can check any of it yourself.

Harness: https://github.com/gersonsebastianx/graft-retrieval-bench

## Method, briefly

Ground truth comes from git: for each merged PR the issue/PR text is the query
and the source files the maintainers changed are the correct answer. The graph
is built at `merge_commit_sha^1` — the base commit — so the index can never
contain the fix. Queries are stripped of paths, filenames and code blocks, with
a hard audit that rejects any case where a literal filename survives.

Repos: pocketbase (Go), django (Python), nestjs/nest (TS), spring-boot (Java).
186 cases, k=10. Measured against **`@nanonets/graft` 0.10.1** (latest on npm at
the time of writing), default `graft build` — no `--deep`, no LSP enrichment.

## 1. `ask` trails a no-graph BM25 baseline

| retriever | R@1 | R@5 | R@10 | MRR | hit% | tok/query |
|---|---|---|---|---|---|---|
| `graft ask` | 13.2% | 37.0% | 48.5% | 0.293 | 57.5% | 811 |
| BM25 over raw files (no graph) | 26.9% | 56.1% | 66.5% | 0.449 | 75.3% | 147 |
| `git grep` on keywords | 10.4% | 37.1% | 47.0% | 0.244 | 53.8% | 142 |

Paired over the 186 cases: **−18.0 pts on R@10**, 95% CI [−24.4, −11.5],
t = −5.46. It trails on all four repos; the widest gap is TypeScript
(27.7% vs 65.7%).

Two fairness notes, since both moved the numbers in graft's favour and I'd
rather surface them than be asked. First: `ask` ranks symbols, and several
symbols routinely share one file, so asking for `-n k` returns fewer than k
[...]

PR #137, opened 2026-08-18
Prioritize distinct files before sibling spans

Skyline-23:
## Context

`ask` returns precise symbol spans, but several high-scoring sibling spans from one file can consume the output prefix before other relevant files appear. This PR implements the A1 control from the file-aware ranking experiments: keep the existing relevance ranking unchanged and change only the final listwise selection order.

This is intentionally smaller than file-level score pooling. It does not alter lexical scoring, BM25, PPR, scope admission, comparable-scope ranking, or cross-repository fusion.

## A1: selection-only file-first projection

Let the existing fully ranked hit list be:

```math
L = [h_1, h_2, \ldots, h_n]
```

Let `g(h)` be the exact file group of hit `h`. For a file `F`, its conceptual baseline score and first occurrence are:

```math
S_F = \max_{h:\,g(h)=F} s(h)
```

```math
first(F) = \min \{j \mid g(h_j)=F\}
```

Because `L` is already sorted by the existing score and tie-break rules, the first occurrence of `F` is exactly its baseline leader. A1 therefore orders files only by that existing first occurrence:

```math
F \prec_{A1} G
\iff
first(F) < first(G)
```

The implementation does not recompute scores or introduce a new comparator.

For each file, A1 keeps a queue containing its hits in their original baseline order:

```math
Q_F = [h_{F,1}, h_{F,2}, \ldots]
```

The final projection is depth-wise round-robin:

```math
\Pi_{A1}(L) =
[Q_{F_1}[0],\ldots,Q_{F_m}[0],Q_{F_1}[1],\ldots]
```

For example:

```text
baseline:   A1(.95), A2(.93), B1(.90), C1(.88), A3(.85), D1(.80)
file order: A, B, C, D
A1 output:  A1, B1, C1, D1, A2, A3
```

This gives A1 its core invariant over the same candidate universe:

```math
\mathrm{FileOrder}_{A0}
=
\mathrm{FileOrder}_{A1}
```

```math
\mathrm{R@k}_{A0}=\mathrm{R@k}_{A1},
\qquad
\mathrm{MRR}_{A0}=\mathrm{MRR}_{A1}
```

Only the placement of additional sibling spans changes. Distinct files are emitted first, then remaining spans are still available when the caller requests more capacity. There is no hard per-file quota.

## Why this is safe

- The original `AskHit` objects, scores, pointers, and symbol spans are preserved.
- The baseline top hit remains the first hit, so `coverage` and `coverageStrong` retain their existing meaning.
- Exact file paths are used as group keys; same basenames do not collide.
- Concepts and non-file structural results are singleton groups.
- Workspace child selection is deferred until cross-repository fusion completes, then applied once using `child + exact path` as the group key.
- No new candidate is created and no score is multiplied, normalized, or replaced.
- The returned array order is the authoritative selection order; `AskHit.score` remains the original relevance score.

## 186-case A/B benchmark

The A1 implementation was compared against the #126-compatible baseline across 186 fixed-base cases: Django 50, Nest 37, PocketBase 50, and Spring Boot 49. Both natural and stem-blind queries were evaluated with 10,000 repository-stratified paired bootstrap samples.

| Metric | Natural: baseline → A1 | Stem-blind: baseline → A1 |
|---|---:|---:|
| R@1 / R@5 / R@10 / MRR | unchanged in 186/186 cases | unchanged in 186/186 cases |
| Recall in first 8 emitted entries | 39.5% → 50.3% (**+10.7pp**, 95% CI +6.8…+15.0) | 22.6% → 30.9% (**+8.3pp**, 95% CI +4.8…+12.1) |
| Recall within 400-token budget | 46.6% → 51.5% (**+4.9pp**, 95% CI +2.4…+7.8) | 29.7% → 34.1% (**+4.4pp**, 95% CI +1.9…+7.2) |
| Pack tokens to 10 unique files | 667 → 364 (**-45.4%**, 95% CI -48.1…-42.6%) | 600 → 361 (**-39.8%**, 95% CI -42.3…-37.2%) |

Regression guards:

- Unique-file top arrays were exact in 186/186 cases.
- Common-prefix and common-order violations: 0.
- Per-repository R@10 or MRR regressions: 0.
- Missing metrics or benchmark errors: 0.

Broader file-aware scoring experiments remain performance-test work and are not included in this PR.

## Changes

- Add a pure file-first round-robin selector.
- Apply it after the existing direct `ask` ranking.
- Defer workspace child selection and apply it once after final workspace fusion.
- Add unit, direct-ask, multi-scope, and workspace regression coverage.

## Validation

- `npm run build`
- Isolated full test suite: **718/718 passed**
- Focused ask/file-selection/scope-comparability/workspace tests: **81/81 passed**
- `git diff --check upstream/main...HEAD`
- Final diff: 6 files, +263/-11

## Limitations

- This does not change one-hop expansion, relation weights, IDF/BM25, PPR, scope admission, or file relevance scores.
- Workspace selection can only diversify the candidates present within the existing child overfetch window.
- The benchmark makes no latency claim for the optimized Spring Boot projection run.

Relates to #117


shhdwi:
Verified locally on top of current `main` (clean merge — built on #126's multi-scope ranking; `npm run build` clean).

Repro (issue's fixture: 3 `validate*` symbols in `a.ts`, one match each in `b.ts`/`c.ts`; `ask "validate token" -n 3`):
- **Before:** all 3 slots are `a.ts` siblings (`validateToken`, `validateTokenExpiry`, `validateTokenScope`) — distinct files crowded out.
- **After:** `validateToken` (a.ts, top hit unchanged), `revokeToken` (c.ts), `refreshToken` (b.ts) — distinct-file leaders surface within the same N.

No regression: top hit preserved; siblings return at N=5 (not dropped); single-symbol query still exact-matches at rank 1; concepts/structural neighbours use singleton groups so "who calls"/"dependencies" answers can't be reordered. `ask*`/`workspace` tests 93/93. Fixes #186. Merging.

Issue #139, opened 2026-08-18
PHP grammar in tree-sitter-wasm 1.1.4 crashes on heredocs, and generic.ts hides the failure


Mult1Hunter:
**Env:** main at 6f871a9, Node 24.16.0, web-tree-sitter 0.25.10. The npm release (0.10.1) is not affected because it predates #123.

Since #123 switched the grammar bundle to `tree-sitter-wasm@1.1.4`, the PHP grammar throws `RuntimeError: memory access out of bounds` on any file that contains a heredoc or nowdoc. This is the smallest file that triggers it:

```php
<?php
namespace App;
class WithHeredoc {
    public function sql(): string {
        return <<<SQL
            SELECT 1
            SQL;
    }
    public function other(): int { return 2; }
}
```

On 0.10.1 (`tree-sitter-wasms@0.1.13`) this yields a module, a class and two functions. On main it yields only the file node. A nowdoc (`<<<'TXT'`) behaves the same way.

The bug is in the grammar build, not in graft. Loading the 1.1.4 `tree-sitter-php.wasm` directly into web-tree-sitter and calling `parser.parse(src)` throws the same error, both with a whole-string input and with graft's chunked callback. The 0.1.13 wasm parses the file without complaint. The `php_only` grammar in the same bundle crashes too, and 1.1.4 is the newest tree-sitter-wasm on npm, so there is currently nothing to upgrade to. I have not bisected tree-sitter-php itself.

The impact is much wider than the files that contain a heredoc. On a Laravel repo with 583 PHP files, only 16 have one, but once the wasm heap is in a bad state other files start throwing as well. In total 373 of the 583 files throw, the number of classes found drops from 454 to 186, and the build still prints `parsed 583 of 583`. That last part is because `extractGeneric` catches the exception and returns just the file node with no message, so `build.ts` never learns about it. The extract cache then stores the empty result as a successful parse and replays it on every incremental build.

There are two separate things to fix here:

- The grammar. Either pin PHP back to a wasm build that does not crash, or take this upstream to tree-sitter-wasm / tree-sitter-php.
- The swallowing. #140 rethrows the exception so the failure is recorded in `errors` and in the cache, the same way every other parse failure already is. That part does not depend on what happens with the grammar.

Crysthamus:
[...]

PR #140, opened 2026-08-18
fix(generic): surface a throwing grammar as a build error instead of swallowing it

Mult1Hunter:
Refs #139.

## Problem
`extractGeneric` catches any exception thrown by `parser.parse` and returns `{ nodes, rawEdges }` containing only the file node. When a grammar actually throws, the file silently loses every symbol, `build` still reports it as parsed, and the extract cache stores the empty result as a successful parse and replays it on every incremental build until the file changes.

The concrete trigger was the tree-sitter-wasm 1.1.4 PHP grammar crashing on heredocs; PHP has since moved to the depth tier (#157), so this now stands as general hardening for any breadth grammar.

## Fix
Rethrow the exception with the language name in the message. Everything downstream already exists: `build.ts` records per-file failures in `errors`, the CLI prints them, and the cache stores an `error` entry for them. A file whose grammar throws now prints an error line, counts as a build error, and no longer gets a file node - the same behaviour depth-tier parse failures already have. `check.ts` wraps its own call to `extractGeneric`, so it is unaffected. A graph built before this change and checked after it will read as STALE for such files until it is rebuilt.

## Test
Reworked per review: the tests no longer depend on the real PHP crash (gone from the breadth tier since #157, and heap-state dependent anyway). A small test seam - `swapGrammarForTest` in `generic.ts`, returning the previous entry so tests restore it, in the same style as the existing seams in `load.ts` and `context/build.ts` - swaps a deterministically throwing fake grammar into the loaded-grammar map. Two tests, both red on main with only the seam applied:

- `extractGeneric` rethrows with the language named and the original message preserved
- through `buildGraph`, seeded under the real `rust` name so build's own warm call is a no-op: the file becomes a per-file entry in `errors`, gets no file node, and an incremental rebuild of the unchanged file replays the cached failure instead of a cached empty success

Suite: 917/917 on Node 24; 913 pass, 4 skipped, 0 fail on Node 20.

### DIFF
--- a/src/graph/generic.ts
+++ b/src/graph/generic.ts
@@ -85,7 +85,7 @@ const KIND: Record<string, Kind> = {
 
 // Loaded grammars + compiled tags queries, keyed by graft lang name. Populated by
 // warmGenericGrammars; read synchronously by extractGeneric.
-interface Loaded { language: unknown; query: unknown | null }
+export interface Loaded { language: unknown; query: unknown | null }
 const loaded = new Map<string, Loaded>();
 let tsMod: typeof import("web-tree-sitter") | null = null;
 let initPromise: Promise<void> | null = null;
@@ -153,6 +153,18 @@ export function isWarm(langName: string): boolean {
   return loaded.has(langName);
 }
 
+/** Test seam: swap in (or, with null, remove) a grammar entry, returning the
+ * previous one so the caller can restore it. The parse-failure path in
+ * extractGeneric needs a grammar that throws deterministically; the real crash
+ * that motivated it (tree-sitter-wasm PHP on heredocs, #139) is heap-state
+ * dependent, so tests inject a fake here instead of depending on it. */
+export function swapGrammarForTest(name: string, entry: Loaded | null): Loaded | null {
+  const prev = loaded.get(name) ?? null;
+  if (entry) loaded.set(name, entry);
+  else loaded.delete(name);
+  return prev;
+}
+
 /** Load one grammar from the tree-sitter-wasms bundle, initialising
  * web-tree-sitter on first call. Null when the wasm is missing or won't
  * instantiate — never throws, so a caller degrades instead of failing the build.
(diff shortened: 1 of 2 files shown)


shhdwi:
Reviewed locally on top of current `main` — **the fix is correct, but the test doesn't pass on main yet, so holding.**

The one-line change (rethrow instead of `catch { return … }`) is a sound general safety net: verified that base main silently returns a file-only result when a breadth grammar throws, while this PR throws `<lang> grammar threw: …`, which `build.ts` catches into `errors` and caches as a failure (replayed on incremental builds). Contract fully wired.

Two blockers before it can land:
1. **The new test fails on current main (15 pass / 1 fail).** #157 removed `.php` from `GENERIC_LANGS`, so `warmGenericGrammars(["php"])` / `isWarm("php")` no longer succeed and the test dies on its first assertion. The test relies on the real PHP heredoc crash through a breadth language that no longer exists.
2. The heredoc crash is heap-state-dependent, so even with `.php` re-added the build-level assertions are flaky (the probe throws but a subsequent `buildGraph` parsed cleanly).

Suggested: rework the test to inject a **deterministically-throwing** fake grammar rather than depend on the removed `.php` breadth path. The swallow-fix is worth keeping as general hardening; it just needs a test that's green on main.

(Note: #139 part (a) — the underlying tree-sitter-wasm PHP grammar heredoc bug — is now moot for PHP specifically, since PHP is depth-tier and no longer breadth-parsed; #139 stays open for that grammar-level issue.)

Mult1Hunter:
Reworked as suggested and rebased onto current main - the test no longer touches the removed PHP breadth path.

- New test seam: `swapGrammarForTest` in `generic.ts` (returns the previous entry so tests restore it; same style as the existing seams in `load.ts` and `context/build.ts`). The tests swap in a fake grammar whose language object throws on any property access, so the failure is deterministic instead of heap-state dependent.
- Coverage: (1) `extractGeneric` rethrows with the language named and the original message preserved; (2) through `buildGraph`, seeded under the real `rust` name so build's own `warmGenericGrammars` call is a no-op: per-file entry in `errors`, no file node, and a `reuse` rebuild of the unchanged file replays the cached failure instead of a cached empty success. Both are red on main with only the seam applied, green with the rethrow.
- PR description and commit message updated to note the fix is now general breadth-tier hardening, with the PHP heredoc crash as the historical trigger (#139 stays open for the grammar-level bug).

Suite: 917/917 on Node 24; 913 pass, 4 skipped, 0 fail on Node 20.

shhdwi:
Re-verified on top of current `main` — the earlier blocker is resolved. 🎉 The test no longer depends on the removed `.php` breadth path; it now injects a deterministically-throwing fake grammar via a `swapGrammarForTest` seam (a Proxy on `rust`), so it's green on main (`generic-extract` 21/21). `generic.ts` rethrows on a grammar crash (`<lang> grammar threw: …`) instead of swallowing to an empty file, and two #139 tests confirm the throw becomes a per-file build error cached as a failure (replayed on incremental rebuild) rather than a silently-empty "clean" result.

Together with #218 (the tree-sitter-wasm 1.1.6 grammar-crash fix already merged), this closes both halves of #139. Thanks for reworking the test! Merging.

PR #147, opened 2026-08-18
fix(claude): hooks, sync-run, and statusline respect GRAFT_DIR

### LINKED ISSUE #146: Hook integration (dist/claude/hooks.js) ignores --dir/GRAFT_DIR, always builds/queries at <CLAUDE_PROJECT_DIR>/graft
**Version:** `@nanonets/graft@0.10.1` (global npm install)

**Description**

The CLI's query/build commands correctly honor an explicit `--dir <path>` (`context graph directory`), letting the graph live somewhere other than `<repo>/graft`. But the Claude Code hook integration (`UserPromptSubmit`, `PostToolUse`, `SessionStart`, `Stop`, wired via `graft init`) never passes that flag through, so it's impossible to point the *automatic* hook-driven graph anywhere but the default location.

Specifically, in `dist/claude/hooks.js`:

```js
// line 30-32
function projectDir(input) {
    return process.env.CLAUDE_PROJECT_DIR || input.cwd || process.cwd();
}
```
```js
// line 268
const askArgs = ['ask', prompt, '.', '--json', '-n', '3'];
```
```js
// line 86-91 (graftJson)
function graftJson(dir, args, timeout = CHILD_TIMEOUT_MS) {
    const cliPath = process.env.GRAFT_TEST_CLI ?? graftCliPath();
    const out = execFileSync(process.execPath, [cliPath, ...args], { cwd: dir, ... });
    ...
}
```

`dir` is always `CLAUDE_PROJECT_DIR` (or cwd), and the CLI child is spawned with `cwd: dir` plus a hardcoded positional `.` for the repo-root argument, never a `--dir`/context-dir override, and `process.env.GRAFT_DIR` is never read here at all (it's only consulted in `ai/providers.js`'s `resolveConfig()`, for the `--deep` LLM build path — a completely separate config surface).

Additionally, `CLAUDE_PROJECT_DIR` itself can't be worked around from the environment: Claude Code re-injects it to the real absolute project root for every hook subprocess, overriding whatever a project's own `.env`/devcontainer config exports for that variable.

**Checked the README first — no documented workaround exists**

- Line 380 documents `graft --dir <path>` as a global flag ("use a context dir other than `<repo>/graft`"), but this is never threaded into the hooks' own `graft ask`/`graft check` invocations (confirmed above).
- Line 108 confirms `graft/` is intentionally a gitignored, per-teammate-regenerated local cache ("like `node_modules`") — so this isn't about wanting the graph committed elsewhere, just about the hooks respecting a locally-configured non-default location.
- The Claude Code hooks section (describing session-start/user-prompt/post-tool-use/stop) documents *that* the hooks keep the graph fresh automatically, but nowhere documents *where* that graph can be told to live other than the default.
- No `GRAFT_DIR` (or any other env var/config file) for this purpose appears anywhere in the README.

**Net effect:** if a project wants its canonical `graft/` graph to live somewhere other than `<repo>/graft` (e.g. a sandboxed config directory outside the checked-out repo, a monorepo subfolder, or any non-default path), every manual `graft ask/grep/skeleton/callers/map --dir <path>` call respects it, but the automatic hooks silently keep building and querying a second, separate graph at the default `<repo>/graft` — with no supported flag, env var, or settings.json field to redirect them.

**Reproduction**
1. `graft build --dir some/other/path` (or build normally, then move `graft/` elsewhere).
2. Wire the Claude Code hooks via `graft init`.
3. Send any prompt in Claude Code — `UserPromptSubmit` fires `graft ask "<prompt>" . --json -n 3` with `cwd` = the real project root, rebuilding/querying `<project-root>/graft` regardless of `--dir` used in step 1.

**Expected:** hooks respect a configurable context-graph directory (e.g. read `GRAFT_DIR` the same way `resolveConfig()` does, or read a persisted `--dir` from `.graft/config.json` the way `--include-dir` is already persisted via `writeBuildConfig`).

**Suggested fix:** thread `process.env.GRAFT_DIR` (or a persisted build-config dir) into `graftJson`'s `askArgs`/`checkArgs` as a `--dir` flag, mirroring how `resolveConfig()` already treats `GRAFT_DIR` for the LLM path.

bhavesh-gupta-investis:
## Summary

Fixes #146.

The Claude Code hooks (`UserPromptSubmit`, `PostToolUse`, `SessionStart`, `Stop`), `sync-run`'s background rebuild, and the statusline all resolved a bare project dir and hardcoded `<projectDir>/graft` internally — ignoring `GRAFT_DIR` entirely, unlike a direct CLI call, which threads `--dir` through `contextDirFor`. That made it impossible to point the automatic, hook-driven graph anywhere but the default location, even though `GRAFT_DIR` already exists and is honored by `resolveConfig()` for the `--deep` LLM path.

- Added `resolveContextDir(projectDir)` in `util/state.ts`, mirroring `contextDirFor`'s override precedence: `GRAFT_DIR` wins (relative values resolve against `projectDir`), else the default `<projectDir>/graft`.
- `cacheDir` and `claude/stats.ts`'s `readWiring` now resolve through it — this alone fixes the stats cache, sync lock, per-session state, and the upkeep stamp for every existing caller (hooks, sync-run, statusline, upkeep) with **no signature changes** anywhere.
- The hooks' own `graft ask`/`graft check`/`graft build` children needed one more change each: append `--dir <resolved>` only when `GRAFT_DIR` is set — a no-op otherwise, so an unconfigured repo's spawned CLI sees byte-identical argv to before this existed.
- `sync-run.ts`'s `realBuild` now also honors the existing `GRAFT_TEST_CLI` test seam (previously only used in `hooks.ts`), so its argv is directly assertable in tests.

## Why

Filed as #146 after hitting this in a real devcontainer setup where the checked-out repo and the actual Claude Code config/home live in different places, and the only supported knob (`GRAFT_DIR`) turned out to be silently ignored by every hook-driven code path.

## Test plan

- [x] `npx tsc -p tsconfig.json --noEmit` — clean
- [x] `npm test` — full suite passes unmodified on `main` (713/713, with `GRAFT_DIR` unset)
- [x] `npm test` — passes at 722/722 with this branch (9 new tests, `GRAFT_DIR` unset)
- [x] Added tests: `resolveContextDir`/`cacheDir` default + relative + absolute `GRAFT_DIR` (`test/util-state.test.ts`); `readWiring` honoring a relocated dir (`test/claude-stats.test.ts`); the prompt/post-edit hooks and `sync-run`'s real build path each carrying `--dir <resolved>` when `GRAFT_DIR` is set and omitting it when unset, plus `SessionStart` reading `INDEX.md` from the relocated dir (`test/claude-hooks.test.ts`)

Note: this container has an ambient `GRAFT_DIR` env var set (from an unrelated devcontainer config), which pollutes ~42 unrelated tests if left set — all commands above were run with it explicitly unset to get a clean baseline; this is purely local environment noise, not something the PR needs to handle.

### DIFF
--- a/src/claude/hooks.ts
+++ b/src/claude/hooks.ts
@@ -4,7 +4,7 @@ import { join, basename, isAbsolute } from 'node:path';
 import { readWiring } from './stats.js';
 import { formatBlastRadius, relevantRetrieval, formatOrientation } from './format.js';
 import { indexFreshness, staleBanner } from '../context/check.js';
-import { patchStats, readStats, acquireLock, readSession, writeSession } from './state.js';
+import { patchStats, readStats, acquireLock, readSession, writeSession, resolveContextDir } from './state.js';
 import { graftCliPath, claudeScriptPath } from './paths.js';
 import { runUpkeep } from '../upkeep-run.js';
 import { runningVersion } from '../upkeep.js';
@@ -78,6 +78,18 @@ function installedHookTimeout(dir: string, event: string): number | null {
   }
 }
 
+/**
+ * Append `--dir <contextDir>` for the hooks' own `graft ask`/`graft check`
+ * children — the one place in this file that spawns the CLI itself rather
+ * than reading `graft/` off disk (which already resolves through
+ * `resolveContextDir` inside `util/state.ts` and `claude/stats.ts`). A no-op
+ * when `GRAFT_DIR` isn't set, so an unconfigured repo's spawned CLI sees
+ * byte-identical argv to before this existed.
+ */
+function withContextDirArg(dir: string, args: string[]): string[] {
+  return process.env.GRAFT_DIR ? [...args, '--dir', resolveContextDir(dir)] : args;
+}
+
 function graftJson(dir: string, args: string[], timeout: number = CHILD_TIMEOUT_MS): any | null {
   try {
     // GRAFT_TEST_CLI is a test seam (mirrors GRAFT_TEST_STDIN/GRAFT_TEST_SYNC_RUN) so
@@ -98,7 +110,7 @@ function graftJson(dir: string, args: string[], timeout: number = CHILD_TIMEOUT_
   }
 }
 function checkStaleCount(dir: string): number {
-  const r = graftJson(dir, ['check', '.', '--json']);
+  const r = graftJson(dir, withContextDirArg(dir, ['check', '.', '--json']));
   const g = r?.graph ?? {};
   return (g.changed?.length ?? 0) + (g.added?.length ?? 0) + (g.removed?.length ?? 0);
 }
@@ -224,7 +236,7 @@ export async function main(event: string): Promise<void> {
     // server fill that cache, this only reads it.
     const upkeep = runUpkeep(dir, runningVersion(), { background: false }).lines;
     try {
-      const idx = readFileSync(join(dir, 'graft', 'INDEX.md'), 'utf8');
+      const idx = readFileSync(join(resolveContextDir(dir), 'INDEX.md'), 'utf8');
       const banner = staleBanner(indexFreshness(dir)) ?? undefined;
       const orientation = formatOrientation(idx, undefined, banner);
       emit('SessionStart', upkeep.length ? `${upkeep.join('\n')}\n\n${orientation}` : orientation);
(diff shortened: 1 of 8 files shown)


shhdwi:
Verified locally on top of current `main` (clean merge; build clean; full suite 874/874).

Confirmed #146: hooks hardcoded `<projectDir>/graft` and ignored `GRAFT_DIR`/`--dir`. The fix adds `resolveContextDir(projectDir)` in `src/util/state.ts` (honors `GRAFT_DIR` — absolute verbatim, relative against projectDir), threaded through `cacheDir`, `readWiring`, the SessionStart INDEX.md read, `sync-run`'s build, and a `withContextDirArg` helper that appends `--dir <resolved>` to the hooks' own `graft ask`/`graft check` children — a byte-identical no-op when `GRAFT_DIR` is unset.

Passing tests prove it: prompt hook passes `--dir <resolved>` to `graft ask` under GRAFT_DIR; post-edit passes `--dir` to `graft check`; sync build passes `--dir`; session-start + readWiring read from a GRAFT_DIR-relocated dir; plus resolveContextDir unit tests (default/relative/absolute) and the unset byte-identical-argv case. Fixes #146. Merging.

PR #149, opened 2026-08-19
feat: add --only-dir whitelist to graft build

jxq0:
Adds a `--only-dir` option to `graft build` that restricts indexing to files under one or more repo-relative paths.

Motivation: in large monorepos you often only care about one subtree. `--include-dir` adds grammars but still walks and indexes everything; `--only-dir` narrows the walk itself.

Details:
- Repeatable, repo-relative path prefix, normalized the same way as `--in` (posix, no `./`, no trailing slash).
- The whitelist is NOT persisted to `.graft/config.json`. It is recorded in the graph fingerprint at build time, so a later build and the hooks/refresh freshness probe walk the same set without leaving a trace in the indexed repo.
- A prefix that normalizes to "" (a bare "/" or ".") is rejected.
- Empty or absent whitelist is a no-op.

Tests: `test/graph-only-dir.test.ts`.


shhdwi:
Verified locally on top of current `main`. Rebased the `cli.ts` conflict with #156 (kept both `--only-dir` and `--no-gitignore`/`--no-ignore` options) and dropped a stale tracked `node_modules` symlink this branch carried.

`--only-dir <path>` (repeatable, repo-relative, normalized like `--in`) restricts the build to a whitelist: a repo with `keep/`, `other/`, `top.ts` built with `--only-dir keep` indexed only `keep/k.ts`; the choice is recorded in the graph fingerprint (read back by `check`/probe/refresh) so a later no-flag build and the hooks path walk the same set, and excluded files don't show as phantom drift. Build clean; `graph-only-dir` + `context` 32/32. Merging.

PR #152, opened 2026-08-19
fix(php): keep enums with array consts in the graph (#145)

### LINKED ISSUE #145: PHP: enums with a const member vanish from the graph (methods leak to file scope)
## Summary

A PHP `enum` that declares a `const` member is **dropped from the graph
entirely**. No `enum` node is emitted, and any methods defined on the enum
**leak into file scope as bare `function` nodes** (losing their `Enum.method`
qualification). The same declarations on a `class` are handled correctly, so the
failure is enum-specific.

Enum constants are valid PHP (allowed since 8.1). The parse simply fails silently
and the resulting `ERROR` node is swallowed, so `graft build` reports success with
a quietly incomplete graph. This is the same "grammar chokes -> node silently
dropped" failure mode as #139 (heredocs), for a different construct.

## Reproduction

`e.php`:

```php
<?php

enum Foo: string
{
    case A = 'a';
    case B = 'b';

    private const MAP = [
        'a' => 'Alpha',
        'b' => 'Beta',
    ];

    public static function nameFor(string $k): ?string
    {
        return self::MAP[$k] ?? null;
    }
}
```

```
graft build <dir> --dir <out>
```

### Observed (non-`file` nodes in `wiring.json`)

```
function:nameFor
```

No `enum` node. `nameFor` is a top-level `function`, not `Foo.nameFor`.

### Expected

```
enum:Foo
method:Foo.nameFor
```

## Isolation

Built minimal dummies and ran the real `graft build` on each:

| Case | Declaration | `enum`/`class` node? | method placement |
|------|-------------|----------------------|------------------|
| enum + **array** const | `enum Foo { case A; private const MAP = [...]; f() }` | **missing** | leaks to file scope |
| enum + array const, no method | `enum Foo { case A; private const MAP = [...]; }` | **missing** | -- |
| enum + **scalar** const | `enum Foo { case A; private const LABEL = 'x'; f() }` | present | `Foo.f` (correct) |
| **class** + array const | `class Foo { private const MAP = [...]; f() }` | present (class) | `Foo.f` (correct) |

So the trigger is a `const` **inside an `enum` body**; an array value makes it
worse (see root cause).

## Root cause

The bundled PHP grammar cannot parse a `const` declaration inside an `enum` body.
tree-sitter-php `0.23.12` (both the `php` and `php_only` grammars), on the
reproduction above:

```
(program (php_tag)
  (ERROR (name) (enum_case name: (name) value: (string ...)) (visibility_modifier))
  (const_declaration (const_element (name) (array_creation_expression ...)))
  (ERROR))
```

The entire `enum_declaration` collapses into `ERROR`; the `const` floats up to
`program` level. Since there is no `enum_declaration` node, the extractor emits
no enum, and the method it walks next is attributed to the file.

With a **scalar** const the recovery is partial -- the `const` still parses to an
`ERROR` inside the body, but the surrounding `enum_declaration` node survives, so
the enum is still emitted (which is why the scalar case above "works"). A `class`
body accepts the identical `const` with no error.

For contrast, a Mago-based (Rust) PHP parser produces the correct
`enum:Foo` + `method:Foo.nameFor` for the exact same input, confirming the source
is valid PHP and the gap is in the tree-sitter-php grammar Graft bundles.

## Suggested direction

- Track/bump the tree-sitter-php grammar once it accepts `const` in enum bodies
  (this is fundamentally an upstream grammar gap), and/or
- Add `ERROR`-node resilience so a single mis-parsed member does not silently
  discard the whole type + reparent its methods (related to the silent-failure
  concern in #139).

## Environment

- `@nanonets/graft` 0.11.0
- tree-sitter-php 0.23.12 (grammar used for the AST dump above)

Frankie-Xu:
## Summary
- tree-sitter-php 0.23.12 collapses a legal `enum` + array `const` into ERROR and recovers the method as a file-scope `function_definition`.
- Reparent that recovery shape in the PHP depth walker: an ERROR that already contains `enum_case` + `name` becomes `enum:Foo`, and the recovered `function_definition` becomes `method:Foo.nameFor`.
- Unknown ERROR nodes are still walked and are never mapped to a type. Clean `class_declaration` / `enum_declaration` trees are unchanged.

## Test plan
- [x] `npx tsx --test test/graph-php.test.ts` (11/11)
- [x] `npm run build`
- [x] `npm test` (717/717)
- [x] Confirm a following top-level `function` after the enum stays `function:topLevel`

Closes #145

### DIFF
--- a/src/graph/extract.ts
+++ b/src/graph/extract.ts
@@ -322,7 +322,7 @@ export function extractFile(rel: string, source: string, lang: Language): Extrac
   // childCtx, so a by-ref Set there would read as ordinary inherited context
   // when it's actually accidental shared mutable state across the whole walk.
   const minted = new Set<string>([rel]);
-  for (const child of root.namedChildren) walk(child, ctx, nodes, rawEdges, minted);
+  walkNamedChildren(root.namedChildren, ctx, nodes, rawEdges, minted);
   // nodes[0] is the file node; the rest are its symbols. Index the module-level
   // residual on the file node so a term outside every symbol still surfaces it.
   nodes[0].body_text = fileResidual(source, nodes.slice(1));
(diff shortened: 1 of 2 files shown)


Frankie-Xu:
Local verification on `fix/145-php-enum-error-recovery` (rebased on current `main`):

- `node --import tsx --test test/graph-php.test.ts` — **11/11**
- `npm run build` — OK
- `npm test` — **717/717**

All test-plan items are checked in the PR description, including the following top-level function case (`function:topLevel` stays file-scope after enum recovery).

CI workflows on this fork PR show **action required** — happy to re-run once approved.

shhdwi:
Verified locally on top of current `main`; `npm run build` clean.

Reproduced #145's exact repro (enum with an array const + a static method) and proved causality by swapping only `src/graph/extract.ts` between main and this PR against the identical file:
- **Before (main):** emits `file:e.php` + `function:nameFor` — no enum node, the method leaked to file scope as a bare function (matches the issue's "Observed").
- **After (this PR):** emits `enum:Foo` + `method:Foo.nameFor` (kind `method`, re-nested under the enum) — matches "Expected".

`graph-php.test.ts` → 11/11 pass, including 4 new #145 tests (array-const enum kept; scalar-const enum control; class-with-array-const control unchanged; recovery doesn't swallow a following top-level function). Fixes #145. Merging.

shhdwi:
Rebased onto current `main` to resolve the `extract.ts` conflict with #176 (Java anonymous-class handling) — the two are additive (PHP enum recovery vs Java `{anonymous}` minting). Verified after resolution: build clean, `graph-php` 11/11 and `graph-java` 24/24 both pass, so neither fix was lost. Merging.

PR #153, opened 2026-08-19
fix(graph): index Dart top-level functions, consts, and call edges

### LINKED ISSUE #134: Dart: top-level functions/consts absent from the graph; callers reports 'no symbol' and skeleton lists function-body locals
## Summary

On Dart sources, the parser appears to skip top-level declarations and descend into function bodies instead. A top-level function with real callers is absent from the graph entirely, so `graft callers` reports `no symbol`, and `graft skeleton` lists locals from inside function bodies as if they were the file's API. Classes and methods are indexed correctly.

`graft check` reports the graph in sync, so nothing signals the omission.

**Version:** graft 0.10.1 (`@nanonets/graft`), macOS (darwin arm64), node 25.6.0.

## Repro

A 14-line Dart file, no dependencies:

```dart
// lib/example.dart
const int kThreshold = 3;

bool isReady(int count) => count >= kThreshold;

String describe(int count) {
  final label = isReady(count) ? 'ready' : 'waiting';
  return label;
}

class Counter {
  int value = 0;

  bool ready() => isReady(value);
}
```

```
git init && git add -A && git commit -m repro
graft build          # parsed: 1 of 1 files
```

### `graft callers isReady` — has two callers (`describe`, `Counter.ready`)

```
no symbol "isReady" in the graph — check spelling or run graft build
```

Same for `describe`, and for the qualified forms `example.isReady` / `isReady(int)`.

### `graft skeleton lib/example.dart`

```
- L6-L6    variable label    final label = isReady(count) ? 'ready' : 'waiting';
- L10-L14  class Counter     class Counter
- L10-L14  class int         class Counter
```

`isReady`, `describe`, `kThreshold` and `Counter.ready` are all missing. `label` is a local inside `describe`. The third row also looks like a mislabeled duplicate of the second (`class int` for `class Counter`).

### What does work

- `graft callers Counter` resolves the class.
- `graft grep "isReady"` finds every occurrence, correctly grouped by enclosing symbol — so the information is in the index; it is the symbol table / edge extraction that misses it.
- `graft check` reports: `graph check: OK — the wiring graph is in sync with the code.`

Reproduced in a real Flutter app (~766 indexed files) on `isObligation`, `isNamed`, `outstandingObligations`, `notificationAskRetired`, `promptNotificationPermission` — all top-level functions, all `no symbol`; classes in the same files resolve.

## Why this one bites harder than a normal coverage gap

The docs make `callers` load-bearing for exactly the case it silently cannot answer — "run before you rename, delete, or change its signature", and `--depth all` before a refactor. Dart puts a lot of its API at top level (providers, predicates, helpers, constants), so on a Flutter codebase this is a large fraction of the symbols someone would check before a rename.

The error text is the sharp edge: `no symbol "X" — check spelling or run graft build` reads as *the caller made a mistake*. The natural inference is "nothing references this, safe to rename", which is the opposite of the truth.

## Suggested fixes, roughly in order of value

1. Index Dart top-level functions, getters, setters and consts as symbols with edges.
2. Until then, make the miss legible rather than blaming the caller — if the name appears in the index but has no symbol entry, say so: "'X' is in the index but not indexed as a symbol (Dart top-level declarations are not yet supported) — try `graft grep`".
3. Have `graft check` report parser coverage gaps rather than an unqualified in-sync OK.
4. `skeleton` should not surface function-body locals as file-level API, independent of the above.

## Workaround

`graft grep "<name>"` for top-level symbols; treat `no symbol` as "ask grep", never as "nothing calls this".

Frankie-Xu:
## Summary
- Dart is a breadth-tier language, but it had no `tags.scm`, so the node-kind walker skipped top-level `function_signature` nodes, minted `class_body` as `class int`, and listed function-body locals (`label`) as file-level API.
- Add a Dart tags query for top-level functions, methods, classes, `const`/`final`, and `identifier(args)` calls; expand signature spans over the sibling `function_body` so `describe → isReady` and `Counter.ready → isReady` resolve.
- Regression fixture from issue #134. Other breadth-tier languages are unchanged.

Fixes #134.

## Test plan
- [x] `node --import tsx --test test/generic-extract.test.ts test/graph-languages.test.ts` (20/20)
- [x] `npm run build`
- [x] `npm test` (716/716)
- [x] `graft build` on issue #134 repro: `graft callers isReady` finds `describe` + `Counter.ready`; `graft skeleton lib/example.dart` lists `kThreshold`, `isReady`, `describe`, `Counter.ready` (no body-local `label`); `graft grep "isReady"` finds all 3 occurrences

### DIFF
--- a/src/graph/generic.ts
+++ b/src/graph/generic.ts
@@ -255,7 +255,7 @@ export function extractGeneric(rel: string, source: string, langName: string): E
   };
 
   if (entry.query) {
-    tagsExtract(entry.query, tree.rootNode as TsNode, rel, mkDef, defs, rawEdges);
+    tagsExtract(entry.query, tree.rootNode as TsNode, rel, mkDef, defs, rawEdges, langName);
   } else {
     walkExtract(tree.rootNode as TsNode, mkDef); // no tags.scm → symbols only
   }
@@ -366,6 +366,7 @@ function tagsExtract(
   mkDef: (name: string, kind: Kind, whole: TsNode) => void,
   defs: Def[],
   rawEdges: RawEdge[],
+  langName: string,
 ): void {
   const q = query as { matches(n: unknown): Array<{ captures: Array<{ name: string; node: TsNode }> }> };
   const matches = q.matches(root);
@@ -381,7 +382,7 @@ function tagsExtract(
     const defKey = Object.keys(cap).find((k) => k.startsWith("definition."));
     if (defKey && cap.name) {
       defNameAt.add(cap.name.startIndex);
-      mkDef(cap.name.text, KIND[defKey.slice("definition.".length)] ?? "function", defScope(cap[defKey]));
+      mkDef(cap.name.text, KIND[defKey.slice("definition.".length)] ?? "function", defScope(cap[defKey], langName));
     }
     if (("reference.call" in cap || "reference.send" in cap) && cap.name)
       calls.push({ name: cap.name.text, at: cap.name.startIndex });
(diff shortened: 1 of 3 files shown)


shhdwi:
Verified locally on top of current `main` (clean merge; `npm run build` clean).

Reproduced #134 on base main: `wiring: 4 nodes (2 class, 1 file, 1 variable), 0 edges`; skeleton showed a body-local `variable label` and a bogus `class int`; `callers isReady` → "no symbol" — matches the report.

With this PR (adds `dart.scm` + a Dart-gated `defScope` that expands `function_signature` to include its `function_body`): `wiring: 6 nodes (2 function, 1 constant, 1 class, 1 method, 1 file), 2 edges`; skeleton lists the real API (`kThreshold`, `isReady`, `describe`, `Counter`, `ready`); `callers isReady` resolves both `describe` and `Counter.ready`. Tests: `generic-extract` 15/15, broad `graph-*` 236/236. Fixes #134. Merging.

PR #154, opened 2026-08-19
fix(ingest): index a directory when the build root is a symlink (#143)

### LINKED ISSUE #143: Symlinked files are skipped
Usage of `lstatSync` results in `isFile()` returning false for symbolic links, instead of resolving the link and indexing the real file.

https://github.com/NanoNets/Graft/blob/main/src/ingest/fs.ts

Output of `graft build <dir>` where `<dir>` is a symlinked directory:

```
✓ wiring: 0 nodes (), 0 edges, 0 cards []
  parsed: 0 of 0 files (0 replayed from cache)
```

After resolving the link myself Graft is able to index the directory correctly.

Unsure if this behavior is intended, but my use case is indexing and building graphs for directories in the nix-store (e.g. -devel packages), which may sometimes be symlinked to their original source.

Frankie-Xu:
## Summary

- Unwrap a symlink-to-directory once when it is passed as the build root, so git and the filesystem walk run inside the real target.
- Keep emitted paths relative to the caller-facing root path the user passed in.
- Leave internal file and directory symlinks unfollowed, avoiding recursion loops and preserving existing walk semantics elsewhere.
- Broken symlink roots fail with a clear `broken symbolic link` error instead of a generic `scandir ENOENT`.
- Added regression tests in `test/ingest-fs.test.ts` and `test/graph-root.test.ts`.

## Test plan

- [x] `node --import tsx --test test/ingest-fs.test.ts test/graph-root.test.ts` (23/23)
- [x] `npm run build`
- [x] `npm test` full suite (720/720)
- [x] Manual smoke: real directory + symlink root → `graft build` parses files (`parsed: 1 of 1`, not `0 of 0`)
- [x] Broken symlink root reports a clear error instead of a generic scandir ENOENT
- [x] Internal file/dir symlinks are not followed (#143)

Closes #143

### DIFF
--- a/src/ingest/fs.ts
+++ b/src/ingest/fs.ts
@@ -2,8 +2,8 @@
  * Filesystem walking used by `init`/`check` to enumerate a repo's source files.
  */
 import { spawnSync } from "node:child_process";
-import { existsSync, lstatSync, readdirSync, statSync } from "node:fs";
-import { join, relative, resolve } from "node:path";
+import { existsSync, lstatSync, readdirSync, realpathSync, statSync } from "node:fs";
+import { isAbsolute, join, relative, resolve, sep } from "node:path";
 
 /** Directories that are dependency/build output, never source. */
 export const SKIP_DIRS = new Set([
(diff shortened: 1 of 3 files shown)


Frankie-Xu:
Still the #143 fix: unwrap a symlink-to-directory once so git and the walk run inside the real target, keep caller-facing paths, and leave internal symlinks unfollowed. Rebased onto current `main` (`ee1ef03`) after the 8/24 merge wave left this hanging.

`npm run build` OK. `npm test` **921/922** (#143 ingest tests 23/23; leftover is `viz --tabs` needing a local `graft/`, which #206 fixes on main).

Frankie-Xu:
Rebased onto `origin/main` `268e30d750b50df39d7f13873f696c4523fea6b1`.

Conflict was only `src/ingest/fs.ts` (`walkDir`): kept this PR's one-level symlink-root unwrap + path remap, and kept main's `WalkOptions` (`followNestedRepos` from #232) by passing `opts` through instead of the old boolean.

Targeted tests 23/23 (`test/ingest-fs.test.ts` + `test/graph-root.test.ts`); full suite 1066/1066.

Still Closes #143.

shhdwi:
Verified locally on top of current `main` (clean merge; build clean; 23/23). The **root-is-a-symlink** case is fixed — `walkDir` over a symlinked root now enumerates `src/real.ts` (0 files before), which resolves the nix-store reproduction in #143.

Scope note: this deliberately does NOT follow symlinks *inside* the tree (the PR documents that and ships a test locking it — `lstat().isFile()` stays false for inner symlinks). That's a reasonable, safe default (avoids double-indexing and symlink loops). Merging this as the scoped root-symlink fix.

I'm leaving #143 open for the remaining question — whether to also index inner symlinked source files — since that's the reporter's literal ask and it's now a conscious policy decision rather than an oversight. @shrishdwi can call it either way.

Issue #144, opened 2026-08-18
PHP: attribute usage and anonymous classes are not extracted


williamdes:
Graft **0.11.0** (PHP support from #64). Two minimal repros of PHP constructs the tree-sitter extractor doesn't surface in the wiring graph. Both are extremely common in real PHP frameworks (routing/DI/validation via attributes; inline handlers via anonymous classes), so the gaps hide a lot of real structure.

Verified by running `graft build <dir>` on the dummy files below and reading `.graph/wiring.json`.

---

## 1. PHP 8 attributes produce no node and no edge

```php
<?php
namespace Poc;

use Poc\Attr\Route;
use Poc\Attr\Deprecated;

#[Deprecated('use NewWidget')]
class Widget
{
    #[Route('/widgets', method: 'GET')]
    public function index(): void { $this->load(); }

    #[Route('/widgets/{id}', method: 'GET')]
    public function show(int $id): void {}

    private function load(): void {}
}
```

**Actual:** the wiring graph has nodes for `Widget` + its 3 methods, an `imports` edge per `use`, and the `index -> load` call. There is **no edge for the attribute usage** at all — nothing links `Widget.index` to `Route`, and nothing marks `Widget` as `Deprecated`. Which handler carries which `#[Route(...)]` — the heart of framework routing — is invisible in the graph.

**Why it's dropped:** a TS/Python *decorator* is a call expression, so it incidentally leaves a `calls` edge; a PHP *attribute* is metadata and never becomes a call, so it produces nothing.

**Suggested:** emit a `references` (or `uses`) edge from the annotated symbol to the attribute class, so `#[Route]`/`#[Deprecated]`/validation attributes are recoverable.

---

## 2. Anonymous classes are not captured (no node, no `implements`)

```php
<?php
namespace Poc;

function make(): Runnable
{
    return new class implements Runnable {
        public function run(): int { return 1; }
    };
}
```

**Actual:** the anonymous class gets **no node and no `implements` edge**. Only its `run` method appears, mis-attributed to the enclosing function (id `...#make.run`), so the type and its interface contract are lost.

**Suggested:** an anonymous-class node (e.g. `{anonymous}`, mirroring the existing `{closure}` naming) carrying its `implements` edge, with its methods nested under it.

---

Both repro on `graft build` with the snippets above as standalone `.php` files. Happy to send a PR for either — capturing attribute usage as `references` edges looks like the higher-value one.

shhdwi:
[...]

PR #155, opened 2026-08-19
feat(php): wire attribute usage as references edges (#144 part 1)

Frankie-Xu:
## Summary

- Extract PHP 8 `attribute_list` on class/method definitions and emit `references` edges from the annotated symbol to the attribute class (e.g. `Route`, `Deprecated`).
- Collect `use` import bindings so attribute names resolve via PSR-4 suffix matching (`resolvePhpUse`), same as other cross-file PHP wiring.
- Add issue #144 minimal repro test fixture.

Part 1 of #144 only — anonymous class extraction is intentionally left for a follow-up PR.

Refs NanoNets/Graft#144 (part 1 — attribute usage only).

## Test plan

- [x] `node --import tsx --test test/graph-php.test.ts` — 8/8 pass, including attribute references (#144)
- [x] `node --import tsx --test test/graph-references.test.ts` — TS references resolution unchanged
- [ ] `graft callers Route --direction in` on a Laravel/Symfony repo with route attributes shows incoming references from controller methods

### DIFF
--- a/src/graph/extract.ts
+++ b/src/graph/extract.ts
@@ -392,6 +392,7 @@ function walk(node: Parser.SyntaxNode, ctx: WalkCtx, out: NodeV1[], edges: RawEd
     // may `implements`, so every type declaration is a heritage site, not just a class.
     const javaTypeDecl = ctx.lang === "java" && JAVA_TYPE_KINDS.has(desc.kind);
     if (desc.kind === "class" || javaTypeDecl) edges.push(...heritageEdges(node, id, ctx));
+    if (ctx.lang === "php") edges.push(...phpAttributeReferenceEdges(node, id, ctx));
 
     const enclosingClass =
       desc.kind === "class" || javaTypeDecl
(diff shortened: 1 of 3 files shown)


shhdwi:
Verified locally on top of current `main` (clean merge, no conflicts; build clean; `graph-php` + `graph-resolve-typed` 27/27).

Confirmed the mechanism: for an attribute class **defined in the analyzed repo**, annotating a class/method with `#[Deprecated]` / `#[Route(...)]` now emits `references` edges (baseline 0 → this PR: `Widget → Deprecated`, `Widget.index → Route`).

Scope note for reviewers/#144: this resolves attribute usage **only when the attribute class lives in the repo**. The issue's motivating case — external/framework attributes (`use Poc\Attr\Route;` from vendor, never defined locally) — still yields no edge, because `resolve.ts` drops references whose target file isn't indexed (graft's emit-resolved-only stance). So this is a real part-1 improvement but not the full external-attribute story; I'm keeping #144 open for that.

Merging (part 1 of #144).

PR #156, opened 2026-08-19
feat(build): opt out of .gitignore and .ignore writes (#148)

### LINKED ISSUE #148: Add a way to opt out of graft writing .gitignore / .ignore (respect a global git ignore)
**Package:** `@nanonets/graft` 0.10.1
**OS:** macOS
**Setup:** A workspace of ~10 sibling git repositories, driven by Claude Code with graft's hooks installed (auto-`build` on Stop, session-start, and post-edit). A **global** git ignore (`core.excludesfile`) already ignores `graft/` for every repo.

## What happens
On every `graft build`, graft:

1. Appends a block to each repo's **local** `.gitignore`:
   ```
   # graft's local graph cache — regenerable, not committed (run `graft build`).
   graft/
   ```
2. Creates/updates a root **`.ignore`** file:
   ```
   # graft's cards are gitignored but should stay greppable: ripgrep reads
   # .ignore before .gitignore, so this re-admits the tree to search only.
   !graft/
   graft/.cache/
   graft/.graph/
   ```

Because builds run automatically (Stop / session-start / post-edit hooks) and across many repos, this shows up as constant churn: every repo I touch gets its `.gitignore` modified and a new `.ignore` created. When I delete the lines (since my global ignore already covers `graft/`), the next automatic build puts them straight back.

## Why this is a problem for my use case
- I already ignore `graft/` **globally**, so the per-repo `.gitignore` edits are redundant noise in `git status` and in diffs across ~10 repos.
- graft only ever reads/writes each repo's **local** `.gitignore`; it never checks `core.excludesfile`, so it can't tell the entry is already handled and keeps re-adding it.
- The writes are idempotent, but that doesn't help — anything I remove is restored on the next automatic build, so there's no way to "just delete it."

## Root cause (from the published `dist/`)
Two functions in `dist/context/node-file.js`, both called unconditionally from `dist/graph/build.js`:

- `ensureGitignored(root, contextDir)` — writes the `graft/` entry into `<root>/.gitignore`.
- `ensureSearchable(root, contextDir)` — writes `<root>/.ignore` to re-admit the gitignored cards to ripgrep.

Both early-return only when the output dir is **outside** the repo root (`rel` starts with `..`). I could not find any flag or env var that disables either write:

- `graft build --help` exposes nothing for it.
- `GRAFT_DIR` exists but is wired into the ask/AI provider path (`dist/ai/providers.js`), not as a general way to relocate the build output per repo, and relocating outside the repo also removes the in-repo greppable cards.

## Why the obvious workarounds don't work
- **"Just use the global ignore."** graft doesn't read it, so it re-adds the local line regardless.
- **"Delete the files."** They're restored on the next automatic build.
- **"Delete `.ignore` at least."** `.ignore` is load-bearing, not a git file: `graft/` is gitignored, ripgrep honors gitignore, so without `.ignore` the cards become invisible to grep and to the ripgrep-backed MCP search tools. A global git ignore does not substitute for it.

## Feature request
Please add a supported way to opt out of graft managing these files. Any one of these would solve it (roughly in order of preference):

1. **A config/env/flag to disable the writes** — e.g. `GRAFT_NO_GITIGNORE=1` (and a matching one for `.ignore`), or a `--no-gitignore` build flag. Users who manage ignores themselves (globally or otherwise) can turn it off.
2. **Respect the global excludes file** — if `graft/` is already ignored via `core.excludesfile` (or any active ignore), skip the local `.gitignore` write.
3. **Make it a one-time opt-in** rather than something re-applied on every automatic build — e.g. only `ensure*` on an explicit `graft build`, never on the hook-driven refresh path.

A note in the docs about what `.ignore` is for (so users don't delete it and silently break search) would also help.

## Repro
1. `npm i -g @nanonets/graft`
2. Add `graft/` to your global git ignore (`git config --global core.excludesfile`).
3. In any repo: `graft build` → local `.gitignore` gets the `graft/` block and a `.ignore` file appears.
4. Delete both additions.
5. `graft build` again → both are back.

Frankie-Xu:
## Summary

- Add `GRAFT_NO_GITIGNORE=1` to skip writing `/graft/` into the repo's `.gitignore` during `graft build` (including hook-triggered builds that inherit env).
- Add `GRAFT_NO_IGNORE=1` to skip writing `.ignore` (the ripgrep re-admit file). Default behavior is unchanged; both are opt-in disable switches using the same truthy parsing as `GRAFT_NO_REFRESH`.
- Add matching CLI flags: `graft build --no-gitignore` and `graft build --no-ignore`.
- Document the env vars in `.env.example`.

Closes #148

## Notes

- If you disable `.ignore` writes, ripgrep will honor `.gitignore` and **won't search `graft/` cards** unless you manage search exclusions yourself (e.g. keep a hand-written `.ignore`, or use `rg -uu`).
- Global `core.excludesfile` detection is intentionally out of scope here — a good follow-up if users want automatic skip without env vars.

## Test plan

- [x] `npm run build && npm test` — all green
- [x] `node --import tsx --test test/context.test.ts` — new unit + CLI integration tests pass
- [x] Existing `ensureGitignored` / `ensureSearchable` tests unchanged (default path)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,6 +39,10 @@
 # Where the graph lives (default: <repo>/.context)
 # GRAFT_DIR=.context
 
+# --- Build side effects (opt-out) ---------------------------------------------
+# GRAFT_NO_GITIGNORE=1  — skip writing graft/ into .gitignore (e.g. you already ignore it globally)
+# GRAFT_NO_IGNORE=1     — skip writing .ignore (ripgrep re-admit); only if you manage search yourself
+
 # --- Benchmark (bench/) -------------------------------------------------------
 # Per-role model overrides for the benchmark harness.
 # BENCH_AGENT_MODEL=anthropic/claude-sonnet-5
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -318,6 +318,8 @@ program
     (val: string, prev: string[]) => [...prev, val],
     [] as string[],
   )
+  .option("--no-gitignore", "skip writing graft/ into .gitignore (same as GRAFT_NO_GITIGNORE=1)")
+  .option("--no-ignore", "skip writing .ignore for ripgrep re-admit (same as GRAFT_NO_IGNORE=1)")
   .action(async (
     dir: string,
     opts: {
@@ -329,10 +331,14 @@ program
       allowPartial?: boolean;
       includeDir?: string[];
       followSubmodules?: boolean;
+      gitignore?: boolean;
+      ignore?: boolean;
     },
     command: Command,
   ) => {
     const buildStartedAt = Date.now();
+    if (opts.gitignore === false) process.env.GRAFT_NO_GITIGNORE = "1";
+    if (opts.ignore === false) process.env.GRAFT_NO_IGNORE = "1";
     const concurrency = opts.concurrency ? Math.max(1, Number(opts.concurrency)) : undefined;
     if (opts.concurrency && !Number.isFinite(concurrency)) {
       console.error(`✗ --concurrency must be a number, got "${opts.concurrency}"`);
@@ -472,7 +478,11 @@ program
     for (const e of g.errors) console.error(`✗ ${e}`);
 
     const rel = relative(process.cwd(), g.contextDir) || "graft";
-    console.log(`  ${rel}/ is git-ignored (added automatically) — a local cache; teammates run \`graft build\` to get their own.`);
+    if (process.env.GRAFT_NO_GITIGNORE) {
+      console.log(`  ${rel}/ is a local cache — add it to your gitignore if you want it untracked.`);
+    } else {
+      console.log(`  ${rel}/ is git-ignored (added automatically) — a local cache; teammates run \`graft build\` to get their own.`);
+    }
 
     // #127: a --deep run whose LLM calls failed used to print the same success
     // footer and exit 0, so a quota-exhausted build looked identical to a clean
--- a/src/context/node-file.ts
+++ b/src/context/node-file.ts
@@ -80,6 +80,12 @@ const MANIFEST_FILE = "manifest.json";
 /** Gitignored cache dir (per-file summaries + extractions), never committed. */
 export const CACHE_DIR = ".cache";
 
+/** Env kill switch — same truthy parsing as `GRAFT_NO_REFRESH` in ../graph/refresh.ts. */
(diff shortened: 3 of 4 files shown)


Frankie-Xu:
Rebased on current `main` (`73f81e5`) to resolve a conflict in `src/cli.ts`: the new `--allow-partial` deep-pass warning block (#127) landed next to the build footer this PR touches. Resolution keeps both — the footer message is now gated on `GRAFT_NO_GITIGNORE` as before, with the #127 block intact after it.

Local verification on the rebased head `e18eac0`:
- `npm run build` — OK
- `npm test` — **757/757**, including the three new #148 tests (`GRAFT_NO_GITIGNORE=1` skips `.gitignore`, `GRAFT_NO_IGNORE=1` skips `.ignore`, full `graft build` touches neither).

Frankie-Xu:
Rebased onto current `main` (`177591e`) to clear the conflict in `src/cli.ts`. Resolution keeps both: telemetry's `buildStartedAt` and the `--no-gitignore` / `--no-ignore` env-flag wiring at the top of the build action.

`npm run build && npm test` — 865/865.

Frankie-Xu:
Still the #148 opt-out: `GRAFT_NO_GITIGNORE` / `GRAFT_NO_IGNORE` (and `--no-gitignore` / `--no-ignore`) skip writing those files during `graft build`. Rebased onto current `main` (`ee1ef03`) after the 8/24 merge wave left this hanging.

`npm run build` OK. `npm test` **917/918** (#148 context tests 27/27; leftover is the same `viz --tabs` vs missing gitignored `graft/` that #206 fixes).

shhdwi:

Fixes #148: adds `--no-gitignore` / `--no-ignore` flags and `GRAFT_NO_GITIGNORE` / `GRAFT_NO_IGNORE` env vars (truthy-parsed, guarded at the top of `ensureGitignored`/`ensureSearchable`). Before/after in fresh git repos:
- Default build → writes the `graft/` block to `.gitignore` and writes `.ignore`.
- `--no-gitignore --no-ignore` (or the env vars) → neither file is created, the graph still builds, and the footer switches to "add it to your gitignore if you want it untracked."

`context.test.ts` 30/30 (new no-op + CLI-integration tests). Fixes #148. Merging.

PR #157, opened 2026-08-19
chore(graph): remove stale .php entry from GENERIC_LANGS (#142)

### LINKED ISSUE #142: PHP: remove stale .php entry from GENERIC_LANGS now that depth tier handles it
Follow-up to #64. `.php` is now a native depth/Tier-1 language (added to `EXTENSIONS`), and depth wins at routing (`build.ts`: `const generic = lang ? null : genericLangOf(f)`). But `.php` is still listed in `GENERIC_LANGS` in `src/graph/generic.ts`, which violates that file's own "must not collide with depth extensions" comment.

Harmless at runtime (depth supersedes), but it's stale/dead config. One-line removal.

Frankie-Xu:
## Summary
- Remove the `.php` row from `GENERIC_LANGS` in `src/graph/generic.ts`.
- PHP is already a depth/Tier-1 language (`EXTENSIONS` in `extract.ts`); `build.ts` routes depth before generic, so this entry was never used at runtime.
- Restores consistency with the registry comment: breadth extensions must not collide with depth extensions.

## Test plan
- [x] `npm run build`
- [x] `npm test`
- [x] `test/graph-php.test.ts` — PHP depth extraction unchanged
- [x] `test/supported-extensions.test.ts` — .php still indexed

Closes #142

### DIFF
--- a/src/graph/generic.ts
+++ b/src/graph/generic.ts
@@ -46,7 +46,6 @@ export const GENERIC_LANGS: readonly GenericLang[] = [
   { name: "c", exts: [".c", ".h"], wasm: "c" },
   { name: "cpp", exts: [".cpp", ".cc", ".cxx", ".hpp", ".hh"], wasm: "cpp" },
   { name: "ruby", exts: [".rb"], wasm: "ruby" },
-  { name: "php", exts: [".php"], wasm: "php" },
   { name: "c_sharp", exts: [".cs"], wasm: "c_sharp" },
   // These ship a tags.scm (calls + symbols); ocaml/zig have none and use the
   // node-kind walker fallback (symbols only) — still one row, zero query.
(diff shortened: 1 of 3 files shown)


shhdwi:
Verified locally on a worktree off current `main`:
- The diff removes the stale `.php` entry from `GENERIC_LANGS` (and its now-redundant breadth-tier tests) — PHP is native depth-tier since #64, and depth wins at routing, so the breadth entry was dead/colliding config exactly as #142 describes.
- `npm run build` clean; **22/22** tests pass (`graph-php`, `generic-extract`, `supported-extensions`).
- Behavioral check: `graft build` + `graft skeleton a.php` on a PHP file still extracts `class A` and methods `run`/`go` via the depth tier — no regression from removing the breadth path.

Merging.

PR #158, opened 2026-08-19
fix(php): resolve trait-inherited method calls via implements edges (#141)

### LINKED ISSUE #141: PHP: calls to trait-inherited methods aren't resolved through the trait-use edge
Follow-up to #64 (native Tier-1 PHP). Verified on 5 real repos.

**Symptom:** a class's `use SomeTrait;` correctly emits an `implements` edge to the trait, but calls to trait-provided methods (`$this->traitMethod()`) are **not** resolved: `$this` → enclosing class, the resolver looks for `Class.method`, doesn't find it on the class itself, and drops the call edge.

**Impact:** narrow — <1% of all call edges even on the most trait-composed PHP libraries measured (guzzle/psr7 0.7%, Carbon 0.9%). Real, but an edge case, not pervasive.

**Fix direction:** when a `$this->m()` call doesn't resolve on the class, follow the class's `implements`→trait edges and look for `m` on the used traits before dropping the edge.

Not a merge blocker (documented at merge time); tracked here.

Frankie-Xu:
## Summary

- Build a `classTraits` index from PHP `implements` raw edges (trait `use` is modelled as implements).
- After owner-qualified lookup on the class fails, walk used traits to resolve `$this->m()` to `Trait.m`.
- Derive method owner from dotted node id when extract omits it on trait/interface bodies (`Loggable.log` → owner `Loggable`).
- Drop rather than guess when multiple traits provide the same method name.

**Impact:** Edge-case PHP call resolution only (<1% edges per #141). Java/TS extends-chain behavior unchanged.

## Test plan

- [x] `node --import tsx --test test/graph-php.test.ts` — `$this->log()` → `Loggable.log`
- [x] `node --import tsx --test test/graph-resolve-typed.test.ts` — unit tests for trait resolve + ambiguity drop
- [x] `npm run build && npm test` — 712 tests pass

Closes #141

### DIFF
--- a/src/graph/resolve.ts
+++ b/src/graph/resolve.ts
@@ -224,6 +224,14 @@ function push<T>(map: Map<string, T[]>, key: string, val: T): void {
   else map.set(key, [val]);
 }
 
+/** Derive a method's owner from its dotted id when extract did not stamp `owner`
+ * (PHP trait/interface methods today). `app.php#Loggable.log` → `Loggable`. */
+function ownerFromMethodId(id: string): string | undefined {
+  const post = id.includes("#") ? id.split("#")[1] : id;
+  const segs = post.split(".");
+  return segs.length >= 2 ? segs[segs.length - 2] : undefined;
+}
+
 /**
  * Resolve a bare symbol name: same-file match first (certain → `extracted`),
  * else a unique cross-file match (→ `inferred`), else null (ambiguous/unknown).
@@ -284,6 +292,7 @@ function resolveTypedMember(
   file: string,
   ownerMethod: Map<string, NodeV1[]>,
   classParents: Map<string, string[]>,
+  classTraits: Map<string, string[]>,
   argCount?: number,
 ): { id: string; confidence: EdgeV1["confidence"] } | "ambiguous" | null {
   const MAX_DEPTH = 3;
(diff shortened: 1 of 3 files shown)


Frankie-Xu:
### Local verification

\`\`\`
npm run build && npm test
# 712 pass, 0 fail

node --import tsx --test test/graph-php.test.ts
# ✔ run → Loggable.log via trait-use (implements) edge

node --import tsx --test test/graph-resolve-typed.test.ts
# ✔ PHP trait resolve + ambiguous multi-trait drop
\`\`\`

Root cause had two parts:
1. `classParents` only walked `extends`, not trait-use (`implements`) edges
2. trait methods lacked `owner`, so `ownerMethod` index missed `Loggable.log` — fixed by deriving owner from id in resolve.ts

shhdwi:
Verified locally on top of current `main` (clean merge — touches `resolve.ts`, disjoint from the recent `extract.ts` work; `npm run build` clean).

Reproduced #141 with a `Loggable` trait defining `log()`, a `Service` doing `use Loggable;`, and `run()` calling `$this->log(...)`, built with both code versions on the same fixture:
- **Before (main):** the `$this->log()` call edge is silently **dropped** — the exact #141 symptom.
- **After (this PR):** it resolves — `wiring.json` shows `Service.run --calls--> Loggable.log`, and `graft callers log` shows `log ← calls ← run`.

`resolveTypedMember` now walks the class's `implements` (trait-use) edges after a same-class miss, and returns `ambiguous` when two traits supply the same method. Tests: `graph-php.test.ts` 7/7; `graph-resolve-typed.test.ts` 15/15 incl. the trait-resolution + two-trait-ambiguity cases. Fixes #141. Merging.

PR #159, opened 2026-08-19
feat: add Nix language support

Ansh-Sonkusare:
## Summary
- Add Nix (`.nix`) to the breadth-tier generic extractor using the existing `tree-sitter-wasm` WASM grammar from [nix-community/tree-sitter-nix](https://github.com/nix-community/tree-sitter-nix)
- Register nix in `GENERIC_LANGS` with `.nix` extension mapping
- Add `nix.scm` tags query capturing function definitions, inherit statements, and function calls (import, builtins.*, etc.)

## Test results
Verified against a 15-file NixOS flake: **50 nodes** (24 function, 15 file, 11 constant), **5 edges**. All 713 existing tests pass.

## Limitation
Tree-sitter queries cannot match children of `binding` nodes inside `let_expression`, so let-bound variables are not captured as individual definitions. They remain searchable via file node residual bodies.

### DIFF
--- a/src/graph/generic.ts
+++ b/src/graph/generic.ts
@@ -59,6 +59,7 @@ export const GENERIC_LANGS: readonly GenericLang[] = [
   { name: "zig", exts: [".zig"], wasm: "zig" },
   { name: "dart", exts: [".dart"], wasm: "dart" }, // surfaced by PR #38 (@muneebshere)
   { name: "clojure", exts: [".clj", ".cljs", ".cljc", ".bb"], wasm: "clojure" },
+  { name: "nix", exts: [".nix"], wasm: "nix" },
 ];
 
 const byExt = new Map<string, GenericLang>();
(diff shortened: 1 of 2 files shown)


shhdwi:
Verified locally on top of current `main`. Rebased past the additive `GENERIC_LANGS` conflict with the just-merged Lua PR #187 (kept both entries). Breadth-tier add (`nix` + `nix.scm`), no npm dep, no committed binary.

Behavioral: on an attrset with lambda bindings, a binding whose value is a lambda becomes a `function` def; non-function bindings aren't emitted; and an application of one binding by another resolves as a call (`greet → helper`). Unresolvable externals (`import`, `mkDerivation`) are correctly dropped, not guessed.

The one gap was **no test** (every other breadth language ships a `SNIPPETS` case). I added one, derived from the actual extraction — `generic-extract.test.ts` now 17/17 including `breadth tier: nix`. Merging.

PR #160, opened 2026-08-19
fix: Java generic type arguments are not supertypes (#104)

### LINKED ISSUE #104: Java heritage: generic type arguments become bogus extends/implements edges, and poison call resolution
Found while fixing #103. Same root cause — a generic's raw text is not the type's name — but this one is worse, because heritage edges feed call resolution.

Also from #83, so this is my defect to report.

## What happens

`typeIdentifiersIn` (extract.ts) collects **every** `type_identifier` under a heritage clause, including the ones inside `type_arguments`. So each type argument becomes a supertype.

```java
public final class Sorted implements Comparable<Item> { … }
```

emits two edges:

```
Sorted --implements--> Comparable
Sorted --implements--> Item        ← Sorted does not implement Item
```

## Real incidence on public repos

Built with current `main`, counting heritage edges whose target is a name that appears only as a type argument in the source:

| repo | heritage edges | pointing at a type argument | share |
|---|---:|---:|---:|
| [`google/gson`](https://github.com/google/gson) | 430 | **186** | 43% |
| [`spring-projects/spring-petclinic`](https://github.com/spring-projects/spring-petclinic) | 25 | **8** | 32% |

Samples, verbatim from the graph:

```
OwnerRepository      --implements--> Owner      (inferred)
OwnerRepository      --implements--> Integer    (inferred)
PetTypeRepository    --implements--> Integer    (inferred)
UtcDateTypeAdapter   --extends-->    Date       (inferred)
InterceptorAdapter   --extends-->    T          (inferred)
```

`OwnerRepository` implements `Repository<Owner, Integer>` — neither `Owner` nor `Integer`. The last one is a bare type **variable**, not a type at all.

Nearly half of gson's heritage graph is wrong, so `graft callers` on any interface with a generic parameter over-reports, and "what implements X" cannot be trusted for a language where `Comparable<T>` / `Function<A,B>` / `List<X>` supertypes are the norm.

## It does not stop at heritage

`extends` edges feed `classParents` (resolve.ts), which `resolveTypedMember` walks to resolve member calls up the inheritance chain. A wrong parent therefore produces wrong **call** edges. Verified on a minimal fixture:

```java
public class Item { public void poison() {} }
public class Base<T> { public void real() {} }
public final class Child extends Base<Item> {
  public void use() {
    Child c = new Child();
    c.poison();          // does not compile — Child has no poison()
  }
}
```

emits:

```
Child --extends--> Base   (inferred)
Child --extends--> Item   (inferred)     ← bogus
Child.use --calls--> Item.poison (inferred)   ← consequence
```

The call edge points at a method that is not reachable on the receiver in real Java. The bogus heritage edge is the only reason it resolves.

## Fix

Take the heritage clause's **named type** and erase its arguments, the same way #103 does for construction — rather than collecting every `type_identifier` beneath the clause. `implements A<X>, B` should yield `A` and `B`, never `X`.

Two decisions worth agreeing on before I write it:

1. **Type variables.** `class Foo<T> extends Bar<T>` currently emits `extends T`. A type parameter is not a supertype and should never produce an edge; the declaration's own `type_parameters` list is the authority on which names are variables.
2. **Nested/qualified supertypes.** `implements Outer.Inner` has the same ambiguity #103 hit for construction. Consistent with that decision, dropping is safer than collapsing to the last segment — but heritage already keeps the *unresolved name* as the edge target by design ("an unresolved base is usually an external/imported type — keep the name"), so the two paths differ and it is worth being explicit.

## Why this is not folded into #103

#103 is a regression fix with a two-file diff. This one changes what heritage means across a whole language and, through `classParents`, changes call resolution. Different blast radius, and it deserves its own before/after measurement rather than riding along.

Happy to implement it with the same treatment as #103 — tests verified to fail when reverted, and before/after counts on both repos above. Say which way you want (1) and (2) resolved and I will open it.

---

Two smaller things found in the same pass, mentioned so they are on record rather than rediscovered. Neither needs a decision here, and I can file them separately if useful:

- **`resolveName`'s same-file branch returns `local[0]` with no ambiguity check**, at `extracted` confidence, which contradicts the module header's own drop-rather-than-guess rule. #103 routes around it per-spelling; a `local.length === 1` guard fixes the class once for every relation. Measured as a no-op on both repos above with the suite green.
- **Methods declared in an anonymous class body take `owner` from the enclosing class**, so `ownerMethod` can resolve a call on the outer type to the anonymous implementation.

dbianco:
Closes #104. Both decisions taken as you specified, and the `resolveName` guard folded in.

## The defect

`typeIdentifiersIn` collected **every** `type_identifier` beneath a heritage clause, including those inside `type_arguments`:

```java
class Sorted implements Comparable<Item>
  -> implements Comparable
  -> implements Item          // Sorted implements no such thing
```

Heritage now reads the clause's supertype **entries** — `superclass` wraps one type, `super_interfaces`/`extends_interfaces` wrap a `type_list` — and names each by erasing its arguments.

## The two decisions

**Type variables — no edge.** Implemented as you described: erase arguments, then filter any surviving name that appears in the declaration's own `type_parameters`.

Worth reporting rather than implying: **erasure alone already handles every case in both repos.** Output is byte-identical with the filter removed (gson 265 heritage edges either way; the single one-letter target that remains is a real `class B extends A` in a test fixture). Java cannot extend or implement a type variable, so a surviving `T` would need a shape neither repo contains. I kept the filter because a wrong supertype is not cosmetic — it feeds `classParents` — but it is belt-and-braces, and no valid-Java fixture can make it fail. Happy to drop it if you would rather not carry unreachable code.

**Qualified supertypes — keep the whole name.** `Outer.Inner` stays `Outer.Inner`, per your lean. It is truthful, it honors heritage's existing keep-the-unresolved-name contract, and it cannot false-match a node id the way a bare `Inner` could. Noted in the code that this deliberately differs from #103, which *drops* a qualified construction — that path has no keep-the-name contract to fall back on, so the safe answers differ.

**`resolveName` same-file guard.** Folded in. The branch returned `local[0]` — first in document order — labelled `extracted`, while the cross-file branch two lines below required a unique match. It now requires a unique local match too.

## Measurement

`origin/main` vs this commit, baselines built from a clean worktree:

| | gson | spring-petclinic |
|---|---:|---:|
| heritage edges | 430 → **265** | 25 → **17** |
| calls edges | 5,358 → **5,344** | 258 → 258 |
| total edges | 12,588 → **12,409** | 987 → **979** |

Every removed heritage edge is a type argument. Spot-checked:

```
InterceptorAdapter extends TypeAdapter<T>            -> TypeAdapter          (was: + T)
UtcDateTypeAdapter extends TypeAdapter<Date>         -> TypeAdapter          (was: + Date)
OwnerRepository implements Repository<Owner,Integer> -> Repository           (was: + Owner, + Integer)
```

Legitimate supertypes are untouched — `JsonArray extends JsonElement`, `LazilyParsedNumber extends Number`, `Circle extends Shape` all survive.

**The 14 `calls` edges gson loses are the point, not a side effect.** `extends` feeds `classParents`, which `resolveTypedMember` walks, so a bogus supertype was resolving member calls against a type not in the receiver's hierarchy at all. The fixture from #104 reproduces it: `Child extends Base<Item>` made `c.poison()` resolve to `Item.poison`, a call that does not compile.

## Tests

Five cases: type argument erased, no call poisoning, type variable never a supertype, qualified name kept whole, and the resolver guard.

Each verified red when the behavior it covers is reverted — **except the type-variable case**, which erasure alone already satisfies, as above. Stating that rather than letting five green tests imply five pinned behaviors.

One methodology note, since it changed a number I would otherwise have reported wrong: my first pass measured the "after" graph while `origin/main`'s build was the one on disk, and I read 25 surviving bogus edges that did not exist. Rebuilt in the right order, those 25 are all legitimate supertypes my heuristic flags because their names also appear as type arguments elsewhere in the repo.

Suite: **718 passing**.

## Not included

The anonymous-class `owner` finding, as you suggested — I will open it separately.

### DIFF
--- a/src/graph/extract.ts
+++ b/src/graph/extract.ts
@@ -771,6 +771,7 @@ function heritageEdges(node: Parser.SyntaxNode, classId: string, ctx: WalkCtx): 
     // `superclass` holds `extends X`; `super_interfaces` holds `implements A, B`
     // (and, on an interface declaration, `extends A, B` — which tree-sitter-java
     // still spells `extends_interfaces`).
+    const typeParams = javaTypeParameterNames(node);
     for (const child of node.namedChildren) {
       const relation: Relation | null =
         child.type === "superclass"
@@ -779,8 +780,16 @@ function heritageEdges(node: Parser.SyntaxNode, classId: string, ctx: WalkCtx): 
             ? "implements"
             : null;
       if (!relation) continue;
-      for (const t of typeIdentifiersIn(child)) {
-        edges.push({ source: classId, relation, name: t, file: ctx.rel });
+      for (const entry of javaSupertypeEntries(child)) {
+        const name = javaSupertypeName(entry);
+        // Belt-and-braces. A type VARIABLE is never a supertype, and erasing the
+        // arguments already removes every case measured on gson and spring-petclinic
+        // (identical output with this filter removed) — Java cannot extend or implement
+        // a type variable, so a surviving `T` would have to come from a shape neither
+        // repo contains. Kept because a wrong supertype is not a cosmetic edge: it
+        // feeds `classParents` and from there call resolution.
+        if (!name || typeParams.has(name)) continue;
+        edges.push({ source: classId, relation, name, file: ctx.rel });
       }
     }
     return edges;
@@ -827,15 +836,54 @@ function heritageEdges(node: Parser.SyntaxNode, classId: string, ctx: WalkCtx): 
   return edges;
 }
 
-/** Every `type_identifier` under a heritage clause, so `implements A, B<C>` yields
- * each named type rather than the clause's raw text. */
-function typeIdentifiersIn(node: Parser.SyntaxNode): string[] {
-  const out: string[] = [];
+/**
+ * The supertypes a heritage clause names, one node each — NOT every `type_identifier`
+ * beneath it.
+ *
+ * `superclass` wraps a single type; `super_interfaces`/`extends_interfaces` wrap a
+ * `type_list` of them. Descending blindly instead walked into `type_arguments`, so
+ * `implements Comparable<Item>` reported `Item` as a supertype too.
+ */
+function javaSupertypeEntries(clause: Parser.SyntaxNode): Parser.SyntaxNode[] {
+  const list = clause.namedChildren.find((c) => c.type === "type_list");
+  return list ? [...list.namedChildren] : [...clause.namedChildren];
+}
+
+/**
+ * What a supertype entry is CALLED, or null when this pass cannot say.
+ *
+ * Type arguments are erased, because they are not part of the supertype's identity:
+ *
+ *     Base           |  Base<Item>          -> Base
+ *
+ * A qualified name is kept WHOLE rather than reduced to its final segment:
+ *
+ *     Outer.Inner    |  Outer.Inner<K>      -> Outer.Inner
+ *
+ * Heritage keeps an unresolved base as the edge target by design ("usually an
+ * external/imported type — keep the name"), so the full string is both truthful and
+ * unable to false-match a node id, where a bare `Inner` could collide with an
+ * unrelated in-repo type. That differs from construction (#103), which drops a
+ * qualified name instead — construction has no keep-the-name contract to fall back on.
+ */
+function javaSupertypeName(node: Parser.SyntaxNode | null | undefined): string | null {
+  if (!node) return null;
+  if (node.type === "generic_type") return javaSupertypeName(node.namedChildren[0]);
+  if (node.type === "scoped_type_identifier") return node.text;
+  return node.type === "type_identifier" ? node.text : null;
+}
+
+/** The names a declaration binds as its own type parameters (`class C<T, U>` → T, U),
+ * so they can never be mistaken for supertypes. */
+function javaTypeParameterNames(decl: Parser.SyntaxNode): ReadonlySet<string> {
+  const params = decl.childForFieldName("type_parameters");
+  if (!params) return new Set();
+  const out = new Set<string>();
   const visit = (n: Parser.SyntaxNode): void => {
-    if (n.type === "type_identifier") out.push(n.text);
+    if (n.type === "type_identifier") out.add(n.text);
     for (const c of n.namedChildren) visit(c);
   };
-  visit(node);
+  visit(params);
   return out;
 }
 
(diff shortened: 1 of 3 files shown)


dbianco:
Added a second commit with three tests for `javaReceiver`, unrelated to the heritage fix but found while checking whether the tests in this PR actually bite.

I ran mutation testing over the Java region of `extract.ts` (Stryker, command runner, full suite per mutant) and **every branch of `javaReceiver` survived** — mutating the `!obj` guard, the `identifier` branch, the `this` branch and the `field_access` branch each left the suite green. The reason is narrow: the only receiver any fixture exercised was a bare local (`store.save(...)`), so the other three shapes were never distinguished. That gap is from #83, so it is mine.

The three cases are `this.repo.run()` (field_access), `this.own()` (this — distinct from the implicit-this call already covered, which has no `object` node and never reaches the function), and `all[0].run()` (array_access, which must resolve to nothing). The fixture gives two classes a same-named `run()` so a wrong receiver lands on a visibly wrong target instead of just dropping.

Happy to split this into its own PR if you would rather keep this one to the heritage change — it is a separate commit, so it lifts out cleanly.

On the tool itself, in case it is of interest: I would **not** suggest adopting mutation testing here. With the command runner it reruns the whole suite per mutant — 22s on `main`, 40s on this branch, i.e. improving the suite makes the measurement more expensive. It was worth one manual run over one region; it would not be worth a gate. The overall score for the region was 69.91%, but a large share of the survivors are equivalent mutants (`inner?.type` → `inner.type` where `inner` cannot be null), so the raw number overstates the gap.

shhdwi:
Verified locally on top of current `main` (`extract.ts`/`resolve.ts` auto-merged; the only conflict was additive test appends in `graph-java.test.ts` vs the merged-in #161 anon test — rebased to keep both suites; build clean).

Before/after on the issue's fixture (`class Child extends Base<Item> implements Marker`):
- **Base main (bug):** `Child --extends--> Item` and `Sorted --implements--> Item` were emitted (the type ARGUMENT `Item` treated as a supertype), and `Child.use --calls--> Item.poison` — a poisoned call fed by the fake `extends`. Exactly #104's two failures.
- **This PR:** the bogus `Item` supertype edges are gone; real supertypes kept (`Child --extends--> Base`, `Sorted --implements--> Comparable`); `Child.use` only keeps the legit constructor edge.

`graph-java.test.ts` 32/32 (incl. #160's 4 heritage tests — type-argument, poisoned-call, type-variable, qualified-name — and the #161 anon test), `graph-resolve-typed`/`graph-php` 26/26. Fixes #104. Merging.

PR #174, opened 2026-08-21
fix(viz): an empty radius says why, and any PR's page can be republished

anirudhkumar-nanonets:
A PR that only touches lockfiles, workflows or docs has no blast radius, so the page published for it was a **blank canvas** behind a link promising a diagram — indistinguishable from a broken export. The graph now carries the reason and the viewer renders it:

> Nothing to draw: no parser claims 2 changed files (.github/workflows/blast-pages.yml, package-lock.json), so this change has no symbols to trace.

Changed code with no dependents gets a different sentence, because it is a different answer.

The note names changed files, and on a fork those paths are written by whoever opened the PR — it reaches a published page, so it is escaped rather than trusted.

`blast-pages.yml` also accepts a PR number via `workflow_dispatch`. A `pull_request` job on a fork gets a read-only token and cannot post the link itself, so publishing that page has to be startable by hand; the number is validated as digits so a dispatch cannot write outside its own directory.

This PR is also the first one whose own page is built with `--name`, so its Context tab should show concept names rather than hub symbols.

755 tests pass.

### DIFF
--- a/.github/workflows/blast-pages.yml
+++ b/.github/workflows/blast-pages.yml
@@ -11,16 +11,26 @@ name: Blast viewer
 # PR's files as data — never runs a script from the PR (no `npm ci` on the PR's
 # package.json, no postinstall). The graph is built from the PR's source text by a
 # tree-sitter parser, which executes none of it.
+#
+# A dispatch by PR number is the second entry point. It exists because a fork PR's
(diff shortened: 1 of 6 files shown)


anirudhkumar-nanonets:
Superseded — folding these two fixes into the panel-copy PR, so the published page's wording lands as one reviewable change.

PR #175, opened 2026-08-21
feat(php): mint anonymous classes as nodes with their implements edges (#144)

Frankie-Xu:
## Summary

Part 2 of #144 — attributes were wired as `references` edges in #155. Together with #155 they address #144 (external/framework attribute references still open).

An anonymous class (`new class implements I {…}`) previously produced no node and no heritage edge; its methods were mis-attributed to the enclosing function (`…#make.run`), so the type and its interface contract were invisible in the wiring graph.

- `describe()` now recognizes `anonymous_class` (a distinct node type in tree-sitter-php 0.23) and mints a `class` node named `{anonymous}`, mirroring the existing `{closure}` naming — deduplicated per file by `mintId` (`{anonymous}~2`, …).
- Because the descriptor's kind is `class`, the existing walk does the rest with no further changes: `heritageEdges` picks up `base_clause` / `class_interface_clause` (both are direct children of `anonymous_class`), so `implements` **and** `extends` edges are emitted; `enclosingClass` becomes `{anonymous}`, so nested methods get the right `owner` and nest under the class in the id path.
- `bindings.ts`'s `defName` mirrors the same scope segment (as it already does for closures), keeping the two scope stacks in sync for typed-parameter bindings inside anonymous-class methods.

Deliberately orthogonal to #155 (attribute `references` edges) and to trait resolution — this branch is based on `main` and only touches the anonymous-class path. Happy to rebase whichever lands second.

## Verification

`graft build` on the issue's exact repro now yields:

```
✓ wiring: 4 nodes (1 file, 1 function, 1 class, 1 method), 4 edges, 1 cards [php]
```

```json
{"id":"poc.php#make.{anonymous}","kind":"class","name":"{anonymous}","span":"L6-L8","signature":"class implements Runnable"}
{"id":"poc.php#make.{anonymous}.run","kind":"method","name":"run","span":"L7-L7","signature":"public function run(): int","owner":"{anonymous}"}

{"source":"poc.php#make","target":"poc.php#make.{anonymous}","relation":"contains","confidence":"extracted"}
{"source":"poc.php#make.{anonymous}","target":"poc.php#make.{anonymous}.run","relation":"contains","confidence":"extracted"}
{"source":"poc.php#make.{anonymous}","target":"Runnable","relation":"implements","confidence":"inferred"}
```

The stale `poc.php#make.run` id is gone; when the interface is defined in the project the `implements` edge resolves to its node (asserted in the tests).

## Test plan

- [x] New: anonymous class mints a `class` node; `run`'s owner is `{anonymous}`, and no `…#make.run` id exists
- [x] New: `implements` → `Runnable` and `extends` → `Base` edges resolve through `buildGraph`
- [x] New: two sibling anonymous classes in one scope dedupe to `{anonymous}` / `{anonymous}~2` (same rule as `{closure}`)
- [x] Control: named classes, variable-named closures, and bare `{closure}` callbacks unchanged
- [x] `npm run build && npm test` — 757/757 green

Part 2 of #144 (anonymous classes)

### DIFF
--- a/src/graph/bindings.ts
+++ b/src/graph/bindings.ts
@@ -71,6 +71,8 @@ export function defName(node: Parser.SyntaxNode, lang: Language): string | null 
     // Closures push a scope segment in extract.ts too — mirror it so a typed
     // parameter bound inside a closure is keyed under the same scope path.
     if (node.type === "anonymous_function" || node.type === "arrow_function") return phpClosureName(node);
+    // Anonymous classes likewise mint an `{anonymous}` scope segment (#144).
+    if (node.type === "anonymous_class") return "{anonymous}";
     return null;
   }
   const defTypes =
--- a/src/graph/extract.ts
+++ b/src/graph/extract.ts
@@ -608,6 +608,22 @@ function describe(node: Parser.SyntaxNode, ctx: WalkCtx): DefDescriptor | null {
     };
   }
 
+  // PHP anonymous classes (`new class implements I {…}`): minted as a class
+  // node named `{anonymous}` (mirroring `{closure}`, deduplicated per file by
+  // mintId). Without this the type vanished — no node, no heritage edge — and
+  // its methods mis-attributed to the enclosing function (issue #144). The
+  // class kind makes the walk emit heritageEdges (base_clause /
+  // class_interface_clause are direct children) and own the nested methods.
+  if (ctx.lang === "php" && node.type === "anonymous_class") {
+    const body = node.childForFieldName("body");
+    return {
+      name: "{anonymous}",
+      kind: "class",
+      headerEnd: body ? body.startIndex : node.endIndex,
+      hashNode: node,
+    };
+  }
+
   const mapped = ctx.kinds[node.type];
   if (mapped) {
     const name = node.childForFieldName("name")?.text;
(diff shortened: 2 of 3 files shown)


shhdwi:
Verified locally on top of current `main` (extract.ts auto-merged; rebased the additive `graph-php.test.ts` conflict against the merged enum #145 + attribute #155 tests, keeping all three suites; build clean; `graph-php` 15/15).

Reproduced #144 part 2 exactly: `function make(): Runnable { return new class implements Runnable {…}; }` — baseline mis-attributes `run` as `poc.php#make.run` with no anon node and no heritage edge; with this PR a `class` node `poc.php#make.{anonymous}` is minted, `run` nests under it with `owner={anonymous}` (stale `#make.run` gone), and an `implements #make.{anonymous} → #Runnable` edge is emitted. Two sibling anon classes dedupe to `{anonymous}`/`{anonymous}~2`.

Merging as part 2 of #144. Keeping #144 open: part 2 (anon classes) is now complete, and part 1 in-project attributes landed in #155, but external/framework attribute references remain unresolved — tracked in #144.

Frankie-Xu:
Conflict with current `main` was positional in `test/graph-php.test.ts` (#145 enum tests vs this PR's anonymous-class tests). Both blocks stay.

Did not force-push: the fork branch already has a maintainer merge onto `main` (`7c9b101`, `pr175r`) that keeps both test suites — same resolution, different test order. Local rebase is 865/865 if a clean rebase head is useful later.

PR #176, opened 2026-08-21
fix(java): anonymous-class methods no longer take the enclosing type's owner (#161)

### LINKED ISSUE #161: Java: anonymous-class methods take the enclosing type's owner, so a real call resolves to the anonymous impl
Split out from #104 as suggested, so it does not widen that PR's blast radius. Reproduced against current `main`.

## What happens

A method declared inside an **anonymous class body** is minted with `owner` set to the *enclosing* type, not to the anonymous implementation. `owner` is what feeds `resolveTypedMember`'s `ownerMethod` index, so the enclosing type's method set gains members it does not have.

```java
public final class App {
  public Greeter make() {
    return new Greeter() {
      @Override public String greet(String n) { return "anon:" + n; }
    };
  }
  public String use() {
    App self = new App();
    return self.greet("x");
  }
  public String greet(String n) { return "real:" + n; }
}
```

Emitted nodes:

```
App.greet         owner='App'    L12
App.make.greet    owner='App'    L5     <- the anonymous impl, same owner
Greeter.greet     owner='Greeter'
```

`ownerMethod["App.greet"]` therefore holds two candidates. `resolveTypedMember` narrows by arity (both take one argument, so no help) and then takes the first same-file candidate:

```
App.use --calls[extracted]--> App.make.greet
```

The anonymous implementation, at the highest confidence level. The real `App.greet` ends with **zero** inbound edges, so asking who calls it returns nothing.

## Incidence

Counting method nodes whose *containing scope segment is itself a method node* — which distinguishes an anonymous/local class body from a nested class, whose owner is correct:

| repo | methods | declared in another method's body | wrong owner | of those, colliding with a real homonym |
|---|---:|---:|---:|---:|
| [`google/gson`](https://github.com/google/gson) | 3,431 | 158 | **158** | **96** |
| [`spring-projects/spring-petclinic`](https://github.com/spring-projects/spring-petclinic) | 187 | 0 | 0 | 0 |

Every anonymous-body method in gson carries the wrong owner, and 96 of them share `(owner, name)` with a real method of the enclosing type — each one an ambiguity resolved by document order. Samples:

```
RuntimeTypeAdapterFactory.create.read   owner=RuntimeTypeAdapterFactory
RuntimeTypeAdapterFactory.create.write  owner=RuntimeTypeAdapterFactory
JavaTimeTypeAdapters.localDateTime.read owner=JavaTimeTypeAdapters
```

petclinic has none, so this is concentrated in the adapter/callback style rather than universal — which is also why gson is the repo to measure it on.

## Note on scope

The `resolveName` same-file guard in #160 does **not** cover this. That guard is in `resolveName`; this path runs through `resolveTypedMember`, which has its own `candidates.find(c => c.path === file)` first-wins tiebreak. Fixing `owner` is the root-cause fix; tightening `resolveTypedMember` would only turn a wrong edge into a dropped one, and would also drop legitimate cases.

## Direction

An anonymous class body is a distinct type with no name, so its members should not join the enclosing type's method set. Two shapes worth choosing between before I write it:

1. **Give the anonymous body a synthetic owner** derived from the interface/class being instantiated (`Greeter` here), so `new Greeter(){…}.greet` joins `Greeter`'s method set. Arguably truer — that *is* the type it implements — but it can then satisfy `greeter.greet()` calls that are genuinely dispatched to it.
2. **Emit no `owner` at all** for members of an anonymous body. Strictly conservative: the node still exists and is findable, it simply never participates in owner-qualified resolution. Loses the dispatch edges (1) would gain.

My lean is (2) for the same reason #103 and #160 landed conservative: a dropped edge is recoverable, a wrong one at `extracted` confidence is not. But (1) is defensible and you may have a view on whether anonymous implementations should be reachable through their interface.

Happy to implement whichever, same bar as #160 — fixtures that fail when reverted, before/after counts on both repos above.

Frankie-Xu:
## Summary
- Java anonymous class bodies (`new Type() { … }`) now mint an `{anonymous}` class node (same naming as PHP #144 / PR #175), so methods inside take `owner: "{anonymous}"` instead of the enclosing type.
- That stops them from joining `ownerMethod["Enclosing.method"]`, which previously made `resolveTypedMember`'s same-file first-wins tiebreak steal call edges from the real method (issue repro: `App.use` → anonymous `greet`, real `App.greet` had zero inbound).
- Also emits an `implements` edge from `{anonymous}` to the constructed type when it is an interface.

**Out of scope / no overlap with #160:** does not touch generic heritage / `javaSupertypeName` / `resolveName` same-file guard. Edit is only the `object_creation_expression` + `class_body` walk path in `extract.ts`, plus the #161 fixture in `graph-java.test.ts`. Safe to land independently of dbianco's #104 PR.

## Incidence (measured by @dbianco on #161)

Method nodes whose containing scope segment is itself a method — anonymous/local bodies, not nested classes (those already had the right owner):

| repo | methods | in another method's body | wrong owner | colliding with a real homonym |
|---|---:|---:|---:|---:|
| gson | 3,431 | 158 | **158** | **96** |
| spring-petclinic | 187 | 0 | 0 | 0 |

gson samples: `RuntimeTypeAdapterFactory.create.read`, `…create.write`, `JavaTimeTypeAdapters.localDateTime.read` — all `owner=<enclosing type>`. The 96 are cases where `(owner, name)` collided with a real method and document order decided the edge. petclinic has none.

Source: [@dbianco on this PR](https://github.com/NanoNets/Graft/pull/176#issuecomment-5372012988), counted when filing #161.

## Test plan
- [x] `node --import tsx --test test/graph-java.test.ts` — new #161 case green (App.use → real App.greet; anon owner ≠ App; Nested control unchanged; implements Greeter)
- [x] `npm run build && npm test` — 755/755 pass, including `graph-resolve-typed.test.ts`

Closes #161

### DIFF
--- a/src/graph/extract.ts
+++ b/src/graph/extract.ts
@@ -466,6 +466,55 @@ function walk(node: Parser.SyntaxNode, ctx: WalkCtx, out: NodeV1[], edges: RawEd
     }
   }
 
+  // Java anonymous class (`new Type() { … }`): tree-sitter-java has no
+  // `anonymous_class` node (unlike PHP) — the body is an optional `class_body`
+  // on `object_creation_expression`. Mint `{anonymous}` (mirroring PHP #144 /
+  // `{closure}`) so nested methods take that owner instead of the enclosing
+  // type's, which otherwise pollutes `ownerMethod` and steals real call edges
+  // (#161). The constructor call edge above still fires for `new Type()`.
+  if (ctx.lang === "java" && node.type === "object_creation_expression") {
+    const body = node.namedChildren.find((c) => c.type === "class_body");
+    if (body) {
+      const idPart = "{anonymous}";
+      const base = `${ctx.rel}#${[...ctx.scope, idPart].join(".")}`;
+      const id = mintId(base, minted);
+      out.push({
+        id,
+        name: "{anonymous}",
+        kind: "class",
+        path: ctx.rel,
+        span: `L${node.startPosition.row + 1}-L${node.endPosition.row + 1}`,
+        signature: clean(ctx.source.slice(node.startIndex, body.startIndex)),
+        exported: false,
+        origin: "ast",
+        body_hash: contentHash(node.text),
+        body_text: searchBody(node.text),
+        summary_state: "pending",
+        summary: null,
+        crux: null,
+      });
+      edges.push({ source: ctx.parentId, relation: "contains", targetId: id, file: ctx.rel });
+      // Single supertype from `new Type()`: emit `implements` so an interface
+      // target resolves (adapters are the common case; a class target drops
+      // under resolve's implements kind filter — drop-not-guess).
+      const superName = javaConstructedTypeName(node.childForFieldName("type"));
+      if (superName) {
+        edges.push({ source: id, relation: "implements", name: superName, file: ctx.rel });
+      }
+      const anonCtx: WalkCtx = {
+        ...ctx,
+        scope: [...ctx.scope, idPart],
+        enclosingKind: "class",
+        parentId: id,
+        enclosingClass: "{anonymous}",
+      };
+      for (const child of node.namedChildren) {
+        walk(child, child.type === "class_body" ? anonCtx : ctx, out, edges, minted);
+      }
+      return;
+    }
+  }
+
   for (const child of node.namedChildren) walk(child, ctx, out, edges, minted);
 }
 
--- a/test/graph-java.test.ts
+++ b/test/graph-java.test.ts
@@ -918,3 +918,97 @@ test("Java scopes: pom.xml, build.gradle, and build.gradle.kts are project marke
     }
   }
 });
+
+/**
+ * Issue #161: a method inside an anonymous class body (`new Greeter() { … }`)
+ * used to take the enclosing type's `owner`, so `ownerMethod["App.greet"]` held
+ * both the real method and the anonymous override. `resolveTypedMember` then
+ * same-file first-wins tied to the anonymous impl, and `App.use → App.greet`
+ * vanished. Named nested classes already own their methods correctly — keep
+ * that as a control so this fix does not retarget lexical nesting.
+ *
+ * Shape mirrors PHP #144: mint `{anonymous}` and hang methods under it.
+ */
+const ANON_GREETER = `package com.acme;
+
+public interface Greeter {
+  String greet(String n);
+}
+`;
+
+const ANON_APP = `package com.acme;
+
+public final class App {
+
+  public Greeter make() {
+    return new Greeter() {
+      @Override public String greet(String n) { return "anon:" + n; }
+    };
+  }
+
+  public String use() {
+    App self = new App();
+    return self.greet("x");
+  }
+
+  public String greet(String n) { return "real:" + n; }
+
+  public static class Nested {
+    public String greet(String n) { return "nested:" + n; }
+  }
+}
+`;
+
+function anonOwnerFixture(): string {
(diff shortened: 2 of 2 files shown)


dbianco:
Thank you for picking this up — and the `{anonymous}` node is a better answer than either option I left in #161.

I offered two: a synthetic owner taken from the instantiated interface, or no owner at all. Yours is a third I didn't think of, and it beats both. Minting the body as its own class node means the anonymous implementation stays reachable instead of being erased, and the `implements` edge to the constructed type gives it a non-`contains` edge — so it doesn't land in the Tier-0 orphan count, which was the trap I'd have walked into with "no owner". It also matches what you already did for PHP, so Java isn't a special case.

## The failing check is not your code

```
##[error]Refusing to check out fork pull request code from a 'pull_request_target' workflow.
```

`publish` runs on `pull_request_target`, which executes with repository secrets, so `actions/checkout` refuses to check out fork code — deliberately, since otherwise a fork PR could run arbitrary code with those secrets. It fails identically on your #175, and it doesn't appear at all on PRs opened before that workflow landed.

Nothing you can do about it from a fork. It needs either an `if` that skips forks or a move to `pull_request` on the repo side — and it will hit every external contributor until then. Might be worth its own issue.

## No collision with #160

I tested it locally rather than assuming: merged your branch into mine.

```
src/graph/extract.ts       auto-merged clean
test/graph-java.test.ts    CONFLICT
```

The conflict is purely positional — we both append tests to the end of the same file. Nothing semantic: your change lives in the `object_creation_expression` + `class_body` walk, mine in heritage and the `resolveName` same-file guard. Whoever lands second keeps both blocks. Your "safe to land independently" note is correct.

I've just rebased #160 onto current `main` (844/844) so it isn't the stale one.

## Numbers, if useful for a before/after

I measured the incidence when filing #161, counting method nodes whose containing scope segment is itself a method node — which separates anonymous/local bodies from nested classes, whose owner was already correct:

| repo | methods | in another method's body | wrong owner | colliding with a real homonym |
|---|---:|---:|---:|---:|
| gson | 3,431 | 158 | **158** | **96** |
| spring-petclinic | 187 | 0 | 0 | 0 |

Samples from gson: `RuntimeTypeAdapterFactory.create.read`, `…create.write`, `JavaTimeTypeAdapters.localDateTime.read` — all carrying `owner=<enclosing type>`. The 96 are the ones where `(owner, name)` collided with a real method and document order decided the edge. petclinic has none, so gson is the repo where this shows up at all.

Take them if they're useful for the PR body.

shhdwi:
Verified locally on top of current `main` (clean merge — the change is disjoint from recent telemetry work; `npm run build` clean; `graph-java.test.ts` 24/24 incl. the new #161 test).

Independent CLI repro of the issue's snippet:
- **Before (main):** the anon method was minted as `App.make.greet` with owner `App`, and `App.use --calls--> App.make.greet` (the anonymous impl) — the real `App.greet` had zero callers. Reproduces #161 exactly.
- **After (this PR):** `App.make.{anonymous}.greet` owner `{anonymous}`; `App.greet` owner `App` (unchanged); `App.make.{anonymous} --implements--> Greeter`; and `App.use --calls--> App.greet` resolves to the **real** method, with no edge to the anonymous impl.

Named nested classes are unaffected (`Nested.greet` owner `Nested`). Fixes #161. Merging.

Frankie-Xu:
@dbianco thanks — the no-collision result matches what I expected (positional test append only). Pulled the gson / petclinic table into the PR body under **Incidence**.

On the red `publish` check: agreed, that's a repo-side `pull_request_target` guard, not this diff. Opened #193 so it isn't buried in a merged PR.

Issue #129, opened 2026-08-17
synth/crux read results only from `toolCalls`; gateways that ignore forced `tool_choice` silently produce empty/degraded graphs


Rajahn:
`ChatSynthesizer` / `ChatCruxSummarizer` take results exclusively from `res.toolCalls[0]`. Behind an OpenAI-compatible corporate gateway we observed two failure modes, both silent:

1. **deepseek-v3.2** ignores forced `tool_choice` and emulates the tool call in `content` instead:
   `content: '[{"name": "emit_json", "parameters": {...}}]'`, `tool_calls: null`.
   → `res.toolCalls[0]` is undefined → `synthesize` returns `[]`, no warning.
2. **qwen3-vl-235b-a22b-instruct** does emit real `tool_calls`, but its synthesis quality degraded badly on the same inputs: one full run produced 23 synthesis batches → **15 nodes, 0 links** (mostly sourced from test files), silently replacing a concept layer that deepseek-v3.2 had produced as 28 well-linked nodes. Nothing per-batch is logged, so the degradation is invisible.

Net effect: `graft build --deep` "succeeds" with a near-empty or badly degraded concept layer, and neither the build log nor `graft check` says anything is wrong.

Evidence a fallback is viable: we implemented content-JSON recovery out-of-tree (strip fences, bracket-match the first `{...}`, parse, validate against the same schema) using the identical SYSTEM_PROMPT and batching with deepseek-v3.2: 19 batches → **207 nodes / 266 links**, merged cleanly.

Request: when `toolCalls` is empty, fall back to parsing JSON from `content`, and surface a warning when a synthesis batch yields 0 nodes — today both failure modes are completely silent.

Environment: graft 0.10.1, Node 22, OpenAI-compatible corporate gateway (internal URL omitted); models: deepseek-v3.2, qwen3-vl-235b-a22b-instruct.

shhdwi:
[...]

PR #177, opened 2026-08-21
fix(deep): retry pending nodes — don't cache empty meaning summaries (#172)

### LINKED ISSUE #172: build --deep leaves nodes permanently `pending`; graft check then reports an unresolvable "not yet summarized" state
## Summary

`graft build --deep` can finish successfully while leaving nodes with `summary: null`, reported as `pending`. `graft check` then counts those nodes and prints:

```
(N node(s) not yet summarized — run `graft build --deep`)
```

But re-running `graft build --deep` does not change the count. There is no command that clears it, so the repo is stuck in a permanently "incomplete" state. Anything that gates on a clean `graft check` (CI, a pre-commit hook) can never pass.

The LLM path itself is working in the same run — the concept pass calls the model and succeeds. Only the meaning/crux tier stays at zero.

## Minimal reproduction

```bash
mkdir /tmp/graft-repro && cd /tmp/graft-repro && git init -q
mkdir src
cat > src/calc.py <<'EOF'
"""Tiny module for the repro."""


def add(a, b):
    """Return the sum."""
    return a + b


def mul(a, b):
    """Return the product."""
    return a * b


def total(items):
    """Sum a list using add()."""
    out = 0
    for x in items:
        out = add(out, x)
    return out
EOF
git add -A && git commit -qm init

export GRAFT_PROVIDER=openai
export GRAFT_BASE_URL=https://api.together.ai/v1
export GRAFT_MODEL=deepseek-ai/DeepSeek-V4-Flash-0731
export GRAFT_API_KEY=...

npx -y @nanonets/graft@0.10.1 build --deep
npx -y @nanonets/graft@0.10.1 build --deep              # again, no edits
npx -y @nanonets/graft@0.10.1 build --deep --no-reuse
rm -f graft/.cache/summaries.json
npx -y @nanonets/graft@0.10.1 build --deep
npx -y @nanonets/graft@0.10.1 check
```

## Observed

Every one of those `--deep` runs prints the same line:

```
✓ concepts: 1 nodes, 0 links from 1 files (1 read, 0 cached)
✓ wiring: 4 nodes (3 function, 1 file), 4 edges, 1 cards [python]
  meaning: 0 computed, 0 cached, 0 stale, 4 pending
```

`meaning: 0 computed` on the very first run, with an empty cache. And `check`:

```
graft check: OK — the graph is in sync with the code.
graph check: OK — the wiring graph is in sync with the code. (4 node(s) not yet summarized — run `graft build --deep`)
```

All four nodes carry `summary: null` in `graft/.graph/wiring.json`:

```
{'id': 'src/calc.py',        'kind': 'file',     'summary': None}
{'id': 'src/calc.py#add',    'kind': 'function', 'summary': None}
{'id': 'src/calc.py#mul',    'kind': 'function', 'summary': None}
{'id': 'src/calc.py#total',  'kind': 'function', 'summary': None}
```

## What does not help

- Re-running `build --deep` (any number of times)
- `build --deep --no-reuse` — this re-parses via tree-sitter but does not appear to touch the meaning tier
- Deleting `graft/.cache/summaries.json`
- Deleting the whole `graft/` directory and rebuilding

## Second observation, larger repo

On a real repo (7 Python files, ~29 nodes) the meaning tier does mostly work — but one node stays behind in the same way:

```
meaning: 0 computed, 28 cached, 1 stale, 0 pending
```

`check` named the file, and repeated `--deep` runs (with and without `--no-reuse`) did not clear it. There, deleting `graft/.cache/summaries.json` *did* fix it — one node recomputed and the count went to zero. That workaround does **not** work in the minimal repro above, so the two may not share a cause.

## Why it matters

`graft check`'s message is an instruction (`run graft build --deep`) that cannot resolve the condition it reports. If the count is expected to be non-zero in some cases — for example if trivially small symbols are intentionally skipped — then `check` probably should not surface them as actionable work, since a user or a CI job cannot tell that state apart from a real failure.

## Environment

- graft 0.10.1 (via `npx -y @nanonets/graft@0.10.1`)
- macOS 15 (arm64), Node 24
- Provider: Together AI serverless, `deepseek-ai/DeepSeek-V4-Flash-0731`, OpenAI wire format
- The same key and model produce concept nodes successfully in the same runs, so this is not an auth or quota failure.

Related but distinct: #128 (crux results lost on interrupted runs) and #127 (deep build swallowing per-file LLM failures).

Frankie-Xu:
## Summary

`graft build --deep` could finish successfully while leaving every wiring node `summary: null` / `pending`, then `graft check` told you to re-run `--deep` forever. That is exactly [#172](https://github.com/NanoNets/Graft/issues/172).

### Root cause

Two silent paths in the meaning tier (not thrown provider errors — those are already covered by #127):

1. **Empty reply treated as success.** When `describeFile` returned `[]` (gateway ignored `tool_choice` and left `toolCalls` empty, or the model omitted every symbol), `collectFileCrux` returned `{ results: ∅ }` with **no error**. Enrich left nodes `pending`, `failedFiles` stayed 0, the CLI exited 0, and check’s only advice was “run `graft build --deep`”.
2. **Blank summaries cached as `ready`.** `node.summary = r.summary || null` still set `summary_state: "ready"`, so the next `--deep` was a permanent body-hash cache hit with `summary: null`.

The reporter’s Together / DeepSeek-V4-Flash setup is the same *trigger* as [#129](https://github.com/NanoNets/Graft/issues/129) (content instead of tool calls). This PR is the **meaning-tier / pending-state** fix: fail loudly, don’t cache empties, recover crux payloads from content when present, and make check name the stuck nodes. Full synth-side recovery for #129 is still separate.

### Changes

- **`enrich`:** total miss → error + `failedFiles` (CLI #127 path exits 1); whitespace-only summaries stay `pending` and remain retryable.
- **`ChatCruxSummarizer`:** if `toolCalls` is empty, recover `{symbols:…}` from `content` (plain JSON, fences, or `[{name, parameters}]` emulation).
- **`graft check`:** list pending ids and say that if a deep build already left them pending, look at that build’s errors — re-running alone will not clear a failed meaning pass.

## Test plan

- [x] `test/graph-enrich-pending.test.ts` — empty reply → `failedFiles`; blank summary not cached as ready; check wording; content recovery
- [x] `npm run build && npm test` — 758 pass
- [ ] Optional with a real key: issue repro (`src/calc.py`) → `build --deep` either fills meaning or exits non-zero with a readable error; `graft check` names pending nodes if any remain

Closes #172

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@nanonets/graft",
-  "version": "0.11.0",
+  "version": "0.12.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@nanonets/graft",
-      "version": "0.11.0",
+      "version": "0.12.0",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
--- a/src/ai/crux.ts
+++ b/src/ai/crux.ts
@@ -114,6 +114,66 @@ function parseResults(obj: { symbols?: unknown } | undefined): NodeCrux[] {
     }));
 }
 
+/**
+ * Some OpenAI-compatible gateways ignore forced `tool_choice` and put the tool
+ * payload in `content` instead (plain `{symbols:…}`, fenced JSON, or an emulated
+ * `[{name, parameters}]` array). Without this recovery the meaning pass sees an
+ * empty `toolCalls` list, leaves every node `pending`, and `graft check` loops
+ * on "run --deep" forever (#172; same trigger as #129 for the crux path).
+ */
+function argsFromResponse(res: { text: string; toolCalls: { name: string; args: unknown }[] }): {
+  symbols?: unknown;
+} | undefined {
+  const call = res.toolCalls.find((c) => c.name === RECORD_TOOL) ?? res.toolCalls[0];
+  if (call?.args && typeof call.args === "object" && !Array.isArray(call.args)) {
+    return call.args as { symbols?: unknown };
+  }
+  return recoverArgsFromContent(res.text);
+}
+
+function recoverArgsFromContent(text: string): { symbols?: unknown } | undefined {
+  const raw = text?.trim();
+  if (!raw) return undefined;
+  const stripped = raw.replace(/^```(?:json)?\s*/i, "").replace(/\s*```$/i, "").trim();
+
+  const asSymbols = (value: unknown): { symbols?: unknown } | undefined => {
+    if (!value || typeof value !== "object") return undefined;
+    if (Array.isArray(value)) {
+      for (const item of value) {
+        if (!item || typeof item !== "object") continue;
+        const rec = item as Record<string, unknown>;
+        const name = typeof rec.name === "string" ? rec.name : "";
+        // Accept the crux tool, or a generic emit_json wrapper some gateways use.
+        if (name && name !== RECORD_TOOL && name !== "emit_json") continue;
+        const params = rec.parameters ?? rec.arguments ?? rec.args;
+        const hit = asSymbols(params);
+        if (hit) return hit;
+      }
+      return undefined;
+    }
+    const obj = value as Record<string, unknown>;
+    if (Array.isArray(obj.symbols)) return obj as { symbols?: unknown };
+    return undefined;
+  };
+
+  try {
+    const hit = asSymbols(JSON.parse(stripped));
+    if (hit) return hit;
+  } catch {
+    /* try bracket slice below */
+  }
+  const start = stripped.indexOf("{");
+  const end = stripped.lastIndexOf("}");
+  if (start >= 0 && end > start) {
+    try {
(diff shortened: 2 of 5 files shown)


shhdwi:
Verified locally on top of current `main` (rebased past a trivial `package-lock.json` version-string conflict — took main's `0.12.1`; the PR changes no deps; build clean).

Fixes #172's two silent-success paths in `enrich.ts`: (1) a blank summary no longer flips a node to `ready` with `summary=null` — the guard is now `if (!r || !r.summary.trim())`, so the node stays `pending` and retryable; (2) a total-miss now surfaces a `fileError` into the degraded-exit gate, and `check.ts` reports the stuck node ids.

Load-bearing evidence: the PR's 4 new #172 tests **fail 0/4 against base main** (base emits the exact stuck `"…not yet summarized; run graft build --deep"`) and **pass 4/4 with the PR**; a working summarizer retries the pending nodes next run (`computed=4, cached=0`). `graph-enrich*` all green. Fixes #172. Merging.

PR #180, opened 2026-08-21
feat: quote the line where the edge happens — in the comment, the CLI, and one shared helper

anirudhkumar-nanonets:
Three surfaces claimed "A calls B" and made you open a file to see it. They now quote the line, from one helper.

**One evidence helper.** The hunk/reach-line shaping moved out of `blast/viz.ts` into `blast/evidence.ts`, with a generalised `referenceLine(path, span, needles, read)` — find the line inside a span that mentions any of these. The page, the comment and `graft callers` all call it, so they cannot quote different lines for the same edge.

**The PR comment.** Its collapsed symbol list cited `src/cli.ts:L1-L866` and stopped there; it now adds the reaching line underneath. Still collapsed, still below the diagram — the comment is five circles and a table, and this is the part nobody reads by default. Asserted in the test: the quote must appear *after* `<details>` and never above the mermaid block.

**`graft callers`.**
```
markdownReport · function · src/blast/render.ts:L111-L146
  calls ← runBlastCommand (src/blast/blast-cli.ts:L58-L104)
      103: process.stdout.write(format === "markdown" ? markdownReport(report, { root }) : textReport(report));
```
Only for a resolved **first-hop** hit: a depth-2 hit references whatever sits between it and the symbol you asked about, so quoting it would point at the wrong line. Never in `--json`, which is a data contract.

**`viz --export --tabs`.** `blast --export-viz` already ships the Context tab alone; this exposes the same control to `viz`. A typo'd tab name exits 1 rather than quietly exporting a page with a tab missing and nothing to say why.

Every read is best-effort — a file that moved costs the snippet, never the command.

764 tests pass.


shhdwi:
Verified locally on top of current `main`. One trivial conflict in `src/cli.ts` (viz action) — combined this PR's `--tabs` option/param with main's telemetry `noteQuery(...)` wrapper. Build clean.

Feature works as claimed: `graft callers greet` now prints the quoted call-site line under each edge, via the shared `referenceLine()` helper in `src/blast/evidence.ts`; `--json` correctly omits the quote (data contract preserved). Touched-file tests pass — `blast-render` + `viz-export` 8/8 once the repo's (gitignored) `graft/` self-index is built (the one initial "failure" was just the missing index in a bare worktree, not a code bug). Merging.

PR #188, opened 2026-08-24
fix(openai): negotiate the four 400s reasoning-default models return

### LINKED ISSUE #183: graft build --deep fails with GPT-5.6 / o-series reasoning models (max_tokens, temperature, tool+reasoning_effort)
### Summary

`graft build --deep` fails outright when pointed at OpenAI's current reasoning-family models (tested with `gpt-5.6-luna` via `GRAFT_PROVIDER=openai`). Three separate 400s come back from `src/ai/llm/openai.ts`, each because the adapter sends a Chat Completions parameter that this model family rejects. All three block the whole `--deep` pass (summarize, crux, and synthesize all use the same adapter).

### Environment

- graft `0.12.0` (built from source, this repo)
- `GRAFT_PROVIDER=openai`, `GRAFT_MODEL=gpt-5.6-luna`, default `api.openai.com` base URL
- Real repo, ~150 TS/TSX files, default `--concurrency`

### Errors observed, in the order they surface

1. **`max_tokens`**
   ```
   400 Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
   ```
   `src/ai/llm/openai.ts` unconditionally sets `params.max_tokens = req.maxTokens` (line ~111). This model family requires `max_completion_tokens`.

2. **`temperature`** (surfaces once #1 is worked around)
   ```
   400 Unsupported value: 'temperature' does not support 0 with this model. Only the default (1) value is supported.
   ```
   `summarize.ts` / `crux.ts` / `synthesize.ts` all hardcode `temperature: 0` for determinism. This model only accepts the default (no `temperature` field at all).

3. **Tool calls + reasoning effort** (surfaces once #1 and #2 are worked around)
   ```
   400 Function tools with reasoning_effort are not supported for gpt-5.6-luna in /v1/chat/completions. To use function tools, use /v1/responses or set reasoning_effort to 'none'.
   ```
   Both `crux.ts` and `synthesize.ts` force tool-calling (`responseFormat: { kind: "tool" }`). On `/v1/chat/completions`, this model's default reasoning effort conflicts with function tools; the API's own error message names the fix.

### Suggested fix

`createChatCompletion` already has a narrow, error-message-driven fallback for one OpenAI-compatible quirk (`isRejectedObjectToolChoice` → retry with `tool_choice: "required"`). The same pattern extends cleanly to all three errors above — detect the specific 400 by message, retry once with the corrected param, never guess from the model name (so older endpoints/models that still expect `max_tokens`/custom `temperature` are untouched).

I have a working patch against `src/ai/llm/openai.ts` doing exactly this (bounded retry loop, one fix applied per detected error, falls through to the original error if nothing matches):

```diff
--- a/src/ai/llm/openai.ts
+++ b/src/ai/llm/openai.ts
@@ -85,6 +85,43 @@ function isRejectedObjectToolChoice(err: unknown): boolean {
   );
 }
 
+/**
+ * Newer reasoning-family models (o1/o3/o4, the gpt-5.x line, …) reject the
+ * classic `max_tokens` param outright and require `max_completion_tokens`
+ * instead, even though both are still in wide use across OpenAI-compatible
+ * servers. Detect the specific 400 rather than guessing from the model name,
+ * so older endpoints that still expect `max_tokens` are untouched.
+ */
+function isRejectedMaxTokens(err: unknown): boolean {
+  return (
+    err instanceof OpenAI.APIError &&
+    err.status === 400 &&
+    /max_tokens.*not supported.*max_completion_tokens/i.test(String((err as { message?: string }).message ?? ""))
+  );
+}
+
+/** Same reasoning-family models: temperature is fixed at 1, not caller-settable. */
+function isRejectedTemperature(err: unknown): boolean {
+  return (
+    err instanceof OpenAI.APIError &&
+    err.status === 400 &&
+    /temperature.*does not support/i.test(String((err as { message?: string }).message ?? ""))
+  );
+}
+
+/**
+ * Same models again: function tools are rejected on /v1/chat/completions while
+ * the model's default reasoning effort is active. The API's own error message
+ * names the fix — `reasoning_effort: "none"` — so apply exactly that.
+ */
+function isRejectedToolsWithReasoning(err: unknown): boolean {
+  return (
+    err instanceof OpenAI.APIError &&
+    err.status === 400 &&
+    /function tools with reasoning_effort/i.test(String((err as { message?: string }).message ?? ""))
+  );
+}
+
 export class OpenAIChatModel implements ChatModel {
   readonly label: string;
   private client: OpenAI;
@@ -141,14 +178,34 @@ export class OpenAIChatModel implements ChatModel {
    * automatically.
    */
   private async createChatCompletion(params: ChatParams): Promise<OpenAI.Chat.Completions.ChatCompletion> {
-    try {
-      return await this.client.chat.completions.create(params);
-    } catch (err) {
-      if (isRejectedObjectToolChoice(err) && typeof params.tool_choice === "object" && params.tools?.length === 1) {
-        return this.client.chat.completions.create({ ...params, tool_choice: "required" });
+    let attempt = params;
+    // Bounded: one retry per known incompatibility below, never an open loop.
+    for (let i = 0; i < 4; i++) {
+      try {
+        return await this.client.chat.completions.create(attempt);
+      } catch (err) {
+        if (isRejectedObjectToolChoice(err) && typeof attempt.tool_choice === "object" && attempt.tools?.length === 1) {
+          attempt = { ...attempt, tool_choice: "required" };
+          continue;
+        }
+        if (isRejectedMaxTokens(err) && attempt.max_tokens !== undefined) {
+          const { max_tokens, ...rest } = attempt;
+          attempt = { ...rest, max_completion_tokens: max_tokens } as ChatParams;
+          continue;
+        }
+        if (isRejectedTemperature(err) && attempt.temperature !== undefined) {
+          const { temperature, ...rest } = attempt;
+          attempt = rest as ChatParams;
+          continue;
+        }
+        if (isRejectedToolsWithReasoning(err)) {
+          attempt = { ...attempt, reasoning_effort: "none" } as ChatParams;
+          continue;
+        }
+        throw err;
       }
-      throw err;
     }
+    return this.client.chat.completions.create(attempt);
   }
 
   private fromResponse(
```

Happy to open this as a PR instead if that's more useful than a patch pasted into an issue — let me know.

### Unrelated but adjacent: rate limits on lower tiers

Not a graft bug, just a heads-up: `gpt-5.6-luna` on a tier-1/2 org caps at 200k TPM, separate from any daily token allowance. The default `--concurrency 5` for the deep pass burst past that within the first few seconds on a ~150-file repo, and the existing "5 consecutive failures → abort" circuit breaker (`ai/failure.ts`) then killed the whole run rather than backing off. Dropping to `--concurrency 2` worked around it. Might be worth a line in the README/`--deep` docs about lowering `--concurrency` for lower-tier accounts, since the abort-on-429 behavior otherwise looks like a hard failure rather than a pacing issue.

### Result once patched

With the patch above, `--deep` completed cleanly end-to-end and produced genuinely useful concept nodes from a real repo — e.g. one node's generated summary: *"A snapshot download plan is a constrained projection of repository state: every selected blob must be declared by the manifest, reside under wiki/public, and be a supported regular file; private paths, traversal, symlinks, unsupported files, missing entries, and incomplete trees must never enter the plan."* So the fix is worth landing — the model family works well once these three params are handled.

nguyenduyhung1989:
Fixes #183.

`graft build --deep` fails on the first LLM call against models that reason by default. Each of these is a request shape the server rejects with a 400, and each is detected by that 400 rather than guessed from a model name, so endpoints that still expect the older shape are untouched:

| rejected | remedy |
|---|---|
| `max_tokens` | retry as `max_completion_tokens` |
| a caller-set `temperature` | drop it (fixed at 1, not settable) |
| function tools while reasoning is on | `reasoning_effort: "none"` |
| DeepSeek's `Thinking mode does not support this tool_choice` | same `reasoning_effort` remedy |

The single try/catch becomes a bounded loop, because one response can surface a second incompatibility once the first is fixed.

**Ordering matters.** DeepSeek's message also names `tool_choice`, so the existing object-form fallback matched it first and rewrote the request to `tool_choice: "required"` — which thinking mode rejects too. The reasoning check therefore runs first. It is also the better remedy: loosening to `"auto"` returns 200 but lets the model decline to call the tool at all, and since `fmt.kind === "json"` uses a forced synthetic tool as the structured-output mechanism, a model that opts out returns prose and the node write fails downstream.

Every branch is gated on the parameter it removes still being present, and the reasoning branch on `reasoning_effort` being unset, so no branch can fire twice. The loop bound is one per known incompatibility; the trailing call exists only so a genuine error still surfaces rather than being swallowed by loop exhaustion.

The detectors match error text rather than model ids on purpose. That is brittle if a provider rewords a message, but a model-name allowlist goes stale every time a provider ships a model and silently sends the wrong shape to anything unrecognised. A wording change degrades to today's behaviour — a surfaced 400. A stale allowlist degrades to a broken build with no clue why.

**Verified end to end** on a repo that previously failed at the first LLM call: `build --deep` completes and writes concept nodes against both `gpt-5.6` and `deepseek-v4-flash`. `tsc --noEmit` clean, existing suite green.

The issue also notes a separate rate-limit problem (`-j 5` trips 429s on a 200k TPM key, and the consecutive-failure gate aborts the build rather than backing off). That is not touched here — happy to open it separately if you agree it is one.

### DIFF
--- a/src/ai/llm/openai.ts
+++ b/src/ai/llm/openai.ts
@@ -85,6 +85,51 @@ function isRejectedObjectToolChoice(err: unknown): boolean {
   );
 }
 
+/**
+ * Newer reasoning-family models (o1/o3/o4, the gpt-5.x line, …) reject the
+ * classic `max_tokens` param outright and require `max_completion_tokens`
+ * instead, even though both are still in wide use across OpenAI-compatible
+ * servers. Detect the specific 400 rather than guessing from the model name,
+ * so older endpoints that still expect `max_tokens` are untouched.
+ */
+function isRejectedMaxTokens(err: unknown): boolean {
+  return (
+    err instanceof OpenAI.APIError &&
+    err.status === 400 &&
+    /max_tokens.*not supported.*max_completion_tokens/i.test(String((err as { message?: string }).message ?? ""))
+  );
+}
+
+/** Same reasoning-family models: temperature is fixed at 1, not caller-settable. */
+function isRejectedTemperature(err: unknown): boolean {
+  return (
+    err instanceof OpenAI.APIError &&
+    err.status === 400 &&
+    /temperature.*does not support/i.test(String((err as { message?: string }).message ?? ""))
+  );
+}
+
+/**
+ * Same models again: function tools are rejected on /v1/chat/completions while
+ * the model's default reasoning effort is active. The API's own error message
+ * names the fix — `reasoning_effort: "none"` — so apply exactly that.
+ *
+ * DeepSeek's v4 line refuses the same combination ("Thinking mode does not
+ * support this tool_choice") for every tool_choice except "auto", and accepts
+ * the identical `reasoning_effort: "none"` remedy. Matching both phrasings here
+ * keeps a forced tool_choice working rather than degrading it to "auto", which
+ * would leave the model free not to call the tool the caller asked for.
+ */
+function isRejectedToolsWithReasoning(err: unknown): boolean {
+  const message = String((err as { message?: string }).message ?? "");
+  return (
+    err instanceof OpenAI.APIError &&
+    err.status === 400 &&
+    (/function tools with reasoning_effort/i.test(message) ||
+      /thinking mode does not support this tool_choice/i.test(message))
+  );
+}
+
 export class OpenAIChatModel implements ChatModel {
   readonly label: string;
   private client: OpenAI;
@@ -141,14 +186,37 @@ export class OpenAIChatModel implements ChatModel {
    * automatically.
    */
   private async createChatCompletion(params: ChatParams): Promise<OpenAI.Chat.Completions.ChatCompletion> {
-    try {
-      return await this.client.chat.completions.create(params);
-    } catch (err) {
-      if (isRejectedObjectToolChoice(err) && typeof params.tool_choice === "object" && params.tools?.length === 1) {
-        return this.client.chat.completions.create({ ...params, tool_choice: "required" });
+    let attempt = params;
+    // Bounded: one retry per known incompatibility below, never an open loop.
+    for (let i = 0; i < 4; i++) {
+      try {
+        return await this.client.chat.completions.create(attempt);
+      } catch (err) {
+        // Checked before the tool_choice fallback below: a reasoning refusal
+        // also names tool_choice, and turning reasoning off keeps the caller's
+        // chosen tool instead of loosening the choice to work around it.
+        if (isRejectedToolsWithReasoning(err) && attempt.reasoning_effort === undefined) {
+          attempt = { ...attempt, reasoning_effort: "none" } as ChatParams;
+          continue;
+        }
+        if (isRejectedObjectToolChoice(err) && typeof attempt.tool_choice === "object" && attempt.tools?.length === 1) {
+          attempt = { ...attempt, tool_choice: "required" };
+          continue;
+        }
+        if (isRejectedMaxTokens(err) && attempt.max_tokens !== undefined) {
+          const { max_tokens, ...rest } = attempt;
+          attempt = { ...rest, max_completion_tokens: max_tokens } as ChatParams;
+          continue;
+        }
+        if (isRejectedTemperature(err) && attempt.temperature !== undefined) {
+          const { temperature, ...rest } = attempt;
+          attempt = rest as ChatParams;
+          continue;
+        }
+        throw err;
       }
-      throw err;
     }
+    return this.client.chat.completions.create(attempt);
   }
 
   private fromResponse(


nguyenduyhung1989:
Heads-up on the red check here (and on #189 / #190): it is the `publish` workflow, not the change.

It runs on `pull_request_target` and checks out `refs/pull/<n>/merge`, which `actions/checkout` now refuses by default for fork PRs — the trusted-context "pwn request" guard:

```
Refusing to check out fork pull request code from a 'pull_request_target' workflow.
```

So it fails before running anything, and will do so for every external contribution, independent of the diff. Flagging it rather than suggesting a setting, since the safe fix depends on what that workflow actually needs the base repo's token for.

Locally on each branch: `tsc --noEmit` clean and the suite green (29/29).

shhdwi:
Verified locally on top of current `main` (clean merge; `npm run build` + `tsc --noEmit` clean).

The change turns the single OpenAI call into a bounded 4-iteration negotiation, each branch gated on the offending param still being present (can't loop), with a trailing call that re-surfaces genuine errors. Verified network-free with a stub SDK throwing each documented 400:
1. `max_tokens` 400 → resend as `max_completion_tokens`.
2. `temperature` 400 → drop `temperature`.
3. tools+reasoning 400 → `reasoning_effort:"none"`, forced `tool_choice` preserved.
4. DeepSeek "Thinking mode does not support this tool_choice" → same remedy; the reasoning check runs BEFORE the object-tool_choice fallback so the choice isn't degraded to `"auto"`.
Compound (all three in sequence) fixed once each; an unknown 400 still surfaces. Existing openai suite 12/12.

Fixes #183. One follow-up worth doing: the PR changes only `openai.ts` and ships no committed test — the negotiation is correct but deserves a checked-in stub test. Merging.

PR #190, opened 2026-08-24
fix(claude): read the hook budget from user-level settings too

nguyenduyhung1989:
The prompt hook sizes its `graft ask` child against the timeout Claude Code will kill it at, read from the repo's `.claude/settings.json`.

But hooks can be declared once at the **user level** instead — one entry in `~/.claude/settings.json` wires every repo on the machine, and none of those repos has a settings file of its own. The lookup finds nothing, falls back to the conservative 8s default, and leaves the child 6s.

In a repo whose graph takes longer than that, the hook then runs, is killed mid-query, and emits nothing: no retrieval pack, no session write, and no error anywhere to say retrieval has stopped. It is indistinguishable from a prompt graft had nothing to say about. One repo here queried in 9.5–12.5s and lost retrieval on **every single prompt** this way — it took timing the hook by hand to notice.

This searches the repo's `settings.json`, its `settings.local.json`, and the user's `settings.json` (honouring `CLAUDE_CONFIG_DIR`).

**Where more than one declares the hook, the smallest budget wins.** Claude Code fires every matching entry and the child cannot tell which one launched it, so only the smallest is safe. The asymmetry justifies it: guessing high gets the whole hook SIGKILLed and loses the turn's retrieval entirely, while guessing low only shortens one query.

One existing test needed a change. `promptAskTimeout` asserting "nothing readable anywhere → 6s" is now also a statement about the machine's home directory — it fails on any developer who has graft's hooks installed at the user level, which is exactly the configuration this fixes. It now pins `CLAUDE_CONFIG_DIR` at an empty directory so the suite stays hermetic.

Verified against the reported repo: the budget goes from 6000ms to 13000ms, and the hook answers in 313ms. `tsc --noEmit` clean, suite green (29/29) including a new test for the fallback and for the smallest-wins rule.

### DIFF
--- a/src/claude/hooks.ts
+++ b/src/claude/hooks.ts
@@ -1,6 +1,7 @@
 import { readFileSync, existsSync } from 'node:fs';
 import { execFileSync, spawn } from 'node:child_process';
 import { join, basename, isAbsolute } from 'node:path';
+import { homedir } from 'node:os';
 import { readWiring } from './stats.js';
 import { formatBlastRadius, relevantRetrieval, formatOrientation } from './format.js';
 import { indexFreshness, staleBanner } from '../context/check.js';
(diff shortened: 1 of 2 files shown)


shhdwi:
Verified locally on top of current `main`. Rebased past a real `src/claude/hooks.ts` conflict with #147 (both edited the region around `installedHookTimeout`) — reapplied this PR's change (add `homedir` import; split into `hookSettingsFiles` + `hookTimeoutIn` + a smallest-wins `installedHookTimeout` that now also reads `~/.claude/settings.json`, honoring `CLAUDE_CONFIG_DIR`) on top of #147's `withContextDirArg`. Both coexist.

Build clean; `claude-hooks` 28/28 including this PR's `promptAskTimeout reads a user-level hook when the repo declares none` and the smallest-wins precedence cases. ("Budget" here = the hook `timeout`, not a token budget.) Merging.

PR #194, opened 2026-08-24
fix(claude): statusline distinguishes an empty graph from a missing one (#185)

### LINKED ISSUE #185: Statusline reports "not built" for a successfully-built but empty graph
### Summary

In a repo whose indexed languages yield no code symbols, `graft build` succeeds and correctly reports an empty graph, but the Claude Code statusline then permanently renders:

```
◤ graft · not built · run graft build
```

It tells the user to run a command they have already run successfully, and keeps saying it after every rebuild, so it reads as a broken install.

### Cause

`renderStatusline()` treats `nodeCount === 0` as "never built" (`dist/claude/format.js:24`), and `resolveStats()` falls back to `readWiring()`, which is also empty — so a legitimately empty graph is indistinguishable from a missing one.

### Reproduce

```bash
mkdir empty-repo && cd empty-repo && git init
echo "# docs only" > README.md
graft build      # succeeds: "✓ wiring: 0 nodes (), 0 edges, 0 cards []"
```

Then open the repo in Claude Code and look at the statusline.

### Why a repo can legitimately have zero nodes

Our case is a docs/markdown repo whose only code lives under `.claude/` (a workflow `.js` and two helper `.cjs`). Graft's walker skips dot-directories, and unlike ordinary skipped dirs this is not overridable:

```
$ graft build --include-dir .claude
✗ --include-dir ".claude": dot-directories are never overridable
```

That's a defensible rule, so the 0-node graph is correct — only the statusline wording is wrong.

### Suggested fix

Distinguish "no graph" from "graph with no symbols". The build already knows it ran, so persist that (a `builtAt` in `graft/.cache/stats.json`, or the existence of the build artifact) and render the empty-but-built case as something accurate, e.g. `◤ graft · 0 nodes · no indexed code`, reserving "not built" for when no artifact exists at all.

`graft ask` and `graft grep` behave correctly throughout; this is purely the status display.

Frankie-Xu:
## Summary

- Statusline "not built" is now "no graph artifact", not "zero nodes".
- A successful `graft build` that produces an empty wiring.json (docs-only repos, skipped languages) renders as `◤ graft · 0 nodes / 0 edges · ✓ synced`.
- Repos with no `wiring.json` still render `◤ graft · not built · run graft build`.
- Non-empty graphs are unchanged.

## Cause

As analyzed in #185: `renderStatusline()` treated `nodeCount === 0` as never-built, and `resolveStats()` could not tell a legitimate empty graph from a missing one — both look like zero. `graft build` already succeeds (`✓ wiring: 0 nodes ...`); only the bar was lying.

Built is now artifact presence (`wiring.json` via `readWiring()`). `renderStatusline` only shows "not built" when stats are `null`. A 0-node cache without wiring is still missing, not empty-and-built.

## Test plan

- [x] `node --import tsx --test test/claude-format.test.ts test/claude-statusline.test.ts test/claude-stats.test.ts` (31/31)
- [x] `npm run build && npm test` (882/882)
- [x] Manual docs-only repo (`README.md` only, `git init` + commit, `graft build`):

```
# graft build
✓ wiring: 0 nodes (), 0 edges, 0 cards []

# statusline after that successful empty build
◤ graft · 0 nodes / 0 edges · ✓ synced

# control: same layout, never built (no wiring.json)
◤ graft · not built · run graft build

# control: this repo (non-empty graph, same bar style as before)
◤ graft · 1524 nodes / 4547 edges · ✓ synced
```

Closes #185

### DIFF
--- a/src/claude/format.ts
+++ b/src/claude/format.ts
@@ -26,7 +26,7 @@ export function renderStatusline(
   session: SessionState | null,
   ctx: { ctxPct: number | null },
 ): string[] {
-  if (!stats || stats.nodeCount === 0) {
+  if (!stats) {
     return [C.muted('◤ graft · not built · run ') + C.text('graft build')];
   }
   const top = [C.muted('◤ ') + C.indigo('graft'), C.text(`${stats.nodeCount} nodes / ${stats.edgeCount} edges`)];
--- a/src/claude/statusline.ts
+++ b/src/claude/statusline.ts
@@ -7,16 +7,18 @@ import { readWiring, computeStats } from './stats.js';
  * The statusline's fast path is the hook-maintained cache (graft/.cache/stats.json).
  * When it's absent — a fresh checkout, or a plain `graft build` that doesn't write the
  * cache — fall back to reading the graph itself (wiring.json) so the bar reflects reality
- * immediately instead of showing "not built". The graph carries no drift signal, so it
- * reads as synced until the next edit repopulates the cache. Still a pure read (no
- * subprocess): the cache is preferred because it carries live dirty/stale state.
+ * immediately instead of showing "not built". An empty wiring.json is still a graph
+ * (docs-only repos legitimately have 0 nodes); "not built" is only when the artifact
+ * is missing. The graph carries no drift signal, so it reads as synced until the next
+ * edit repopulates the cache. Still a pure read (no subprocess): the cache is preferred
+ * because it carries live dirty/stale state.
  */
 export function resolveStats(dir: string): Stats | null {
   const cached = readStats(dir);
   if (cached && cached.nodeCount > 0) return cached;
   const wiring = readWiring(dir);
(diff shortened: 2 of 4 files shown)


shhdwi:

Before/after (statusline line, ANSI-stripped):
- **Missing graph** (no `wiring.json`): `◤ graft · not built · run graft build` — unchanged, still correct.
- **Empty-but-built graph** (docs-only repo, `wiring.json` with 0 nodes): OLD `◤ graft · not built · run graft build` (the #185 bug) → NEW `◤ graft · 0 nodes / 0 edges · ✓ synced`.

Fix is `format.ts` (`nodeCount === 0` no longer forced into "not built") + `resolveStats()` returning `null` only when the artifact is genuinely absent. 3 new tests for the empty-graph case; `claude-*` 105/105. Fixes #185. Merging.

PR #196, opened 2026-08-24
feat(graph): index HTML files so templates are findable by name (#150)

### LINKED ISSUE #150: Support HTML files
**Version**: `@nanonets/graft@0.10.1` via `npm install -g`

I am working with Django and there is currently no support for template files (.html).
It would be very handy to have html indexed at least by name, so that the search returns the file names directly (instead of checking the python file -> View class -> template name)

Frankie-Xu:
## Summary

- Register HTML (`.html` / `.htm`) on the breadth tier so template files get a `kind: "file"` graph node and show up in `ask` / `grep` by name — Django users no longer have to walk View class → `template_name` to find the template (#150).
- `tree-sitter-wasm` already ships `tree-sitter-html.wasm`, so this is one `GENERIC_LANGS` row (same path as Clojure/Dart). No `tags.scm`: HTML has no definition-shaped nodes, and the node-kind walker correctly yields a file node only (ocaml/zig precedent).
- Grammar load stays lazy — a repo with no HTML files does not warm the HTML wasm.

Closes #150

## Route tradeoff

| Option | Why / why not |
|---|---|
| **GENERIC_LANGS + html wasm** (this PR) | Grammar is already in the bundle. File collection, lazy warmup, `graft build -e`, and banner labeling all fall out of the existing registry. |
| File-only ingest in `fs.ts` / `build.ts` | Would work without a grammar, but would duplicate "is this a source file?" outside the three-tier extension tables. Unnecessary once the wasm is there. |
| Deep HTML parse (ids/classes as `@definition`) | Out of scope. Useful later; not needed to find a template by filename. |
| Python `render` / `template_name` → HTML file edges | Cross-language, larger. Follow-up below. |

## Test plan

- [x] `npm run build && npm test` — 868 pass, 0 fail
- [x] `test/generic-extract.test.ts` — `.html`/`.htm` routing; extract is file-node-only; Django-style `templates/index.html` + `views.py` build: grep hits template content, `resolveSymbol("index.html")` and `ask("index.html")` hit the file node
- [x] `test/supported-extensions.test.ts` — `.html` / `.htm` in the `-e` supported set
- [x] Manual fixture (`templates/index.html` + Django `TemplateView`):

```
✓ wiring: 3 nodes (2 file, 1 class), … [html, python]

graft ask — "index.html"
1. index.html · file  [symbol]
   templates/index.html
2. Home · class  [symbol]
   views.py:L3-L4

graft grep welcome-to-the-shop
templates/index.html (module level)
  L6: {% block content %}welcome-to-the-shop{% endblock %}
```

## Follow-up

- Cross-language edges: resolve Python `render("…")` / `template_name = "…"` string literals onto the matching HTML file node. That is the "go to the template from the View" direction; this PR is the "find the template by name" direction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ compiler-grade layer — all `$0` and deterministic (no model, no key):
 - **Broad** — symbols (functions, classes, methods, types, …) plus name-resolved
   call edges via a generic tree-sitter extractor, one grammar per language:
   **Rust, C, C++, C#, Ruby, Scala, Elixir, Solidity,
-  OCaml, Zig, Dart, Clojure, Nix, Lua**.
+  OCaml, Zig, Dart, Clojure, Nix, Lua, HTML**.
 
 - **Compiler-grade edges (opt-in)** — `graft build --lsp` adds precise
   `lsp_resolved` call edges (member calls the static pass can't type) when a
@@ -219,8 +219,9 @@ compiler-grade layer — all `$0` and deterministic (no model, no key):
   **gopls** (Go), **pyright** (Python), **typescript-language-server** (TS/JS).
   It's best-effort — with no server installed the graph is unchanged.
 
-Twenty-three languages in total. A file whose language isn't listed is skipped, not
-indexed. Adding a broad-tier language is a small contribution — see
+Twenty-four languages in total. HTML is indexed as a file node (no HTML-structure
+extraction), so templates are findable by name. A file whose language isn't listed
+is skipped, not indexed. Adding a broad-tier language is a small contribution — see
 [CREDITS.md](CREDITS.md) for the folks who added the current set.
 
 ---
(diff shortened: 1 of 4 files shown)


Frankie-Xu:
Rebased onto current `main` (`ee1ef03`). Conflicts kept Nix/Lua (and Kotlin on the depth tier) and added HTML as a file-node-only breadth row so templates stay findable by name.

`npm run build` OK. `npm test` **918/918** (4 load flakes — PageRank timing, MCP stdio — passed on isolated re-run).

Frankie-Xu:
Rebased onto current main. Conflict resolution kept the current language list (PHP/Kotlin/Swift stay full-fidelity; Nix stays in the broad list) and added HTML as a file-node-only breadth language. Waiting on CI before marking ready.

Frankie-Xu:

PR #199, opened 2026-08-24
feat(init): never clobber a user-defined statusline (#191)

### LINKED ISSUE #191: my custom Statusline was overrided
Please add and option to not ovverride statusline or if possibile to add it in a new "line"

Frankie-Xu:
## Summary

- `graft init` no longer hides a custom Claude Code statusline. A project-level `statusLine` that is not Graft's is left untouched (this was already true); **`--no-statusline`** / **`GRAFT_NO_STATUSLINE=1`** skip installing one at all, so a bar in `~/.claude/settings.json` stays visible.
- Graft recognises its own bar by the helper path `graft-statusline.cjs` (not a full-command string match) and will refresh that command on re-init. The choice is recorded in the wiring stamp, so a later session refresh cannot put Graft's bar back.

Closes #191

### What was actually happening

Merge already kept a *project-level* foreign `statusLine`. The override people hit is:

1. Custom bar lives in **user** settings (`~/.claude/settings.json`).
2. First `graft init` writes a **project** `statusLine` because `.claude/settings.json` has none.
3. Claude Code prefers the project field, so the user-level bar disappears.
4. Deleting the project field and restarting is not enough: upkeep re-runs init and writes it again.

`--no-statusline` is the opt-out (same shape as `GRAFT_NO_GITIGNORE` / `--no-global`): default install is unchanged for anyone who does not already have a custom bar.

### Why not "append as a new line"

Claude Code has **one** `statusLine` command per session (the merge warning already said this). Graft's helper already prints multiple lines from that one command (`renderStatusline(…).join('\n')`). Combining two independent commands would need a wrapper that tees stdin JSON (consumed once) and merges timeouts — too fragile to do by default. To show both, point `statusLine` at a small wrapper that runs your command then `.claude/helpers/graft-statusline.cjs`.

## Behavior matrix

| Existing `statusLine` | First `graft init` | Re-init / wiring refresh |
|---|---|---|
| None | Writes Graft's helper (unchanged default) | Keeps / refreshes Graft's helper |
| Graft's (`graft-statusline.cjs` in the command, even an old path) | — | Updates to the current command |
| Foreign (any other command) | **Kept**, with a warning | **Kept**, with a warning |
| None, with `--no-statusline` / `GRAFT_NO_STATUSLINE=1` | **Not written** | Stamp replays the skip; a prior Graft field is stripped so a user-level bar can show |
| Foreign, with `--no-statusline` | **Kept** | **Kept** |

Hooks, skill, MCP, and the statusline **shim file** are still installed. Only the `statusLine` / `subagentStatusLine` fields are skipped.

## Test plan

- [x] `npm run build && npm test` — 891 passing
- [x] `node --import tsx --test test/claude-settings-merge.test.ts test/claude-init.test.ts test/upkeep.test.ts`
  - custom `statusLine` preserved
  - old Graft helper command updated
  - empty settings still get Graft's bar
  - `--no-statusline` / `GRAFT_NO_STATUSLINE=1` skip install
  - stamp replay of `--no-statusline`
- [ ] Manual: `graft init --no-agents --no-statusline` in a repo whose `~/.claude/settings.json` has a custom bar; project settings have no `statusLine`; Claude Code still shows the custom bar after restart

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -43,6 +43,12 @@
 # GRAFT_NO_GITIGNORE=1  — skip writing graft/ into .gitignore (e.g. you already ignore it globally)
 # GRAFT_NO_IGNORE=1     — skip writing .ignore (ripgrep re-admit); only if you manage search yourself
 
+# --- Init (Claude Code) -------------------------------------------------------
+# Skip writing statusLine / subagentStatusLine during `graft init` (same as
+# `--no-statusline`). Use this when you already have a custom Claude Code
+# statusline — a project-level field would hide one in ~/.claude/settings.json.
+# GRAFT_NO_STATUSLINE=1
+
 # --- Benchmark (bench/) -------------------------------------------------------
 # Per-role model overrides for the benchmark harness.
 # BENCH_AGENT_MODEL=anthropic/claude-sonnet-5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- **`graft init --no-statusline`** (and `GRAFT_NO_STATUSLINE=1`) skips writing
+  Claude Code's `statusLine` / `subagentStatusLine`. A custom bar — in the
+  project's `.claude/settings.json` or in `~/.claude/settings.json` — stays in
+  front: a project-level field would otherwise hide the user-level one. The
+  choice is recorded in the wiring stamp, so a later session refresh cannot
+  put Graft's bar back. Graft still recognises its own helper
+  (`graft-statusline.cjs`) and will update that command on re-init.
+
 ## 0.15.0
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -271,6 +271,7 @@ With no TTY to prompt on — CI, a Dockerfile, a piped shell — `init` writes *
 | `--list-agents` | print the known agent ids and exit |
 | `--no-mcp` | skip MCP server registration |
 | `--no-hooks` | skip hook installation |
+| `--no-statusline` | skip writing Claude Code `statusLine` (same as `GRAFT_NO_STATUSLINE=1`) |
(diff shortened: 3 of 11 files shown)


Frankie-Xu:
Rebased onto current `main` (`ee1ef03`). CHANGELOG conflict: kept the 0.13.0 notes and put `--no-statusline` under Unreleased.

`npm run build` OK. `npm test` **924/924**.

Frankie-Xu:
Converted back from Draft now that the review queue is smaller.

Rebased onto current `main` (`268e30d`). Conflicts were CHANGELOG (kept 0.15.0 notes; `--no-statusline` under Unreleased), `.env.example` (kept both opt-out blocks), and `settings-merge.ts` (kept main's allowlist/footer helpers; this PR's statusline merge: leave a user-defined `statusLine` alone, refresh Graft's own helper).

Fixes #191: `graft init` does not overwrite a user-defined Claude Code statusline. `--no-statusline` / `GRAFT_NO_STATUSLINE=1` skip installing one so a user-level bar is not hidden.

Same Claude statusline area as merged #194 (empty-graph statusline copy), complementary rather than a substitute.

`npm run build` OK. `claude-settings-merge` + `claude-init` + `upkeep` tests green.

shhdwi:
Verified locally on top of current `main` (clean merge; build clean). Reproduced #191: a pre-existing `{"statusLine":"~/my-custom-bar.sh"}` is now **preserved** (with a "left untouched" warning) while graft still wires hooks, skill, MCP, helpers, and subagentStatusLine. `--no-statusline` / `GRAFT_NO_STATUSLINE` leaves the user's statusline alone but still wires hooks, and the opt-out is replayed on the upkeep refresh so a later session can't silently re-install the bar. Also a real improvement over warn-only: graft recognizes its own statusline shim by helper filename, so a stale graft bar is refreshed instead of mis-warned. 52/52 init/settings/upkeep tests. Fixes #191. Merging — thanks!

PR #200, opened 2026-08-24
feat(graph): add OCaml tags.scm — call edges for the breadth tier

### LINKED ISSUE #197: ocaml/zig: no tags.scm — symbols come from the node-kind walker fallback, call edges are absent
## Status

OCaml (`.ml`/`.mli`) and Zig (`.zig`) are already registered in the breadth-tier `GENERIC_LANGS` table, but they have no vendored `src/graph/queries/<lang>.scm`. `generic.ts` says so explicitly:

> These ship a tags.scm (calls + symbols); ocaml/zig have none and use the node-kind walker fallback (symbols only) — still one row, zero query.

The walker mints symbols from node-kind suffixes. It does not emit `@reference.call`, so `graft callers` on these languages has no edges to walk — same-file calls never resolve.

This is the same hole Dart had before #153 / #134: language is on the registry, grammar warms, but without a tags query the graph is symbols-only.

## Plan

Two independent PRs, each adding one language the way #153 added Dart (query + `generic-extract` fixture; drop-not-guess — only AST shapes confirmed against the WASM grammar):

1. **OCaml** — `ocaml.scm`: `@definition.function` (let / let rec with a parameter, or `fun` body), `@definition.module`, `@definition.type`, `@reference.call` on `application_expression`. No infix/`|>`/`@@` (noisy; not a call node).
2. **Zig** — `zig.scm`: `function_declaration`, `const`+`struct_declaration`, named `test_declaration`, `call_expression`.

Each PR: one `.scm` + tests + the `generic.ts` comment line for that language. No registry-row changes (both langs are already registered).

I'll follow up with the PRs.

Frankie-Xu:
## Summary
- OCaml is already on the breadth-tier registry, but had no `tags.scm`, so the node-kind walker minted `let helper x = …` as kind `variable` (`value_definition` matches `(^|_)(val|…)`) and emitted no call edges.
- Add an OCaml tags query for `let` / `let rec` functions (parameter or `fun`/`function` body), modules, types, and `application_expression` calls. Bare `let n = 1` values and infix/`|>`/`@@` are left untagged (drop-not-guess). Node types confirmed against the `tree-sitter-wasm` ocaml grammar AST.
- Same pattern as Dart in #153. Zig is unchanged (still walker fallback). Independent of #201.

Fixes #197.

## Test plan
- [x] `node --import tsx --test test/generic-extract.test.ts` — OCaml fixture: `go → helper`, `M.wrap → go`; `let n = 1` is not a symbol
- [x] `npm run build`
- [x] `npm test` (884/884)

### DIFF
--- a/src/graph/generic.ts
+++ b/src/graph/generic.ts
@@ -47,7 +47,7 @@ export const GENERIC_LANGS: readonly GenericLang[] = [
   { name: "cpp", exts: [".cpp", ".cc", ".cxx", ".hpp", ".hh"], wasm: "cpp" },
   { name: "ruby", exts: [".rb"], wasm: "ruby" },
   { name: "c_sharp", exts: [".cs"], wasm: "c_sharp" },
-  // These ship a tags.scm (calls + symbols); ocaml/zig have none and use the
+  // These ship a tags.scm (calls + symbols); zig has none and uses the
   // node-kind walker fallback (symbols only) — still one row, zero query.
   { name: "scala", exts: [".scala", ".sc"], wasm: "scala" },
   { name: "elixir", exts: [".ex", ".exs"], wasm: "elixir" },
--- a/src/graph/queries/ocaml.scm
+++ b/src/graph/queries/ocaml.scm
@@ -1,0 +1,38 @@
+; OCaml — breadth-tier tags.scm
+;
+; Confirmed against the tree-sitter-wasm `ocaml` grammar AST (not ocaml_interface).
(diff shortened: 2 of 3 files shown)


Frankie-Xu:
Rebased onto current main. Kept the existing PHP wasm / throwing-grammar tests and layered the OCaml tags.scm fixture on top. Waiting on CI before marking ready.

Frankie-Xu:
Rebased onto current main. CI is green. Ready for review.

PR #203, opened 2026-08-24
Reuse PageRank topology across scopes

### LINKED ISSUE #202: ask: multi-scope PageRank rebuilds the full graph topology once per scope
## Problem

For a multi-scope repository, `ask` runs personalized PageRank separately for each participating scope. Each call currently applies a `nodeFilter` by rebuilding the real-node set and scanning every graph edge to reconstruct the filtered adjacency.

With `S` matching scopes, topology preparation therefore costs approximately:

```text
O(S × (nodes + edges))
```

even though the scopes are disjoint partitions of the same immutable graph. Large monorepos can turn a single query into many repeated full-graph scans before scope fusion completes.

## User-visible effect

Preparing all disjoint scope topologies in one pass reduces CPU work and query latency in multi-scope repositories. The benefit grows with the number of matching scopes because the graph is scanned once instead of once per scope.

It does not change retrieval order or PageRank values. It also deliberately has no effect on single-scope repositories, `graphRank: false`, or no-match queries.

## Synthetic graph-ranking benchmark

Local Node.js v22.22.0 arm64 microbenchmark with 12,000 nodes, 36,000 intra-scope edges, the normal 25 PageRank iterations, two warm-ups, and the median of seven measured runs:

| matching scopes | current filtered walks | prepared partitions | speedup |
|---:|---:|---:|---:|
| 5 | 9.87 ms | 6.27 ms | 1.6× |
| 20 | 28.44 ms | 11.98 ms | 2.4× |
| 50 | 72.96 ms | 21.43 ms | 3.4× |

This isolates the graph-ranking stage rather than claiming an end-to-end `graft ask` latency improvement. Real impact depends on graph shape, the number of matching scopes, and the share of query time spent outside PageRank.

## Expected behavior

Partition the graph topology once, then reuse each prepared partition for its scope-specific seed walk:

```text
O(nodes + edges) topology preparation
+ per-scope PageRank iteration
```

The optimization should:

- preserve exact PageRank scores, edge multiplicity, and deterministic order
- exclude cross-scope and unresolved edges exactly as independent filtered walks do
- remain lazy so `graphRank: false` and no-match queries do not scan graph edges
- leave single-scope behavior unchanged

## Reproduction / regression guard

A graph with two independent scopes can expose counters on its `nodes` and `edges` arrays. Preparing both partitions should read each array once, and running both prepared walks should produce byte-equivalent entries to two independent `nodeFilter` walks.

I have a focused implementation and regression coverage ready.

Skyline-23:
## Context

In a multi-scope repository, `ask` currently calls personalized PageRank once per participating scope. Every filtered call rebuilds the node set and scans all graph edges, so topology preparation grows with the number of matching scopes:

```text
before: O(scopes × (nodes + edges))
after:  O(nodes + edges) + per-scope PageRank iteration
```

Fixes #202.

## Effect

The change reduces CPU work and query latency in multi-scope repositories, with a larger benefit as more scopes participate. Retrieval order and PageRank values remain unchanged.

A local Node.js v22.22.0 arm64 synthetic microbenchmark used 12,000 nodes, 36,000 intra-scope edges, the normal 25 PageRank iterations, two warm-ups, and the median of seven measured runs:

| matching scopes | current filtered walks | this PR | speedup |
|---:|---:|---:|---:|
| 5 | 9.87 ms | 6.27 ms | 1.6× |
| 20 | 28.44 ms | 11.98 ms | 2.4× |
| 50 | 72.96 ms | 21.43 ms | 3.4× |

This isolates the graph-ranking stage rather than claiming an end-to-end `graft ask` speedup. Real impact depends on graph shape, matching-scope count, and time spent in the rest of the retrieval pipeline. Single-scope, `graphRank: false`, and no-match queries do not benefit and remain on their existing paths.

## Changes

- Separate immutable topology preparation from PageRank iteration.
- Add one-pass preparation for disjoint graph partitions.
- Lazily prepare scope topologies on the first ranked multi-scope walk.
- Reuse each prepared topology with the scope-specific lexical seeds.
- Keep the existing `personalizedPageRank` API by delegating through the prepared path.

## Behavioral invariants

- Prepared walks are entry-for-entry identical to independent `nodeFilter` walks.
- Edge multiplicity and adjacency order are preserved.
- Cross-scope and unresolved edges remain excluded.
- `graphRank: false` and no-match queries do not trigger topology preparation.
- The single-scope path is unchanged.

## Validation

- Rebased onto current `main` at `d92f856`.
- `npm ci` / package prepare build: passed
- Focused `ask` + PageRank tests: **48/48 passed**
- Full local suite: **940/940 passed**
- `git diff --check`: passed

Final diff: 3 files, +186/-32.

### DIFF
--- a/src/ask/ask.ts
+++ b/src/ask/ask.ts
@@ -32,7 +32,11 @@ import { resolveSymbol } from "../graph/traverse.js";
 import type { EdgeV1, GraphV1, NodeV1, Relation } from "../graph/types.js";
 import { rankScopesAndFuse } from "./fuse.js";
 import { fileFirstRoundRobin } from "./file-selection.js";
-import { personalizedPageRank } from "./graphrank.js";
+import {
+  personalizedPageRank,
+  personalizedPageRankPrepared,
+  preparePageRankPartitions,
+} from "./graphrank.js";
 import { readSourceFile } from "../util/source.js";
 import { counts, tokenize, type AskIndex, type AskIndexDoc } from "./index-file.js";
 
@@ -567,12 +571,20 @@ function lexical(
     // can be ordered by score. Same scoring functions and blend as the
     // single-scope path; the combination itself lives in fuse.ts. ──
     const byScope = new Map<string, typeof symbolDocs>();
+    const scopeById = new Map<string, string>();
     for (const d of symbolDocs) {
       const s = scopeOf(d.n.path, scopes).prefix;
+      scopeById.set(d.n.id, s);
       const list = byScope.get(s);
       if (list) list.push(d);
       else byScope.set(s, [d]);
     }
+    // Lazily build every scope's PageRank topology in one node/edge pass. The
+    // old nodeFilter path rebuilt the full repo adjacency once per matching
+    // scope; on large multi-project repos that turned one query into hundreds
+    // of full graph scans before the participation gate could run. Waiting for
+    // the first walk keeps no-match and graphRank:false queries edge-scan free.
+    let pageRankByScope: ReturnType<typeof preparePageRankPartitions> | undefined;
     const fusion = rankScopesAndFuse(
       [...byScope.keys()].sort(),
       {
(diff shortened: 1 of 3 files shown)


shhdwi:
Verified locally on top of current `main`. Rebased the `ask.ts` conflict with #205 (kept both the file-rank imports/state and this PR's prepared-PageRank imports/state — they're independent).

Fixes #202: `preparePageRankPartitions` builds every scope's PageRank topology in ONE node/edge pass instead of rebuilding the full repo adjacency once per scope. Correctness is unchanged — a 2-scope `ask --graph-rank` run is byte-identical (same order, scores to 6dp) before/after — and a guard test asserts `nodeReads==1`/`edgeReads==1` plus float/order equivalence vs independent per-scope walks. Composes cleanly with #205 (file-rank) and #137/#126: ask suites 112/112, full suite 1042/1042. Fixes #202. Merging.

PR #207, opened 2026-08-24
feat(java): wire annotation usage as references edges (#89)

### LINKED ISSUE #89: Java annotations: the text is already retrievable — the gap is a traversable edge (scope question before a PR)
Following up on #83, where the annotation-extraction gap was offered as a separate PR. Before writing it I went to measure what's actually missing, and the answer surprised me enough that it seems worth agreeing on scope first rather than sending a PR built on a wrong premise.

Everything below reproduces on two public repos: [`spring-projects/spring-petclinic`](https://github.com/spring-projects/spring-petclinic) (JPA + Spring MVC annotations) and [`google/gson`](https://github.com/google/gson) (almost none), both at `--depth 1`.

## Annotation text already reaches the graph

Class and method annotations land in `signature` today, unchanged:

```
@Entity      in signature:  6   @Entity @Table(name = "owners") public class Owner extends Person
@Table       in signature:  6
@GetMapping  in signature: 12   @GetMapping("/owners/new") public String initCreationForm()
@PostMapping in signature:  5
@Column      in signature:  0
@JoinColumn  in signature:  0
```

Field annotations are absent from `signature` because Java field declarations aren't nodes — but they are not unreachable either:

```
$ graft grep "@JoinColumn"
"@JoinColumn" — 5 hits in 3 symbols across 3 files (searched 49 indexed files)

Pet · class · src/main/java/org/springframework/samples/petclinic/owner/Pet.java:L44-L85 · 9 in-edges
  L53: @JoinColumn(name = "type_id")
  L57: @JoinColumn(name = "pet_id")
```

Every hit, with exact lines, attributed to the enclosing symbol. So "extract annotations" as a text-retrieval feature is already done — an agent asking *which entity maps to which table* or *what column is this field* can get an exact answer today.

## The real gap is traversable structure

What's missing is an **edge**. You can grep `@JoinColumn(name = "type_id")` on `Pet`, but there is no `Pet → PetType` relation, so `callers`, impact walks and graphrank can't follow the association. The information is retrievable but not *navigable*.

## Why field nodes look like the fix and aren't

The obvious approach — emit a node per field declaration, carrying its annotations in `signature` — collides with the Tier-0 gate from #87:

```js
for (const e of edges) { if (e.relation !== "contains") { withEdge.add(e.source); ... } }
```

`contains` is excluded from the orphan calculation, deliberately and correctly. A field node's only structural edge *is* its `contains` from the declaring type, so every field would count as an orphan. And a declared-type edge doesn't rescue them, because the fields that carry `@Column` are overwhelmingly `String` and primitives.

On gson that would be 367 new orphan symbol nodes against 4,446 today. That trades the 0%-orphan property for annotations that are already greppable — a bad deal in both directions.

## Two shapes that avoid it, and what each is worth

Neither adds a node, a `Relation`, or a schema change. Both reuse `references`, which is already in `WALK_RELATIONS`, so `callers`/impact/graphrank follow them for free.

**A — a field's declared type becomes a `references` edge when it resolves in-repo.** Framework-neutral; JPA associations fall out as a special case, alongside constructor-injected collaborators.

**B — only JPA relationship annotations (`@ManyToOne`, `@OneToMany`, `@OneToOne`, `@ManyToMany`) produce the edge.** Literally "annotation extraction", but framework-coupled.

Edge yield, counting fields whose declared type resolves to a type defined in the same repo:

| repo | field declarations | in-repo type (A) | of those, annotated (B) |
|---|---|---|---|
| spring-petclinic | 24 | **7** | 1 |
| gson | 367 | **170** | 2 |

That last column is the argument against B: on a repo with no JPA it contributes two edges, and it puts framework knowledge inside a general extractor. A is strictly more general and captures B's cases.

## What I'd propose, and a question

My inclination is A, and to say plainly in the README that annotation *text* is served by `grep`/`signature` rather than by a dedicated extractor — that's true today and worth writing down, since it's not obvious.

But this is your call on scope, and there are two honest counterarguments I don't want to paper over:

1. A is not what "annotation extraction" means to most readers. If the intent was that `@Entity`/`@Column` become first-class queryable metadata, A doesn't deliver it — and doing *that* properly probably does need a node or field shape, which means engaging with the orphan metric rather than dodging it.
2. A's value on a repo like gson (170 edges) is mostly the composition graph, not annotations at all. It might belong under a different title.

Happy to implement whichever you prefer, with the same treatment as #83 — tests that are verified to fail when reverted, and a same-repo before/after measurement on public repos. Or to leave it if the honest answer is that the existing `grep` path is enough.

Frankie-Xu:
## Summary

- Extract Java `marker_annotation` / `annotation` on already-noded definitions (class / interface / enum / record / method / constructor) and emit `references` edges from the annotated symbol to the annotation type.
- Resolve in-repo `@interface` hits via `resolveName` (same-file, then globally unique `interface` kind). Unresolved targets keep the bare name — heritage's contract, not PHP #155's drop.
- Annotation types that are classes of the same name are not a match, so `@Entity` cannot collapse onto an in-repo `class Entity` (#103).
- Arguments are ignored: `@MyAnno(value = "x")` emits the same edge as marker `@MyAnno`. Field annotations are not edges (fields are not nodes).

Mirrors https://github.com/NanoNets/Graft/pull/155 (PHP attributes → `references`). Does not touch the #176 anonymous-class path.

## Scope

Slice **C** as aligned on https://github.com/NanoNets/Graft/issues/89 (Frankie-Xu 2026-08-21, endorsed by @dbianco: *"Your C is a better slice than my A… The #155 precedent settles the design questions… Nothing left to adjudicate."*):

| Do | Don't |
|---|---|
| class / method (and other already-noded) annotations → `references` | parse annotation arguments |
| source = annotated symbol, target = annotation type | mint nodes for external library types |
| keep the bare name when unresolved (heritage) | field nodes / field-annotation edges |
| relation already in `WALK_RELATIONS` | framework-special-case JPA (option B) |

## Yield

Measured by @dbianco on current `main` ([issue comment](https://github.com/NanoNets/Graft/issues/89#issuecomment-3728500165)):

| | annotation usages | distinct types | in-repo `@interface` |
|---|---:|---:|---:|
| spring-petclinic | 268 | 63 | **0** |
| gson | 3,184 | 34 | 8 |

Usages that point at a type the repo actually declares (gson):

```
@JsonAdapter      93
@SerializedName   45
@Expose           19
@Since            11
@Until             9
@Intercept         3
@Foo               1
@IgnoreJRERequirement 1
                 ---
                 182  of 3,184 usages  (~6%)
```

Petclinic is **0 of 268** — every annotation there is Spring / JPA / Jackson, so every edge targets an unresolved name. That is still a navigable `Owner → Entity` string, same as heritage.

Honest headline: a name-level annotation graph, mostly to external types; gson is where the resolved slice is visible. Those 182 counts include field usages; this PR only wires class/method-level ones (`@JsonAdapter` on a type, a method `@SerializedName`, …). Field annotations stay in `signature` / `graft grep`.

## Test plan

- [x] `node --import tsx --test test/graph-java.test.ts` — 35/35, including:
  - in-repo `@interface MyAnno` on class + method → `references` to the annotation node
  - `@MyAnno(value = "x")` same edges as marker `@MyAnno`
  - external `@Override` / `@Entity` keep the bare name
  - `@Entity` does not false-match an in-repo `class Entity`
  - field `@Transient` is not a class-level edge
- [x] `npm run build` clean
- [x] `npm test` — 913 pass; 5 failures were load flakes (`PageRank` 3s budget, MCP stdio timeout, `viz --tabs` against a worktree with no `graft/`). Isolated re-run of those files + `graph-php` / `graph-references` / `graph-invariants` is green except the pre-existing `viz --tabs` cwd-graph assumption.
- [x] Gson-style fixture (in-repo `@JsonAdapter` / `@SerializedName` / `@Expose`, class + method + field usages):

```
Person          -> …/JsonAdapter.java#JsonAdapter
Person.id       -> …/SerializedName.java#SerializedName
```

Field `@SerializedName` / `@Expose` produced **0** extra class-level edges.

Closes #89

### DIFF
--- a/scripts/graph-quality.mjs
+++ b/scripts/graph-quality.mjs
@@ -48,8 +48,8 @@ for (const n of nodes) {
   else if (Number(m[1]) > Number(m[2])) problems.push(`inverted span ${n.span}: ${n.id}`);
 }
 // an edge target may be a node id OR a deliberately-unresolved external string
-// (import specifier / bare heritage name); only calls/references/contains must
-// point at real nodes.
+// (import specifier / bare heritage name / unresolved Java annotation type);
+// only calls/contains must point at real nodes.
 let dangling = 0, selfLoops = 0, unresolvedExternal = 0;
 for (const e of edges) {
   if (!RELATIONS.has(e.relation)) problems.push(`bad relation '${e.relation}'`);
@@ -57,7 +57,7 @@ for (const e of edges) {
   if (!ids.has(e.source)) { problems.push(`dangling source: ${e.source}`); dangling++; }
   const targetIsNode = ids.has(e.target);
   if (!targetIsNode) {
-    if (e.relation === "imports" || e.relation === "extends" || e.relation === "implements") unresolvedExternal++;
+    if (e.relation === "imports" || e.relation === "extends" || e.relation === "implements" || e.relation === "references") unresolvedExternal++;
     else { problems.push(`dangling ${e.relation} target: ${e.source} → ${e.target}`); dangling++; }
   }
   if (e.relation === "calls" && e.source === e.target) selfLoops++;
--- a/src/graph/extract.ts
+++ b/src/graph/extract.ts
@@ -583,6 +583,7 @@ function walk(node: Parser.SyntaxNode, ctx: WalkCtx, out: NodeV1[], edges: RawEd
     if (desc.kind === "class" || javaTypeDecl || kotlinTypeDecl)
       edges.push(...heritageEdges(node, id, ctx));
     if (ctx.lang === "php") edges.push(...phpAttributeReferenceEdges(node, id, ctx));
+    if (ctx.lang === "java") edges.push(...javaAnnotationReferenceEdges(node, id, ctx));
 
     const enclosingClass =
       desc.kind === "class" || javaTypeDecl || kotlinTypeDecl
@@ -894,6 +895,36 @@ function phpAttributeClassRef(
   const imported = ctx.importedSymbols.get(bare);
   if (imported) return { name: imported.name, specifier: imported.specifier };
   return { name: bare };
+}
+
+/** Java annotations on a definition → `references` edges to the annotation type. */
+function javaAnnotationReferenceEdges(node: Parser.SyntaxNode, sourceId: string, ctx: WalkCtx): RawEdge[] {
+  const edges: RawEdge[] = [];
+  const mods = node.namedChildren.find((c) => c.type === "modifiers");
+  if (!mods) return edges;
+  for (const child of mods.namedChildren) {
+    if (child.type !== "marker_annotation" && child.type !== "annotation") continue;
+    const name = javaAnnotationTypeName(child);
+    if (name) {
+      edges.push({
+        source: sourceId,
+        relation: "references",
+        name,
+        file: ctx.rel,
+      });
+    }
+  }
+  return edges;
+}
+
+/** The type named by `@Foo` / `@a.b.Foo(...)`. Arguments are ignored (issue #89).
+ * A scoped name is kept whole, matching heritage: a bare last segment would
+ * false-match an unrelated in-repo type (#103). */
+function javaAnnotationTypeName(anno: Parser.SyntaxNode): string | null {
+  const nameNode = anno.childForFieldName("name");
+  if (!nameNode) return null;
+  if (nameNode.type === "identifier" || nameNode.type === "scoped_identifier") return nameNode.text;
+  return null;
 }
 
 function collectTsImportBindings(
--- a/src/graph/invariants.ts
+++ b/src/graph/invariants.ts
@@ -8,8 +8,9 @@
  *   - edge relations and confidences are in the allowed sets
  *   - every edge source is a real node
  *   - every edge target is a real node, EXCEPT the relations that deliberately keep
- *     an unresolved external string (an import specifier, or a bare heritage name for
- *     an out-of-repo supertype) — those are a feature of "drop rather than guess",
+ *     an unresolved external string (an import specifier, a bare heritage name for
+ *     an out-of-repo supertype, or a Java annotation type with no in-repo
+ *     `@interface`) — those are a feature of "drop rather than guess",
  *     not a dangling edge.
  *
  * Self-loop `calls` are NOT a violation: direct recursion is a real edge a function
@@ -33,9 +34,10 @@ const CONFIDENCE = new Set<string>([
   "lsp_resolved", "lsp_dispatch", "extracted", "inferred",
 ]);
 // Relations whose target may be a deliberately-unresolved external string rather
-// than an in-repo node id: an import's module specifier, or a heritage clause naming
-// a supertype defined outside the repo (or a generic type parameter).
-const TARGET_MAY_BE_EXTERNAL = new Set<string>(["imports", "extends", "implements"]);
+// than an in-repo node id: an import's module specifier, a heritage clause naming
+// a supertype defined outside the repo (or a generic type parameter), or a Java
+// annotation whose type is not declared in-repo.
+const TARGET_MAY_BE_EXTERNAL = new Set<string>(["imports", "extends", "implements", "references"]);
 
 export interface InvariantResult {
   /** One human-readable line per violation; empty when the graph is well-formed. */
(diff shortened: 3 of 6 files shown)


dbianco:
Thanks for taking this up — the shape is right, the bare-name fallback matches heritage's contract, and the field-annotation exclusion is sound even for imported in-repo annotations (the annotation-name identifier is suppressed from the imported-symbol path via `isDeclarationName`, so fields genuinely produce nothing).

One finding worth addressing before merge: **`@Service` can collapse onto an in-repo plain `interface` of the same name** — the #103 hazard in interface form.

`resolveName(..., ["interface"], ...)` can't tell `@interface MyAnno` from `interface MyAnno` — both mint kind `interface`. So an external annotation whose bare name collides with a non-annotation interface in the repo resolves to it at `inferred`:

```java
public interface Service {}   // Service.java — plain interface, not an annotation
@Service public class Foo {}  // @Service is the external (Spring) annotation
```

→ emits `Foo → Service.java#Service` as `references`/`inferred`. That edge is false. The current guard only covers the class case (`@Entity` vs `class Entity`) and the test only pins that case; the interface case is the same family, at the same `inferred` confidence the codebase says to drop rather than guess.

The fix is cheap because the discriminator is reliable: restrict resolution to annotation-type nodes. E.g. filter the candidate by its `signature` containing the literal `@interface` — a meta-annotated annotation type's header is `@Documented @Retention(...) @Target(...) public @interface JsonAdapter`, so `startsWith("@interface")` fails there but `includes("@interface")` works, and a plain interface header can never contain `@interface`. Alternatively add an `annotationType` marker to NodeV1. A test like `@Service` + in-repo `interface Service` → stays bare would pin it.

Minor, non-blocking:
- Every `@Override`/`@Deprecated` emits a bare `references` edge to that string. Harmless for ranking (target isn't a node) but it's edge-count noise in JDK-heavy repos — worth a deliberate keep-or-drop decision (e.g. an allowlist of the well-known JDK meta-annotations).
- The invariants/graph-quality relaxation of `references` to external targets is language-agnostic. No other producer leaves an unresolved references target today, so the net isn't weakened in practice, but a future bug elsewhere would be masked. Fine as-is; a one-line note would help.

Frankie-Xu:
Thanks for the catch — `@Service` collapsing onto a plain in-repo `interface Service` was the same #103 hole, in interface form.

**Blocking: `@Service` vs in-repo `interface Service`**

Fixed in `resolve.ts`: after `resolveName(..., ["interface"], ...)`, we only accept a candidate whose `signature` `includes("@interface")`. `includes` rather than `startsWith`, because a meta-annotated header is `@Documented @Retention(...) public @interface JsonAdapter`; a plain `interface` header never contains that literal. Unresolved names still keep the bare-name fallback (heritage), rather than dropping the way PHP attributes do. No `annotationType` flag on NodeV1 — the signature check is the smaller diff.

Pinned by `Java extraction: external @Service does not collapse onto a plain interface Service (#89)`:

```java
public interface Service {}   // in-repo, not an annotation
@Service public class Foo {}  // external Spring annotation
```

→ `Foo` references `"Service"`, not `Service.java#Service`. The existing `@Entity` vs `class Entity` guard is unchanged and still green, as is in-repo `@interface MyAnno` resolution.

**Non-blocking: `@Override` / `@Deprecated` bare `references` edges — keep**

Keeping them. The heritage contract this PR already follows (and you flagged as matching) is keep-the-bare-name rather than drop; `@Override` is the same unresolved-external case as `@Entity`. Dropping every unresolved annotation would also drop `@Entity`/`@Service`. An allowlist of JDK meta-annotations is extra policy that's easy to get incomplete, and these edges don't hurt ranking (the target isn't a node). The existing test `Java extraction: external annotation keeps its bare name (#89)` already pins `@Override` as a bare `references` edge.

**Non-blocking: invariants relaxation is language-agnostic**

Added a note on `TARGET_MAY_BE_EXTERNAL` that the set is language-agnostic and that a future bug in another producer would be masked here. No code change.

shhdwi:

Reproduced #89 (fixture: `@interface MyAnno` used on a class `@MyAnno("svc")` and a method `@MyAnno`):
- **Base main:** 0 `references` edges (only `contains`).
- **This PR:** 2 `references` edges — `Service → MyAnno` and `Service.handle → MyAnno`, resolved to the in-repo `@interface`. `graft callers MyAnno` lists both with the exact annotation lines (`@MyAnno("svc")`, `@MyAnno`) — the traversable edge #89 asked for.

`graph-java.test.ts` 36/0 (incl. new #89 tests: in-repo annotation usage resolves to references, and annotation arguments don't emit extra edges), invariants 3/0, full suite 988/0. Fixes #89. Merging.

PR #208, opened 2026-08-24
feat(init): detect the package runner for generated MCP configs, add --runner (#60)

Frankie-Xu:
## Summary
- Generated MCP configs (`.mcp.json`, `.cursor/mcp.json`, `.gemini/settings.json`, `.kiro/settings/mcp.json`, `opencode.json`, Codex/Grok TOML) no longer hardcode `npx -y`. One constructor (`serverEntry`) picks the launch command for every host.
- `graft init --runner <npx|bunx|pnpm|yarn>` writes that runner into every graft MCP entry. Re-running `init` updates an existing graft entry (JSON merge and TOML `[mcp_servers.graft]` alike). `--runner` is recorded in the wiring stamp so a later refresh replays it.
- Without the flag, the repo lockfile decides (bun/pnpm/yarn beat a local `graft` on PATH so committed configs stay portable for teammates). `GRAFT_MCP_NPX=1` remains the test pin / escape hatch, below `--runner`.

### Detection (first match wins)

| Lockfile | Launch |
|---|---|
| `bun.lock` or `bun.lockb` | `bunx @nanonets/graft mcp` |
| `pnpm-lock.yaml` | `pnpm dlx @nanonets/graft mcp` |
| `yarn.lock` | `yarn dlx @nanonets/graft mcp` |
| none / `package-lock.json` | `graft mcp` if `graft` is on PATH, else `npx -y @nanonets/graft mcp` |

`bunx` is not passed `--bun`: graft's shebang is node and the tree-sitter addons expect it. bunx still fetches the package without npm ([bunx docs](https://bun.sh/docs/pm/bunx)). `pnpm dlx` / `yarn dlx` match the [pnpm](https://pnpm.io/cli/dlx) and [yarn](https://yarnpkg.com/cli/dlx) docs; neither needs an npx-style `-y`.

```
graft init --runner bunx
graft init --runner pnpm
```

## Test plan
- [x] lockfile detection: bun.lock / bun.lockb / pnpm-lock.yaml / yarn.lock / no lockfile
- [x] `--runner` beats lockfile, PATH, and `GRAFT_MCP_NPX`
- [x] re-run updates JSON and TOML graft entries; same runner stays `unchanged`
- [x] CLI `--runner bunx` writes bunx into `.cursor/mcp.json` and `.mcp.json`
- [x] unknown `--runner` exits non-zero and writes nothing
- [x] wiring stamp replays `--runner` on refresh
- [x] `npm run build`
[...]

Issue #60, opened 2026-08-06
graft init hardcodes npx in generated MCP configs; no way to specify an alternative runtime.


JuxtNorth:
## What happened

I ran `graft init` in a project that uses **Bun** as its sole runtime (no npm/npx). The generated MCP configuration files all contained:

```json
{
  "command": "npx",
  "args": ["-y", "@nanonets/graft", "mcp"]
}
```

This was written into:
- `.mcp.json`
- `.gemini/settings.json`
- `opencode.json`
- `.kiro/settings/mcp.json`

I had to manually edit every single one of these to swap `npx -y` for `bunx --bun`.

## What I expected

`graft init` should either:

1. **Auto-detect the package manager** - check for `bun.lockb`, `pnpm-lock.yaml`, `yarn.lock`, etc. and use the matching runner (`bunx`, `pnpx`, `yarn dlx`).
2. **Accept a flag** - something like `--pm bun` or `--runner bunx` so I can specify it once and have all generated configs use the right command.

## Why this matters

More and more projects enforce a single runtime. Having to hunt through 4+ generated files after every `graft init` to fix the package runner is tedious and error-prone — especially since re-running `graft init` overwrites the fixes.

## Environment

- OS: Windows 11
- Runtime: Bun 1.3.14
- Graft: 0.8.0

PR #210, opened 2026-08-24
fix(init): generated instructions work without a global graft install (#61)

### LINKED ISSUE #61: Agent instructions reference bare graft CLI but graft init doesn't install the global binary
## What happened

After running `graft init`, the injected instructions in `AGENTS.md` and `GEMINI.md` tell AI agents to run commands like:

```
graft ask "<your question>" --source
graft skeleton <file>
graft callers <symbol>
graft build
```

But `graft init` doesn't install the global `graft` binary. The first time an AI agent tries to follow these instructions, it gets:

```
graft: The term 'graft' is not recognized as a name of a cmdlet, function, script file, or executable program.
```

I had to separately run `bun install -g @nanonets/graft` (or `npm install -g @nanonets/graft`) to make the commands work.

## What I expected

Either:

1. **`graft init` should prompt or auto-install the global CLI** when it detects the binary isn't on PATH.
2. **The generated instructions should use `npx`/`bunx` fallbacks** e.g. `npx @nanonets/graft ask "..."` so agents can run Graft without a global install.
3. **At minimum, print a warning**, something like _"Note: `graft` CLI is not on your PATH. Run `npm install -g @nanonets/graft` to enable the commands referenced in AGENTS.md."_

## Why this matters

The whole point of `graft init` is zero-friction setup. If the generated instructions don't work out of the box, agents silently fail or waste turns debugging a missing binary. This is especially confusing because `graft init` itself ran fine (via `bunx`), so you'd assume everything is wired up.

## Environment

- OS: Windows 11
- Runtime: Bun 1.3.14
- Graft: 0.8.0

Frankie-Xu:
## Summary
- `graft init` does not install a global `graft` binary, but AGENTS.md / GEMINI.md / SKILL.md (and the other host copies of the same body) told agents to run bare `graft ask`. The first command then failed with command-not-found.
- Chose **scheme 2** (a fallback note) over rewriting every example as `npx @nanonets/graft …`: smaller diff, and re-running `init` only replaces graft's fenced section / owned skill file rather than making every listed command noisier for repos that already have the CLI.
- One shared sentence, `cliFallbackNote()`, is interpolated into `instructionBody()` and `skillTemplate()`. Agents that hit a missing binary are told to use `npx -y @nanonets/graft` in its place (example included); `bunx` / `pnpm dlx` / `yarn dlx` are named as equivalents. Commands stay `graft ask` so a global install keeps working unchanged.

Independent of #60 / #208 — the note is a static npx fallback, not lockfile detection. Those can merge later if instructions should follow `--runner`.

## Test plan
- [x] `hosts-instructions` and `claude-skill-template` assert the fallback wording
- [x] `graft init --agents claude cursor agents --no-global` writes the note into AGENTS.md, `.cursor/rules/graft.mdc`, and SKILL.md
- [x] `graft init --dry-run` still only lists paths
- [x] `npm run build`
- [x] related init/instruction tests green; full `npm test` matches origin/main aside from an unrelated `viz --tabs` failure when this worktree has no `graft/` graph

Closes #61

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- **Generated agent instructions work without a global `graft` install.**
+  `graft init` never puts `graft` on PATH, but AGENTS.md / GEMINI.md / SKILL.md
+  told agents to run bare `graft ask`. They now say to use
+  `npx -y @nanonets/graft` (or `bunx` / `pnpm dlx` / `yarn dlx`) when the
+  binary is missing.
+
 ## 0.18.0
 
 ### Added
--- a/src/claude/skill-template.ts
+++ b/src/claude/skill-template.ts
@@ -1,3 +1,5 @@
+import { cliFallbackNote } from '../hosts/instructions.js';
+
 // The graft Claude Code skill, bundled as a string so `graft init` can write it into a
 // consumer repo's .claude/skills/graft/SKILL.md (no network fetch, version-locked to the
 // installed graft). This is the single source of truth for the skill text; graft's own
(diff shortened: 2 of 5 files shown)


Frankie-Xu:
Rebased onto current main. Waiting on CI before marking ready.

Frankie-Xu:

PR #214, opened 2026-08-24
fix(build): fall back to WASM grammars when native tree-sitter is unavailable (#119)

### LINKED ISSUE #119: ARM64 Incompatibilies
The @nanonets/graft CLI cannot be initialized on an ARM64 Linux platform due to a fundamental native module limitation.
What happened:
1. npm install -g @nanonets/graft ✅ — Successfully installed (36 packages)
- Warnings about blocked install scripts (expected on this platform)
2. graft init ❌ — Failed with native build error:
Error: No native build was found for platform=linux arch=arm64
runtime=node abi=147 uv=1 armv=8 libc=glibc node=26.6.0
- The tree-sitter native module (required by graft) has no pre-built binary for ARM64 Linux
- node-gyp cannot compile the C++ addon on this architecture
Why this happens:
- @nanonets/graft uses tree-sitter for parsing source code into ASTs
- tree-sitter compiles native C++ addons via node-gyp
- ARM64 Linux lacks pre-built binaries in the npm registry for this combination
- node-gyp build fails at runtime when trying to require('tree-sitter')
Result:
- No graft/ folder was created — the init step errored before any graph could be built
- 0 map files built — the process errored out during initialization

Frankie-Xu:
## Summary
- `graft init` / `graft build` no longer crash on linux/arm64 when `tree-sitter@0.21.1` has no native prebuild.
- Native parsers still load when they exist (darwin, linux-x64). If `require("tree-sitter")` fails, depth languages (TypeScript/JavaScript, Python, Go, Java, Kotlin, PHP, R) extract through the existing WASM breadth path.
- The build prints `native parser unavailable — falling back to WASM (reduced binding fidelity)`. `graft check` and the statusline do not treat that as an error.

Closes #119.

## Root cause
Failure is at **`require("tree-sitter")` on CLI load**, not `npm install` and not `scripts/postinstall.mjs` (that script is a one-line nudge and never throws). `tree-sitter@0.21.1` ships prebuilds for darwin-arm64/x64, linux-x64, and win32-x64 only — **no linux-arm64**. With install scripts blocked, install succeeds and every command then dies with `No native build was found for platform=linux arch=arm64`.

Bumping to `tree-sitter@0.25` (#40) would pick up linux-arm64 prebuilds, but that PR is still **OPEN, CONFLICTING, last Ubuntu CI failed**, and Alpine/musl / “no compiler” would still lose. Mixing `0.25` with today’s `tree-sitter-typescript@0.23.2` is not a small bump.

## Fidelity (not papered over)
WASM fallback has **no** `bindings` / receiver-type resolution. Call edges come from `tags.scm` when the query compiles (symbols only if it does not). Depth extensions stay owned by `languageOf`; `genericLangOf` still does not claim `.ts`/`.py`/`.go`/etc., so darwin/linux-x64 native extraction is unchanged.

## Tests
- `test/native-fallback.test.ts`: mock native throw → empty grammars; `{ wasm: true }` still emits function nodes with `origin: "generic"`; default path on darwin stays `origin: "ast"` and `warnings: []`; `graft check` stays OK.
- Full `npm test` on darwin: **919 pass**. The 2 failures (`PageRank` timing, `viz --tabs` missing a local graph) are unrelated flakes/environment.

## Docker evidence (linux/arm64, Node 22)
`tree-sitter@0.21.1 --ignore-scripts` then `require("tree-sitter")`:

```
REQUIRE_FAIL
No native build was found for platform=linux arch=arm64 runtime=node abi=127 uv=1 armv=8 libc=glibc node=22.23.2
```

Prebuilds in the tarball: `darwin-arm64`, `darwin-x64`, `linux-x64`, `win32-x64` — no `linux-arm64`.

After this change, `graft build` of a tiny `hello.ts` (`hello()` → `world()`):

```
NATIVE_FAIL
No native build was found for platform=linux arch=arm64 ...
✓ wiring: 3 nodes (2 function, 1 file), 1 edges, 1 cards [typescript]
⚠ native parser unavailable — falling back to WASM (reduced binding fidelity)
graph check: OK
```

Nodes are `origin: generic` (`hello.ts`, `hello.ts#hello`, `hello.ts#world`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- **`graft init` / `graft build` no longer crash on linux/arm64** when
+  `tree-sitter@0.21.1` has no native prebuild (`No native build was found for
+  platform=linux arch=arm64`). Native parsers still load when they exist
+  (darwin, linux-x64). If `require("tree-sitter")` fails, depth languages
+  (TypeScript/JavaScript, Python, Go, Java, Kotlin, Swift, PHP, R) fall back to the
+  existing WASM breadth grammars. Fidelity is reduced: no bindings /
+  receiver-type resolution; call edges come from `tags.scm` (symbols only if
+  a query is missing). The build prints
+  `native parser unavailable — falling back to WASM (reduced binding fidelity)`
+  and `graft check` / the statusline do not treat it as an error. (#119)
+
 ## 0.18.0
 
 ### Added
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -547,6 +547,7 @@ program
       { repo: buildRoot },
     );
     for (const e of g.errors) console.error(`✗ ${e}`);
+    for (const w of g.warnings) console.error(`⚠ ${w}`);
 
     const rel = relative(process.cwd(), g.contextDir) || "graft";
     if (process.env.GRAFT_NO_GITIGNORE) {
--- a/src/graph/build.ts
+++ b/src/graph/build.ts
@@ -17,7 +17,7 @@ import { readFileSync } from "node:fs";
 import { basename, dirname, resolve } from "node:path";
 import { walkDir } from "../ingest/fs.js";
 import { contextDirFor, ensureGitignored, ensureSearchable } from "../context/node-file.js";
-import { extractFile, languageLabelOf, languageOf, type RawEdge } from "./extract.js";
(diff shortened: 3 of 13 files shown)


Frankie-Xu:
Rebased onto current main. Kept `--only-dir` fingerprint checking, added Swift to the WASM fallback set (it is depth-tier now), and kept `strip`/container warmup from main. Waiting on CI before marking ready.

Frankie-Xu:
Rebased onto current main. CI is green — please review when you can.

PR #215, opened 2026-08-24
fix(check): don't treat root-level file cards as missing concept nodes (#213)

### LINKED ISSUE #213: `graft check` reports permanent `indexDrift` for source files at the repository root
## Summary

After a clean `graft build --deep`, `graft check` immediately reports `STALE`
for every source file that lives at the **repository root**. The drift never
clears — re-running `graft build --deep` reproduces it exactly.

The cause is that per-file cards for root-level sources are written to
`graft/<stem>.md`, i.e. the same directory `readNodes()` scans for **concept
node** cards. `readNodes()` treats every top-level `*.md` as a node and looks
its slug up in `manifest.nodes`, but per-file cards are recorded in
`manifest.files` — so they can never be found.

Files in subdirectories are unaffected, because their cards land in
`graft/<dir>/…` and `readNodes()` is not recursive.

## Minimal reproduction

Two files, one at the root and one nested:

```bash
mkdir -p repro/lib && cd repro
echo 'export function rootLevelFn(a: number): number { return a * 2; }' > root.ts
echo 'export function nestedFn(a: number): number { return a + 1; }'    > lib/inner.ts
git init -q . && git add -A && git commit -qm init

npx -y @nanonets/graft@0.10.1 build --deep     # completes successfully
npx -y @nanonets/graft@0.10.1 check --json
```

Cards produced:

```
graft/INDEX.md
graft/lib-inner-ts.md      <- concept node  (in manifest.nodes)
graft/lib/inner.md         <- per-file card (in manifest.files)  -> not scanned
graft/root-ts.md           <- concept node  (in manifest.nodes)
graft/root.md              <- per-file card (in manifest.files)  -> SCANNED as a node
```

Note: the bug only manifests after a `--deep` build (which requires an LLM
key). Without the deep layer `context.missing` is `true` and this check path
never runs, so a structural-only build will not reproduce it.

`check --json`, run immediately after the build that produced those cards:

```json
{
  "context": { "ok": false, "indexDrift": ["root: node file not in manifest"] },
  "graph":   { "ok": true }
}
```

Expected: `context.ok === true` — nothing changed between build and check.
Actual: permanent `indexDrift` for `root`, none for `lib/inner`.

## Root cause

`dist/context/node-file.js` — `readNodes()` is a flat `readdirSync`, and when a
card has no `slug` in its frontmatter it falls back to the **filename**:

```js
export function readNodes(dir) {
  for (const entry of readdirSync(dir)) {           // top level only
    if (!entry.endsWith(".md") || entry === "INDEX.md") continue;
    const fm = matter(readFileSync(join(dir, entry), "utf8")).data;
    out.push({ slug: String(fm.slug ?? entry.replace(/\.md$/, "")), ... });
  }
}
```

`dist/context/check.js` then requires every returned slug to exist in
`manifest.nodes`:

```js
const manifestNodes = new Map(manifest.nodes.map((n) => [n.slug, n]));
for (const node of readNodes(outDir)) {
  if (!manifestNodes.get(node.slug))
    result.indexDrift.push(`${node.slug}: node file not in manifest`);
}
```

For `graft/root.md` the fallback yields slug `root`, which is not a node — the
manifest lists it under `files` as `root.ts`.

The two card kinds are already distinguishable by frontmatter. In the
reproduction above:

| card | frontmatter |
| --- | --- |
| `graft/root-ts.md` (concept node) | `name`, `slug`, `type`, `sources`, `sources_digest`, `links` |
| `graft/root.md` (per-file card) | `covers: []` — **no `slug`** |
| `graft/lib/inner.md` (per-file card) | none at all |

## Why this can't be worked around

- `graft build` has no exclude/ignore option. `--extensions` is an *include*
  list and `--include-dir` only widens `SKIP_DIRS`.
- The affected files usually cannot be moved. In our repo they are
  `next.config.ts`, `next-env.d.ts` and `playwright.config.ts` — all three are
  required by their frameworks to sit at the repository root.

## Impact

`graft check` is documented as the CI/pre-commit freshness gate ("Fail if
`graft/` is stale relative to the code (for CI)"). Any repository with a
root-level source file therefore has a gate that fails permanently, on every
commit, regardless of whether anything is actually stale.

The practical outcome is worse than a noisy gate. The available workarounds are
to stop running the check or to filter out `indexDrift`; both weaken exactly the
protection versioned cards exist to provide, because `indexDrift` is the only
detector for a stray or orphaned `.md` landing in `graft/`.

## Suggested fix

Have `check` cross-reference `manifest.files` before reporting `indexDrift`: if a
slug resolves to a recorded **file card**, stay silent; otherwise report it as
today.

```js
const manifestNodes = new Map(manifest.nodes.map((n) => [n.slug, n]));
// per-file cards are recorded under `files`, not `nodes` — a slug that resolves
// to one is a file card, not an orphaned node.
const fileStems = new Set(
  manifest.files.map((f) => f.path).filter((p) => !p.includes("/"))
                .map((p) => p.replace(/\.[^.]+$/, "")),
);
for (const node of readNodes(outDir)) {
  if (!manifestNodes.get(node.slug) && !fileStems.has(node.slug))
    result.indexDrift.push(`${node.slug}: node file not in manifest`);
}
```

I verified this against the same `--deep` graph (424 cards, 137 at the top
level): it silences exactly the three false positives and nothing else, and a
hand-dropped `notes.md` / `TODO.md` / `scratch.md` in `graft/` is still
reported. The detector keeps full strength.

### A simpler variant, if the frontmatter contract is load-bearing

Real node cards carry `slug` in frontmatter; per-file cards do not. So
`readNodes()` could skip entries without one:

```js
if (!fm.slug) continue;               // per-file card, not a node
```

Measured on the same graph:

| | count |
| --- | ---: |
| top-level `*.md`, excluding `INDEX.md` | 137 |
| entries in `manifest.nodes` | 134 |
| top-level cards **with** `slug` in frontmatter | 134 |
| of those, missing from `manifest.nodes` | **0** |
| top-level cards **without** `slug` | 3 — `next.config.md`, `next-env.d.md`, `playwright.config.md` |

I am **not** proposing this as the primary fix, for two reasons. It would
silently skip a hand-written `.md` dropped into `graft/` without frontmatter —
weakening the very detector this issue argues for. And the existing
`fm.slug ?? entry.replace(...)` fallback suggests slug-less node files were
anticipated at some point; my measurements only sample cards that 0.10.1
generates itself, so they cannot rule that out. Use it only if you know the
frontmatter contract holds.

### Other options

1. Write per-file cards for root-level sources into a subdirectory
   (e.g. `graft/_root/root.md`), so the node namespace stays exclusive.
2. Add an ignore/exclude option to `build` and `check`.

## Environment

- graft `0.10.1` (via `npx -y @nanonets/graft@0.10.1`)
- Node.js `v25.9.0`
- macOS (arm64)
- Reproduced on both a two-file scratch repo (above) and a ~340-source-file
  Next.js repository, where the drift lists exactly the three root-level
  framework files and nothing else.

Frankie-Xu:
## Summary
- After a clean `graft build --deep`, `graft check` treated every **root-level** source file as permanent `indexDrift` (`<stem>: node file not in manifest`). Nested files were fine: their per-file cards live in `graft/<dir>/…`, which `readNodes()` does not recurse into.
- Root-level per-file cards are written to `graft/<stem>.md` — the same directory concept nodes use — but they are recorded in `manifest.files`, not `manifest.nodes`. `readNodes()` used the filename as a slug and `check` then looked that slug up in `manifest.nodes`.
- Fix is on the read side (option A): `readNodes()` skips a top-level `.md` whose stem matches a **root** path in `manifest.files` and that has no concept `slug` in frontmatter. Concept nodes are unchanged. A hand-dropped `graft/notes.md` is still reported as `indexDrift`.

## Test plan
- [x] New test: root `main.ts` + nested `src/foo.ts`, `buildContext` (fake LLM) then `buildGraph`; `checkContext` is clean (`indexDrift: []`); a stray `graft/notes.md` still fails
- [x] Existing check tests (content drift, coverage, include-dir, re-init) still pass
- [x] `npm run build && npm test` — 916 pass, 0 fail
- [x] Manual two-layer repo (root `main.ts` + `src/foo.ts`):

```
$ graft check --json
{
  "context": {
    "ok": true,
    "missing": false,
    "contentDrift": [],
    "removed": [],
    "coverage": [],
    "indexDrift": []
  },
  "graph": { "ok": true, "missing": false, … }
}

graft check: OK — the graph is in sync with the code.
graph check: OK — the wiring graph is in sync with the code.
```

`--deep` here is the context layer plus wiring cards (same layout as `graft build --deep`); no LLM path was changed.

Closes #213

### DIFF
--- a/src/context/node-file.ts
+++ b/src/context/node-file.ts
@@ -253,15 +253,39 @@ export interface ParsedNode {
   links: NodeLink[];
 }
 
-/** Read and parse every `.md` node file in a context dir (skips the manifest). */
+/**
+ * Stems of per-file wiring cards that sit at the graft/ top level — i.e. cards
+ * for source files in the repo root. Those cards share a directory with concept
+ * nodes (`graft/<stem>.md`) but are recorded in `manifest.files`, not
+ * `manifest.nodes`. Nested file cards live in subdirs and are never scanned here.
+ */
+function rootFileCardStems(dir: string): Set<string> {
+  const manifest = readManifest(dir);
+  if (!manifest) return new Set();
+  const stems = new Set<string>();
+  for (const f of manifest.files) {
+    if (f.path.includes("/")) continue;
+    stems.add(f.path.replace(/\.[^./]+$/, ""));
+  }
+  return stems;
+}
+
+/** Read and parse every concept-node `.md` in a context dir (skips INDEX.md
+ * and root-level per-file cards already recorded in `manifest.files`). */
 export function readNodes(dir: string): ParsedNode[] {
   if (!existsSync(dir)) return [];
+  const fileCardStems = rootFileCardStems(dir);
   const out: ParsedNode[] = [];
   for (const entry of readdirSync(dir)) {
     if (!entry.endsWith(".md") || entry === "INDEX.md") continue;
     const fm = matter(readFileSync(join(dir, entry), "utf8")).data as Record<string, unknown>;
+    const fallbackSlug = entry.replace(/\.md$/, "");
+    // A root-level file card has no concept `slug` and its stem matches a
+    // recorded source file. Skipping it here (instead of all slug-less `.md`)
+    // keeps a hand-dropped notes.md visible to check's indexDrift detector.
+    if (fileCardStems.has(fallbackSlug) && fm.slug == null) continue;
     out.push({
-      slug: String(fm.slug ?? entry.replace(/\.md$/, "")),
+      slug: String(fm.slug ?? fallbackSlug),
       name: String(fm.name ?? ""),
       type: String(fm.type ?? ""),
       sources: Array.isArray(fm.sources) ? (fm.sources as SourceRef[]) : [],
--- a/test/context.test.ts
+++ b/test/context.test.ts
@@ -11,6 +11,7 @@ import { execFileSync } from "node:child_process";
 import { buildContext } from "../src/context/build.js";
 import { checkContext, indexFreshness, staleBanner } from "../src/context/check.js";
 import { contextDirFor, ensureGitignored, ensureSearchable } from "../src/context/node-file.js";
+import { buildGraph } from "../src/graph/build.js";
 import { writeBuildConfig } from "../src/util/state.js";
 import { fakeProviders } from "./helpers.js";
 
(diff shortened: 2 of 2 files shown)


shhdwi:

Reproduced #213 on an identical `graft/` dir (root `main.ts` + nested `src/foo.ts`), running `check --json` from each code version:
- **Base main:** `context.ok = false`, `indexDrift: ["main: node file not in manifest"]`, exit 1 — permanent false drift on the root-level file card.
- **This PR:** `context.ok = true`, `indexDrift: []`, exit 0.

Only `readNodes()` changed between runs, proving cause. Fix skips a top-level `graft/<stem>.md` only when its stem matches a recorded root source file and it has no concept `slug`; nested files and concept nodes are untouched. New test `check: root-level file card is not indexDrift after build (#213)` also asserts a stray `graft/notes.md` is still flagged, so the detector isn't weakened. `context.test.ts` 25/0. Fixes #213. Merging.

PR #219, opened 2026-08-24
feat(graph): full-fidelity Swift extraction

mattstrayer:
## What

Promotes Swift (`.swift`) from the generic breadth tier to a hand-written depth-tier extractor, following the Kotlin promotion (#130). Today the breadth tier's `swift.scm` has **no `@reference.call` captures at all**, so a Swift repo gets symbols and zero wiring — no `callers`, no blast radius, no coupling-ranked retrieval.

## How

**Extractor** (`src/graph/extract.ts`, mirroring `describeKotlin`):
- One `class_declaration` node covers `class` / `struct` / `enum` / `actor` / `extension`, told apart by their own keyword token → `class`, `struct`, `enum`, `class`, and the extension case below. `protocol` → interface, `typealias` → type, top-level `let`/`var` → variable.
- **Extensions**: `extension Point` takes the extended type's name, so its members mint as `Point` methods and member calls on a Point receiver resolve to them. The extension node itself is kind `module`, not a second class named `Point` — that duplicate made every `Point()` initializer call and `: Point` heritage target ambiguous (resolve never guesses), and a test pins the regression.
- `init` → a method named after its type (like Java's `constructor_declaration`); `func` → method inside any type, function elsewhere.
- Calls via `call_expression` with `navigation_expression` / `self` / `super` receivers; `:` inheritance clause → `extends` edges (bare names — Swift can't say syntactically which specifier is the superclass); `import` → module-path import edges.
- Visibility: only `private` / `fileprivate` hide a symbol. Swift's default `internal` is module-wide, and a repo is typically one module, so it counts as API surface.

**Resolution** (`src/graph/resolve.ts`, two existing precedents reused):
- A bare lowercase call inside a type body is ONE edge carrying both readings in Swift's inner-scope-first order: the member reading first (owner-qualified index + in-repo ancestor chain, so `clearLogs()` in a test subclass finds the base class's method), then the free-function reading only when no member exists on the chain — a name defined as both yields the member edge alone, as Swift dispatches it.
- `super.method()` resolves against the declaration's own superclass (the first `:` entry — Swift's grammar puts the superclass before any protocol), climbing past the current class's override.
- Overloads disambiguate by declared arity vs call-site argument count, Java's exact mechanism (defaults/variadics make the arity a minimum); a set arity can't split (`save(Int)` vs `save(String)`) drops rather than taking the same-file first-declaration tiebreak.
- `Animal(legs: 4)` is an ordinary call node with no `new`; it falls back to class/struct/enum targets once functions find nothing — Python's constructor-fallback shape.

**Dependency**: `tree-sitter-swift@^0.7.1` — the first release whose install compiles the shipped `parser.c` instead of regenerating it via tree-sitter-cli (0.6.0's gyp action breaks under npm hoisting: `No rule to make target node_modules/tree-sitter-cli`). Its peer wants `tree-sitter@^0.22`; a root `overrides` entry pins it to the repo's 0.21 runtime for dev installs, and I verified a transitive (end-user) install resolves cleanly without any override, with the compiled grammar loading and parsing fine on the 0.21 runtime (ABI 14).

## Receiver-type bindings

A second commit adds the Swift `collectBindings` pass, so untyped receivers now resolve too. Bound clues, all syntax-local: typed parameters (argument labels handled), typed properties (one optional level unwrapped), initializer-call assignments (`let vet = Vet()` — an UpperCamelCase callee names the constructed type, the same convention trust as Go's `NewX`), and fields bound both bare and `self.`-prefixed like Java's. `self.repo.save()` binds one hop off self like TS's `this.x`, and `Animal.census()` resolves via an uppercase-receiver rule mirroring PHP's scoped calls. A receiver with no local clue (a chained call's result) stays unresolved rather than guessed.

## Dogfooded on two real Swift codebases

Alamofire (108 files, ~43k lines) and swift-composable-architecture (796 files, ~79k lines): both build 100% of files with zero parse failures, `graft check` reports OK (deterministic) on both, and the graphs come out at 3,067 nodes / 6,530 edges and 6,127 nodes / 12,433 edges respectively.

Sampling cross-file `inferred` method-target call edges caught two false-positive classes in the first implementation of the implicit-self rule (then a bare-name `kinds` widening): `Text("hi")` — SwiftUI's own init — bound to an unrelated in-repo `extension Text { init(template:) }`, and `contains(element)` inside `extension Set` — a stdlib call — bound to an unrelated type's only in-repo `contains`. The third commit replaces the widening with the ancestor-chain intent described above; after it, both classes are gone, while 280 inherited `stored(...)` calls in Alamofire's tests and 144 `clearLogs`/`assertLogs` calls in TCA's still resolve to their base-class methods through the chain.

## Testing

- New `test/graph-swift.test.ts`: 20 tests — kind mapping for all five `class_declaration` keywords, extension member ownership + the ambiguity regression, initializer methods, call shapes (bare / member / self / super / trailing closure), heritage, imports, visibility incl. `private(set)`, and resolution-level pins through `buildGraph` (implicit-self, constructor edges, ambiguous-drops, receiver-typed member calls incl. extension methods).
- `test/graph-languages.test.ts` / `test/supported-extensions.test.ts` updated for the tier move.
- Full suite: 935/935 pass.
- End-to-end on a demo Swift package: `extends Animal -> Greeter`, `run -> walk` (extension → class), `walk -> greet` (implicit self), `makeAnimal -> Animal` (initializer, inferred), cross-file bare calls — all resolving.

README's supported-languages section is also brought current (Kotlin and PHP were still listed as breadth-tier; Nix was missing).


mattstrayer:
Tested against two substantial real-world Swift codebases before review:

- [Alamofire](https://github.com/Alamofire/Alamofire) — 108 files, ~43k lines of protocol/extension-heavy Swift
- [swift-composable-architecture](https://github.com/pointfreeco/swift-composable-architecture) — 796 files, ~79k lines of modern Swift (actors, macros, result builders)

Results on both: 100% of files parsed with zero failures, `graft check` reports OK (deterministic rebuild), no malformed nodes. Graphs came out at 3,067 nodes / 6,530 edges (Alamofire) and 6,127 nodes / 12,433 edges (TCA).

Sampling cross-file `inferred` method-target call edges on these corpora caught two false-positive classes in the first implicit-self implementation (`Text("hi")` binding to an unrelated in-repo `extension Text { init(template:) }`; `contains(element)` inside `extension Set` binding to an unrelated type's only `contains`). Commit 7c9f129 replaces that mechanism with the ancestor-chain intent described in the PR body — after it, both classes are gone while the correct inherited-method edges remain (280 `stored(...)` calls in Alamofire's tests, 144 `clearLogs`/`assertLogs` calls in TCA's, all resolving to their base-class methods through the chain).

mattstrayer:
An adversarial review pass (Codex) against this branch surfaced three dispatch-precision findings; commit c3e4c21 addresses all three:

1. **Bare-call double edges** — a lowercase bare call in a type body emitted both a member intent and a free-function intent, and when both targets existed (`func helper()` at top level AND `A.helper`) the graph kept both. Now it is one edge resolved in Swift's inner-scope-first order: member reading first, free-function reading only when the whole owner chain has no such member.
2. **Overloads picked the first declaration** — with no arity data, `save("x")` against `save(Int)`/`save(String)` resolved to whichever overload appeared first in the file, stamped `extracted`. Swift callables now carry declared arity (defaults/variadics mark it a minimum) and call sites carry argument count (trailing closure included), reusing Java's existing `narrowByArity`; a set arity can't split drops instead of guessing by source order.
3. **`super.method()` never resolved** — `resolveRecvType` mapped `super` through a field only R populated. Swift class declarations now populate it from their own `:` clause's first entry, so a `super.ping()` in an override resolves to the parent's method, past the current class's own.

Four new tests pin member-vs-global precedence, equal-arity drop, distinct-arity resolution, and super dispatch (Swift suite now 20 tests; full suite 935/935). Re-dogfooded on Alamofire and swift-composable-architecture: 100% parse, `graft check` OK, duplicate-reading and first-overload edges gone.

shhdwi:
Verified locally on top of current `main`. Rebased two trivial conflicts (CHANGELOG keep-both; package.json — kept the renamed `@davisvaughan/tree-sitter-r` key from #233 alongside the new swift `allowScripts` entry + `overrides`). `extract.ts`/`resolve.ts` auto-merged.

Depth-tier Swift (native `tree-sitter-swift@0.7.1`, no committed binaries). Behavioral on a class+struct+protocol+cross-type fixture: 12 nodes, 16 edges, all resolved — `extends` (inheritance + protocol conformance), implicit-self calls up the chain, and **receiver-type resolution** (`Zoo.openGates → Dog.feed` via a `keeper: Dog` binding); ambiguous same-name overrides dropped, not guessed. Full suite **1040/1040** incl. 8 `graph-swift` cases. Merging.

Issue #73, opened 2026-08-10
Initialized Git submodules are silently excluded from monorepo graph


yescine:
## Summary

Graft 0.9.0 silently excludes initialized Git submodules when the working directory is a Git superproject.

The graph indexes source tracked directly by the parent repository, but source inside initialized submodules is absent. The graph can therefore appear healthy and fresh while omitting significant code.

## Environment

- Graft: `@nanonets/graft@0.9.0`
- Host: Codex through the Graft MCP server
- Repository shape: Git superproject with initialized nested submodules

## Steps to reproduce

1. Create a Git repository containing normal tracked source files.
2. Add and initialize a submodule at `path/to/submodule`.
3. Run:

```sh
graft build
graft map
graft ask "project entrypoint" --in path/to/submodule
```

4. Query or inspect the submodule path in the resulting graph.

## Actual behavior

The root graph indexes only files tracked directly by the superproject. A scoped lookup returns:

```text
nothing indexed under "path/to/submodule/" (or any path prefix)
```

The freshness check still reports that the wiring graph is synchronized because the omitted submodule files are not part of the fingerprint.

## Evidence pointing to file enumeration

At the superproject root:

```sh
git ls-files --cached -- path/to/submodule
```

returns only the gitlink. However:

```sh
git ls-files --recurse-submodules --cached -- path/to/submodule
```

returns the tracked files from that initialized submodule.

The current root enumeration uses the equivalent of:

```sh
git ls-files --cached --others --exclude-standard
```

Git does not recurse into gitlinks in that mode. Git also rejects combining `--recurse-submodules` with `--others`:

```text
fatal: ls-files --recurse-submodules unsupported mode
```

The `--include-dir` option does not solve this because it only overrides built-in skipped directory names; the root Git command never returns the submodule's internal files.

## Expected behavior

An initialized submodule should either:

1. be included in a unified graph with paths such as `path/to/submodule/src/index.ts`; or
2. be discovered as a child repository and federated automatically with the parent graph.

Scoped queries, MCP searches, freshness checks, and post-edit refreshes should cover the same submodule content.

## Suggested product direction

A supported opt-in such as `--follow-submodules` would be sufficient.

Possible approaches:

- enumerate tracked files with `git ls-files --cached --recurse-submodules`;
- enumerate untracked/non-ignored files separately in the root and each initialized submodule, prefixing child paths;
- discover initialized children from `.gitmodules` and build or federate their graphs;
[...]

PR #232, opened 2026-08-25
feat: add opt-in indexing of nested git clones

tpoignonec:
## The issue

`followSubmodules` only follows gitlinks. A repo that was simply cloned into the tree has no index entry, so it can never carry mode `160000` and is never indexed (even with `followSubmodules: true` set).

It fails silently: the omitted files never enter the fingerprint, so the graph reports itself fresh.
Similar to #73/#74, one Git relationship over.

Manifest-driven multi-repo tools often check dependencies out this way: `west` (Zephyr), `repo` (Android), or `vcstool` in robotics, among others.
My use case is the last of those, a ROS 2 workspace whose source dependencies are imported with `vcstool`: the parent repo has few indexable files, but many imported repos, so most the reachable code is invisible to the graph. 
A single project graph allows cross links between the packages imported through `vcstool`: an import or `#include` crossing two of them resolves to a real edge rather than an unresolved string, which building each repo separately cannot do.

The same applies to anyone who clones an upstream into `deps/` to patch it locally.

## Git already marks the boundary, in the command graft already runs

```
$ git ls-files -t --stage --cached --others --exclude-standard
? nested_clone/                 <- collapsed to a DIRECTORY, trailing slash
? plain_untracked_dir/a.txt     <- expanded to individual files
? plain_untracked_dir/b.txt
H 100644 587be6b4c3f9... 0	tracked.txt
```

Git expands untracked directories file by file. The only time it emits a bare directory entry is when it cannot descend, which means a repository boundary. `gitVisibleFiles` already runs this exact command, so detection costs no extra process, syscall, or Git call.

## The change

New opt-in `--follow-nested-repos`, default off, persisted in `.graft/config.json` exactly like `--follow-submodules`. Three hunks in `src/ingest/fs.ts` mark trailing-slash entries as nested and widen the `if (entry.gitlink)` recursion trigger to cover them. Cycle protection, skip policy, path handling and fail-soft are reused unchanged. `gitVisibleFilesShallow` is untouched, so the default walk is byte-for-byte what it was. No new dependencies.

Kept separate from `followSubmodules` rather than widening it, so nobody already opted in gets silently re-indexed on upgrade, and because a pinned submodule and a stray clone are not the same relationship. Neither flag implies the other.

## Testing

`test/graph-follow-nested-repos.test.ts`, shaped after `graph-follow-submodules.test.ts`: opt-in, persistence across a no-flag rebuild, explicit reversal, a cross-repo import resolving to a real edge, `SKIP_DIRS` still applying inside a followed clone, and the regression guard that a plain untracked directory is still expanded file by file rather than mistaken for a repo boundary.


shhdwi:
Verified locally on top of current `main` (rebased a one-line `cli.ts` opts-type conflict — kept `followNestedRepos` alongside the `gitignore`/`ignore` fields; fs.ts/state.ts auto-merged). Build clean.

Before/after (repo with tracked `top.ts` + an untracked nested git clone `nested_dep/nested.ts`):
- Default `graft build`: `parsed: 1 of 1` — the nested clone is skipped.
- `graft build --follow-nested-repos`: `parsed: 2 of 2` — the nested clone is indexed.
- Persisted + reversible (a later plain build honors the stored choice), and independent of `--follow-submodules`.

`graph-follow-nested-repos` + `graph-only-dir` 4/4; full suite 1045/1045. Merging.

PR #245, opened 2026-08-27
fix(check): mirror buildGraph's container tier so .vue never reads as removed

### LINKED ISSUE #236: `graft check` reports every `.vue` node as `removed` after a clean build — `checkGraph` has no container-tier branch
### Summary

`graft check` reports every `.vue` node as `removed` immediately after a clean `graft build`, and `graft build` can never repair it. `checkGraph` handles the depth tier and the breadth tier, but has no branch for the **container tier** that `buildGraph` uses to extract `.vue`, so container-tier nodes are extracted by the build and then invisible to the check.

This is the same class of bug the comment at `src/graph/check.ts:28-31` already anticipates for the breadth tier:

> else breadth-tier files (.rs, …) would re-extract as empty here and read as `removed` against a graph that built them, so `graft check` would never report OK.

That reasoning applies verbatim to the container tier; it just wasn't extended when `.vue` support landed.

### Environment

- graft 0.13.0, Node 22.15.0, Linux
- No LLM involved — structural tier only, `$0`, reproduces with no key configured

### Minimal repro

```bash
mkdir repro && cd repro && git init -q .
mkdir src && cat > src/Hello.vue <<'VUE'
<template><div @click="greet">{{ msg }}</div></template>
<script setup lang="ts">
const msg = 'hi';
function greet(): void {
  console.log(msg);
}
</script>
VUE
git add -A && git -c user.email=a@b -c user.name=t commit -qm init

graft build .
jq -r '.nodes[].id' graft/.graph/wiring.json
#   src/Hello.vue
#   src/Hello.vue#greet
jq -c '.meta.languages' graft/.graph/wiring.json
#   ["vue"]

graft check .
```

Output, with nothing changed between the build and the check:

```
graph check: STALE

removed (2):
  - src/Hello.vue
  - src/Hello.vue#greet

Run `graft build` to rebuild the structure, then commit graft/.
```

Running `graft build` again changes nothing — the check stays STALE forever. On our real monorepo this is 814 nodes across 137 `.vue` files, and the `removed` list contains nothing else.

### Cause

`buildGraph` has three extraction branches (`src/graph/build.ts`):

```ts
await warmGenericGrammars(...);
// Container tier (.vue and friends) loads its wrapper grammars the same way,
// for the same reason: extractContainer runs inside the sync loop below.
await warmContainerGrammars(...);
...
const lang = languageOf(f.abs);
const container = lang ? null : containerLangOf(f.abs);
const generic = lang || container ? null : genericLangOf(f.abs);
```

`checkGraph` has only two (`src/graph/check.ts:23-24`, `:88-101`) — there is no reference to `container` anywhere in the file:

```ts
import { extractFile, languageOf } from "./extract.js";
import { extractGeneric, genericLangOf, warmGenericGrammars } from "./generic.js";
...
const lang = languageOf(file);
const generic = lang ? null : genericLangOf(file);
...
const { nodes } = lang
  ? extractFile(relPosix(root, file), source, lang)
  : extractGeneric(relPosix(root, file), source, generic!.name);
```

For a `.vue` file `languageOf()` is null and `genericLangOf()` is also null (`.vue` is claimed only by the container registry, `src/graph/container.ts:31`), so line 100's `generic!.name` throws a `TypeError`. The `catch` on line 102 swallows it as a parse failure and the committed nodes fall through to `removed`.

### Impact

- `graft check` can never exit 0 on any repo containing a `.vue` file, so it is unusable as the CI drift gate it is built for.
- The remediation it prints (`Run graft build`) cannot fix it, which is exactly the failure mode the `src/graph/build.ts` comment about coarse mtime granularity warns against: *"`graft check` reports drift (it always re-hashes) and the `graft build` it tells you to run refuses to repair it — forever."*
- Retrieval is unaffected — the nodes really are in `wiring.json` — so this is a false positive, not missing data.

### Suggested fix

Mirror `buildGraph`'s branch in `checkGraph`: import `containerLangOf` / `warmContainerGrammars` / `extractContainer`, warm the container grammars alongside the generic ones, and add the `container` case ahead of the breadth tier. Replacing the `generic!` non-null assertion with a real guard would also turn any future third-tier omission into a visible error rather than a silent `removed`.

Happy to send a PR if that shape is agreeable.

anirudhkumar-nanonets:
Fixes #236.

`buildGraph` extracts in three tiers — depth (`languageOf`), container (`containerLangOf`, what claims `.vue`), breadth (`genericLangOf`). `checkGraph` only had two. For a `.vue` file both `languageOf` and `genericLangOf` return null, so the re-extraction hit `generic!.name` on a null, the surrounding `catch` swallowed the `TypeError` as a parse failure, and every committed `.vue` node fell through to `removed`.

That made it a false STALE no rebuild could clear: the `graft build` the report tells you to run rebuilds those nodes exactly as before, so `graft check` can never exit 0 on a repo containing a `.vue` file — unusable as the CI drift gate it exists to be.

## The change

- Import the container tier and warm its grammars alongside the generic ones, for the same reason the existing comment gives for the breadth tier.
- Add the `container` branch ahead of the breadth tier, matching `buildGraph`'s order.
- Replace the `generic!` non-null assertion with a real guard, placed **outside** the `try` so the catch cannot swallow it. It is unreachable today — `listSourceFiles` only yields files one of the three tiers claims — but the silent path is exactly how this hid, and a fourth tier added to `buildGraph` and forgotten here should be a loud error, not a permanent phantom `removed`.

## Verification

The issue's exact repro, before and after:

```
graph check: STALE          →   graph check: OK — the wiring graph is in sync with the code.

removed (2):
  - src/Hello.vue
  - src/Hello.vue#greet
```

New regression test in `test/container-extract.test.ts`, mirroring the breadth-tier one in `test/generic-extract.test.ts`: build a `.vue` repo, then assert `check` reports nothing removed or changed. Confirmed it fails without the `check.ts` change (`not ok 257`) and passes with it.

Full suite: **916 pass, 0 fail**.

### DIFF
--- a/src/graph/check.ts
+++ b/src/graph/check.ts
@@ -22,6 +22,7 @@ import { relPosix } from "../util/paths.js";
 import { contextDirFor } from "../context/node-file.js";
 import { extractFile, languageOf } from "./extract.js";
 import { extractGeneric, genericLangOf, warmGenericGrammars } from "./generic.js";
+import { containerLangOf, extractContainer, warmContainerGrammars } from "./container.js";
 import { listSourceFiles } from "./build.js";
 import { readGraph, wiringPath } from "./write.js";
 import { readSourceFile } from "../util/source.js";
@@ -49,10 +50,11 @@ export interface GraphCheckOptions {
   contextDir?: string;
 }
 
-// async: the breadth tier's WASM grammars load asynchronously and must be warmed
-// before the (synchronous) re-extraction below, exactly as buildGraph does — else
-// breadth-tier files (.rs, …) would re-extract as empty here and read as `removed`
-// against a graph that built them, so `graft check` would never report OK.
+// async: the breadth and container tiers' WASM grammars load asynchronously and
+// must be warmed before the (synchronous) re-extraction below, exactly as
+// buildGraph does — else those files (.rs, .vue, …) would re-extract as empty
+// here and read as `removed` against a graph that built them, so `graft check`
+// would never report OK.
 export async function checkGraph(
   dir: string,
   opts: GraphCheckOptions = {},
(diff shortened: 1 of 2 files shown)


anirudhkumar-nanonets:
Superseded by #248, which carries the identical fix plus a CHANGELOG entry linking #236. Closing as a duplicate — no code from this PR is lost.

PR #247, opened 2026-08-27
feat(hosts): graft uninstall, and init retracts the agents it isn't writing

anirudhkumar-nanonets:
`init` was additive. It wrote the files the *selected* agents needed and never looked at the rest, so a repo wired by an older version — or by the same version with different `--agents` — kept that run's files forever. Worse, `reconcileWiring` takes the union of the stamp and what's on disk, so those stray files were then **refreshed to the current version** on every upgrade: a `GEMINI.md` block from a run nobody remembers isn't stale, it's maintained.

This adds the missing half.

## `graft uninstall`

The inverse of `init`. Dry-run until `-y`.

```
$ graft uninstall
claude:
  - would remove: .claude/settings.json (statusline + hooks + allowlist + footer regex — deleted)
  - would remove: .claude/skills/graft/SKILL.md (graft skill — deleted)
  ~ would remove: .mcp.json (mcpServers.graft)

graph:
  - would remove: graft (local graph cache — deleted)

Dry run — nothing was touched. Re-run with -y to remove.
```

`--keep-cache` spares `graft/`; `--no-global` leaves `~/.codex` and `~/.gemini` alone.

## `init` converges

It now retracts every agent it is *not* about to write:

```
$ graft init --agents claude          # repo was wired for gemini + cursor + claude
- removed .cursor/rules/graft.mdc (graft-owned instruction file) — agent not selected
- removed GEMINI.md (fenced graft section) — agent not selected
✓ wrote .claude/settings.json …
```

`GEMINI.md` goes from *team notes + graft block* to *team notes*. The repo ends up matching this version's intent instead of the union of every version that ever ran.

Note the contract change: `graft init --agents cursor` on a Claude-wired repo now removes the Claude wiring. That is the declarative reading of the flag, but it is a behaviour change worth knowing about. `--no-agents` still means Claude-only.

## Two invariants

**Never delete what graft didn't write.** Inside a file the user owns, only the marker-fenced region and graft-named keys are touched. Verified preserved: a `sentry` server beside `graft` in `.mcp.json`, a hand-written `statusLine`, a foreign `PostToolUse` hook, `[mcp_servers.keepme]`, unrelated `.gitignore` lines, and prose either side of a fenced block.

**Never leave an empty shell.** A file holding nothing but graft's contribution is deleted, not truncated to `{}` or a blank document — an orphan file is the residue this exists to remove. Directories that empties are pruned. An unparseable config is reported and left byte-for-byte alone.

## Why there's no manifest

The obvious design is to record what each run wrote and diff against it. Two things in the history make that unnecessary: the section marker `<!-- graft:start -->` has exactly one value across every commit that ever defined it, and no other brand name was ever written into a host file. graft's artifacts identify themselves — `graft.mdc`, `mcpServers.graft`, `graft-hooks.cjs`, the fenced block — so retraction can find them without being told.

So the target list is *derived* from the same registries `init` writes through (`HOSTS`, `mcpTargets`, `hookTargets`, `claudeTargets`). A host added later is retractable for free. `LEGACY_TARGETS` is the one hand-kept list, and only a host **removed** from the registry needs an entry there — deleting from `HOSTS` without adding it is what strands a file in every existing repo. The doc comment says so.

## Not in this PR

- `upsertCodexToml` still returns `unchanged` the moment `[mcp_servers.graft]` exists, so a plain `init` won't refresh a stale Codex/Grok launch command. `uninstall && init` does.
- `mergeGraftSettings`' allow-list is still append-only. Retract prunes it, but a plain upgrade with claude selected excludes claude from retraction, so a renamed entry would survive.

## Test

19 new tests in `test/hosts-retract.test.ts`, covering each primitive, both invariants, purity of `planRetract`, idempotency of a second sweep, `exclude`, `--no-global`, and a full `init` → retract → nothing-left cycle. `npm test`: 966/966.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,24 @@
 
   The target list is derived from the same registries `init` writes through, so a
   host added later is retractable for free; only a host *removed* from the registry
-  needs a hand-written entry, and `LEGACY_TARGETS` says so.
+  needs a hand-written entry, and `LEGACY_TARGETS` says so. Exclusion is by path as
+  well as by host id: three hosts write `AGENTS.md`, and keeping any one of them has
+  to spare that block.
+
+### Fixed
+
+- **A stale `[mcp_servers.graft]` is now replaced instead of skipped.** The TOML
+  writer returned early the moment the header existed, which froze the launch
+  command at whatever the first `init` wrote — a repo wired when graft wasn't on
+  `PATH` kept the slow `npx` form forever, and no upgrade could correct it. Codex
+  and Grok both went through that path. Foreign tables are untouched either way.
+
+- **`.claude/settings.json` no longer accumulates graft's own entries.** The
+  allowlist and `footerLinksRegexes` were append-only, so a renamed invocation form
+  stayed in the user's settings beside its replacement with nothing able to remove
+  it. Both now drop graft's prior entries before adding the current set, the same
+  way the hooks merge already did. Scoped to the forms graft is actually invoked as,
+  so a hand-written `Bash(graft-mytool:*)` survives.
 
 - **`graft blast` suggests who to tag.** The comment already named the areas a
   diff changes and the areas it can affect; it now names the people behind them,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,24 @@
 
 ### Added
 
+- **`graft init` converges instead of accumulating, and `graft uninstall` removes
+  graft entirely.** `init` wrote the files the selected agents needed and never
+  looked at the rest, so a repo wired by an older version — or by the same version
+  with different `--agents` — kept that run's files forever, and the session-start
+  refresh then kept them *up to date*. `init` now retracts every agent it isn't
+  about to write, and `graft uninstall` retracts the lot.
+
+  Only graft's own contribution is touched: inside a shared file just the
+  marker-fenced block, inside a config just the `graft` key — foreign MCP servers,
+  hooks, statuslines and ignore entries survive byte for byte. A file left holding
+  nothing is deleted rather than truncated to an empty shell, and the directories
+  that empties are pruned. An unparseable config is reported and left alone.
+  `uninstall` is dry-run until `-y`.
+
+  The target list is derived from the same registries `init` writes through, so a
+  host added later is retractable for free; only a host *removed* from the registry
+  needs a hand-written entry, and `LEGACY_TARGETS` says so.
+
 - **`graft blast` suggests who to tag.** The comment already named the areas a
   diff changes and the areas it can affect; it now names the people behind them,
   read from git history with no API call and no config file. One `git log` per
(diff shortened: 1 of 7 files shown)


anirudhkumar-nanonets:
Follow-up pushed: the two append-only writers now converge too, and I tested the whole thing against **real published versions** rather than only unit fixtures.

## The two fixes

**`upsertCodexToml` replaces its own stale section.** It used to return `unchanged` the moment `[mcp_servers.graft]` existed, which froze the launch command at whatever the first `init` wrote — a repo wired when graft wasn't on `PATH` kept the slow `npx` form forever, and no upgrade could correct it. The strip helper is now exported from `mcp-config.ts` and `retract.ts` imports it, so writer and retractor cannot disagree about what "graft's section" is.

**`mergeGraftSettings` stops stacking.** `permissions.allow` and `footerLinksRegexes` were append-only, so a renamed invocation form stayed beside its replacement with nothing able to remove it. Both now filter graft's prior entries before adding the current set — the shape the hooks merge already used. `isGraftAllowEntry` is scoped to the forms graft is actually invoked as, so a hand-written `Bash(graft-mytool:*)` survives; `retract.ts` imports that too instead of keeping its own regex.

## A bug the cross-version run caught

Exclusion was by host **id**. Three registry hosts write `AGENTS.md` (`agents`, `hermes`, `antigravity`), so `init --agents claude agents` retracted the block on behalf of `hermes` — a host nobody selected — and printed a misleading `removed AGENTS.md — agent not selected`. The end state was right only because `init` rewrote the block a moment later.

Exclusion is now by **path as well as id**, and targets are deduped by path so three hosts naming one file produce one retraction instead of three. Two regression tests cover it. This is exactly the class of bug the unit fixtures missed and a real upgrade surfaced.

## Cross-version testing

Installed `@nanonets/graft@0.7.1`, `@0.9.1` and `@0.12.1` from npm, ran each one's real `graft init` in a fixture repo carrying a user's own content, then ran this build's `init` over the top. Isolated `HOME`, so `~/.codex` writes were exercised for real.

Two scenarios per version:

- **same agent set** — asserts no duplicated blocks, sections or hook entries; every JSON graft touched still parses; the skill is refreshed to current; a second `init` is byte-for-byte a no-op; and `uninstall` returns the repo to exactly the user's own keys.
- **shrinking agent set** — old version wires claude + cursor + agents + kiro + windsurf, this build runs `--agents claude agents`. Asserts the four dropped agents are retracted (with empty dirs pruned), the two kept ones survive, and a deliberately planted `command = "STALE-BINARY"` in the Codex TOML is refreshed while `[mcp_servers.keepme]` is untouched.

All six runs pass. Preserved throughout: team prose in `AGENTS.md`, a `sentry` server beside `graft` in `.mcp.json`, the user's `model` and `Bash(ls:*)` in `settings.json`, and a foreign Codex table.

Unit suite: 972 passing, 25 of them in `test/hosts-retract.test.ts`.

PR #254, opened 2026-08-28
fix(deep): don't abort a 96% meaning pass the same way as a quota error (#235)

### LINKED ISSUE #235: 0.13.0: deep pass hard-fails (circuit breaker) on files where 0.10.1 published — DeepSeek-V4/Together crux returns "no usable symbol summaries"
### Summary

Upgrading `@nanonets/graft` 0.10.1 → 0.13.0 turns a previously-successful `graft build --deep` into a deterministic build failure on the **same repo, model, and key**. This appears to be the intended behavior change from #177 (`fix(deep): treat empty meaning replies as failures, not permanent pending`, commit `bd6bf6b2df`) doing exactly what it says — but combined with the `LlmFailureGate` circuit breaker and no-partial-by-default, it converts a long-standing *silent* model limitation into a fully blocked build, with no per-file diagnostic to act on. Filing because the 0.13 experience for a model that 0.10.1 handled "fine" is a hard stop, and the underlying empty-crux path (#129) is still not recovered for this provider.

### Environment

- graft 0.10.1 (works) vs 0.13.0 (fails), Node 22, macOS
- Provider: `GRAFT_PROVIDER=openai`, `GRAFT_BASE_URL=https://api.together.ai/v1`
- Model: `deepseek-ai/DeepSeek-V4-Flash-0731`
- Repo: ~336 source files (TS/TSX/Python/JS), ~2133 symbols

### What happens

**0.10.1 (same repo/model/key, minutes earlier):** `--deep` completes and publishes. 443 cards; reports "89 pending" (unattempted/deferred) but exits 0 and produces a usable deep artifact.

**0.13.0, run 1:** `20 file(s) failed to summarize, 14 never attempted. meaning coverage: 2047/2133 symbols (96%).` Circuit breaker: `8 files in a row failed, so the pass stopped.` Per-file error: `model returned no usable symbol summaries`.

**0.13.0, run 2 (resume from cache):** `9 file(s) failed to summarize, 26 never attempted.` Coverage stuck at the **exact same 2047/2133**. Same recurring files fail (e.g. `app/components/Header.tsx`, `app/components/VehicleCard.tsx`, `app/components/AuthForm.tsx`, `app/components/LeadForm.tsx`, and several large domain modules).

Structural build + `graft check --json` on 0.13.0 is **clean** (`graph.ok: true`, added/changed/removed/stale all 0) — only the deep/meaning tier is affected.

### Analysis

- The `~89 pending` on 0.10.1 and the `86 missing` (2133 − 2047) on 0.13.0 are the same set of files the model can't summarize. 0.10.1 deferred them silently and published; 0.13.0 (correctly, per #127/#172) surfaces them, then the circuit breaker + default-no-partial block the whole artifact.
- The crux request (schema, system prompt, `temperature`, `max_tokens: 8192`, forced `tool_choice`) is unchanged between 0.10.1 and 0.13.0, so this is not a stricter validation of valid output — it's that these replies were always empty/unusable and are now fatal.
- 0.13 added `recoverArgsFromContent()` (great, addresses #129's ask), but for DeepSeek-V4-Flash on Together the recovery still yields nothing for these files, so `argsFromResponse` → `parseResults` → empty.

### Questions / requests

1. The failing files are the largest / most symbol-dense ones. Could the crux call be hitting `max_tokens: 8192` and finishing with `finish_reason: "length"`, truncating the tool-call arguments? If so, per-file `max_tokens` scaling (or batching/symbol chunking) would fix it rather than failing the file.
2. Can the per-file error include the observed `finish_reason` and whether the miss was empty-`toolCalls` vs unparseable-args vs truncation? Today `model returned no usable symbol summaries` doesn't say which of #129's paths hit, so users can't tell a model quirk from a token cap.
3. Consider not letting `LlmFailureGate` (`MAX_CONSECUTIVE_FAILURES = 5`) abort a publish for *content-quality* misses the way it does for terminal quota/auth errors — or make `--allow-partial` the default with a loud summary, so a model that produces a 96% tier still ships (as 0.10.1 did).

Happy to run a minimal repro that replays the exact crux request (same schema / prompt / temperature / max_tokens / tool_choice) against Together and report the raw `finish_reason` / `tool_calls` / `content` for one failing file if useful.

Frankie-Xu:
## Summary
- After #177, empty/unusable crux replies correctly fail the file instead of being cached as `ready`. Combined with `LlmFailureGate`'s consecutive-failure cutoff, that turned a ~96% meaning pass (the same files 0.10.1 used to defer as pending) into a hard abort: remaining files were skipped, resume died on the same set, and `graft build --deep` exited like a quota error.
- **A — diagnostics:** per-file errors now include a miss class (`empty-toolCalls` / `unparseable` / `truncated` / `empty-parsed`) and the provider `finish_reason`, so truncation vs empty vs parse-miss is one line.
- **B — gate:** content-quality misses still count as `failedFiles` (exit 1 unless `--allow-partial`) and are **not** written as successful cache. They no longer increment the consecutive-failure cutoff, so the rest of the pass is attempted. Quota/auth still stop immediately (#127).
- **Not rolled back:** #177 remains — an empty crux is not a cache hit.
- **Not in this PR:** per-file `max_tokens` scaling / symbol chunking (issue request 1). Maintainer diagnosis confirmed truncation on dense files, but this window has no Together key and no raw `finish_reason=length` dump from the reporter's files, so a token policy would be a guess. The new miss class will make that follow-up obvious when it happens.

## Test plan
- [x] Consecutive empty `toolCalls` across 12 files: no `fatal`, every file attempted, nodes stay `pending` (not cached ready)
- [x] Consecutive 401: still stops on the first file (#127)
- [x] Per-file errors match `empty-toolCalls` / `truncated` / `unparseable` / `empty-parsed` plus `finish_reason=…`
- [x] Blank parsed summaries stay pending and retry on the next `--deep` (#177 / #172)
- [x] Existing `graph-enrich-failures`, `graph-enrich-pending`, `llm-ops`, `cli-deep-failure` tests
- [x] `npx tsc -p tsconfig.json --noEmit` and `npm test` — 1066 pass, 0 fail (Node 24; `npm install` needed `--ignore-scripts` here because tree-sitter-kotlin's native rebuild cannot find C++ headers on this machine)
- [ ] No Together/OpenAI key in this window — did not replay a live dense TSX file against DeepSeek-V4-Flash

Closes #235

### DIFF
--- a/src/ai/crux.ts
+++ b/src/ai/crux.ts
@@ -14,7 +14,7 @@
  * sees each symbol's neighbours, which sharpens the summaries. Line numbers are
  * consumed once, at write time, to slice the crux text verbatim from source.
  */
-import type { ChatModel } from "./llm/types.js";
+import type { ChatModel, ChatResponse } from "./llm/types.js";
 import { recoverToolArgsFromContent, warnToolChoiceIgnored } from "./llm/recover-tool.js";
 import type { Kind } from "../graph/types.js";
 
@@ -42,6 +42,40 @@ export interface NodeCrux {
 
 export interface CruxSummarizer {
   describeFile(input: FileCruxInput): Promise<NodeCrux[]>;
+  /** Set by {@link ChatCruxSummarizer} after each call; optional on fakes. */
+  lastMiss?: CruxMiss | null;
+}
+
+/** Why a crux call produced no usable summaries (#235). */
+export type CruxMissKind = "empty-toolCalls" | "unparseable" | "truncated" | "empty-parsed";
+
+export interface CruxMiss {
+  kind: CruxMissKind;
+  finishReason: string | null;
+}
+
+function isTruncatedStop(reason: string | null): boolean {
+  if (!reason) return false;
+  const r = reason.toLowerCase();
+  return r === "length" || r === "max_tokens";
+}
+
+/** Classify an empty/unusable crux reply. `null` means at least one usable summary. */
+export function classifyCruxMiss(res: ChatResponse, parsed: NodeCrux[]): CruxMiss | null {
+  const finishReason = res.stopReason;
+  if (parsed.some((p) => p.summary.trim())) return null;
+  if (isTruncatedStop(finishReason)) return { kind: "truncated", finishReason };
+  if (parsed.length > 0) return { kind: "empty-parsed", finishReason };
+  const emptyTools = res.toolCalls.length === 0;
+  const emptyText = !res.text?.trim();
+  if (emptyTools && emptyText) return { kind: "empty-toolCalls", finishReason };
+  return { kind: "unparseable", finishReason };
+}
+
+/** Per-file error text: miss class + the provider's finish_reason (#235). */
+export function formatCruxMiss(kind: CruxMissKind, finishReason: string | null): string {
+  const fr = finishReason == null || finishReason === "" ? "null" : finishReason;
+  return `model returned no usable symbol summaries [${kind}, finish_reason=${fr}]`;
 }
 
 const SYSTEM_PROMPT = `You explain code definitions for a code graph that helps engineers navigate a codebase.
@@ -139,9 +173,12 @@ function argsFromResponse(res: { text: string; toolCalls: { name: string; args: 
 
 /** Crux summarizer backed by any {@link ChatModel} via forced tool calling. */
 export class ChatCruxSummarizer implements CruxSummarizer {
+  lastMiss: CruxMiss | null = null;
+
   constructor(private model: ChatModel) {}
 
   async describeFile(input: FileCruxInput): Promise<NodeCrux[]> {
+    this.lastMiss = null;
     if (input.nodes.length === 0) return [];
     const res = await this.model.create({
       temperature: 0,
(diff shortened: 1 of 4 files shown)


shhdwi:
This is exactly the right fix for #235 — thank you for taking it on, and for the careful separation of concerns. Verified locally on top of current `main` (clean merge; `npm run build` + `tsc --noEmit` clean):

- **Content-quality misses no longer abort the build.** Base main: 12 empty-crux files → `fatal="5 files in a row failed…"`, 7 skipped, whole pass aborts. This PR: same 12 → `fatal=undefined`, all attempted, 0 skipped, nodes stay `pending` (uncached, so a retry is free). A ~96% pass now publishes instead of hard-failing.
- **Quota/auth stays terminal.** A consecutive 401 still trips `fatal` on the first file — `terminalReason()` fires even under the `{quality:true}` flag. That's the important half: the breaker still stops a spent key immediately.
- Nice touch classifying the miss (`empty-toolCalls`/`unparseable`/`truncated`/`empty-parsed`) + surfacing `finish_reason`, which directly answers the reporter's truncation question.

New `graph-enrich-quality.test.ts` (7 cases) + the existing enrich/failure/deep suites: 28/0. Fixes #235. Merging — this unblocks deep builds on DeepSeek/Together/o-series for everyone. 🙌

PR #262, opened 2026-08-29
fix(deep): honor --only-dir in the concept pass (#252)

### LINKED ISSUE #252: `graft build --deep` ignores `--only-dir`: concept pass walks the whole repo (out-of-scope summaries + full re-synthesis on any new file)
## Summary

`graft build --deep` ignores the `--only-dir` whitelist. The Tier-1 wiring pass honors it (`graph/build.js` → `filterByOnlyDirs`), but the Tier-2 concept pass (`context/build.js` → `buildContext`) walks the whole repo, so out-of-scope files get summarized and pulled into concept synthesis.

**Version:** graft 0.15.0 (npm `@nanonets/graft`), Node 23.7, macOS.

## Steps to reproduce

```bash
graft build --only-dir src --only-dir tools          # wiring: N files, fingerprint records onlyDirs
graft build --deep --only-dir src --only-dir tools   # watch the progress counter
```

Observed: `reading concepts i/582` — the denominator is the full repo walk (here: code files that live under `docs/**` as evidence copies, which are deliberately excluded via `--only-dir`).
Wiring for the same flags: `✓ wiring: … 493 file`.

Expected: the deep pass enumerates the same file set as the wiring pass (plus whatever `CODE_EXTENSIONS` adds beyond the parser set, e.g. `.sh`), i.e. `reading concepts i/525` here — never files outside the whitelist.

## Root cause

`dist/cli.js` resolves `onlyDirs` from `--only-dir` and passes it only to `engine.graph(dir, { …, onlyDirs })`. The `--deep` branch calls `engine.init(dir, { extensions, onProgress })` without it, `Graft.init` builds the `buildContext` opts without it, and `context/build.js#buildContext` does:

```js
const files = walkDir(root, readIncludeDirs(root), { followSubmodules, followNestedRepos })
    .filter((f) => exts.some((e) => f.toLowerCase().endsWith(e)))
    .filter((f) => !f.startsWith(outDir));
```

— `readIncludeDirs` is honored, `onlyDirs` never reaches this function, and unlike `probeDrift` (which reads `fp.onlyDirs` from the fingerprint) there is no fallback to the persisted whitelist either.

## Impact

1. Concept nodes are synthesized from files the user explicitly excluded (in our case: copies of backend source kept as evidence under `docs/`), which pollutes `graft ask` results with out-of-scope "systems".
2. Because `batchBySize` runs over the full sorted file list, **any** new out-of-scope file shifts the batch boundaries → every synthesis batch key changes → the whole concept layer is re-synthesized instead of the incremental update the cache is designed for (for us: ~40 min on a local model instead of a few minutes, after a merge that only added a handful of `docs/**/*.java` files).

## Suggested fix

Three small changes (we run them as a local patch on `dist/` and can turn them into a PR if you prefer):

1. `cli.js`: pass `onlyDirs` to `engine.init(dir, { extensions, onlyDirs, onProgress })`.
2. `engine.js` `Graft.init`: forward `onlyDirs: opts.onlyDirs` into `buildContext`.
3. `context/build.js` `buildContext`: apply the same filter as the wiring pass, falling back to the fingerprint when the CLI didn't pass it (mirrors `probeDrift`):

```js
import { filterByOnlyDirs } from "../graph/source-files.js";
import { readFingerprint } from "../graph/fingerprint.js";
// …
const onlyDirsList = opts.onlyDirs?.length ? opts.onlyDirs : (readFingerprint(outDir)?.onlyDirs ?? []);
const files = filterByOnlyDirs(walked, root, onlyDirsList.length ? new Set(onlyDirsList) : undefined);
```

Verified locally: with the patch the deep pass reports `reading concepts i/525` both with the flags and without them (fingerprint fallback); `525 = 493 wiring files + 32 .sh` — the `.sh` delta is expected (`CODE_EXTENSIONS` vs parser set) and not part of this report.

Related: `graft check` / `probeDrift` already read `fp.onlyDirs`, so the whitelist is available wherever the graph dir is.

Frankie-Xu:
## Summary
- `graft build --deep` now applies the same `--only-dir` whitelist as the wiring walk (`filterByOnlyDirs`), so the concept pass no longer summarizes or synthesizes files outside the prefixes.
- `checkContext` reads the same file set (CLI/API, else the fingerprint, matching `checkGraph` / `probeDrift`) so `graft check` does not report excluded files as coverage.
- Wiring behavior is unchanged; `--only-dir` prefix / repeatable semantics are the same as #149.

Closes #252

## Test plan
- [ ] `node --import tsx --test test/context-only-dir.test.ts` — concept summaries, multi `--only-dir`, fingerprint fallback, checkContext, `Graft.init`, wiring still scoped
- [ ] `node --import tsx --test test/graph-only-dir.test.ts` — existing whitelist + fingerprint probe still clean
- [ ] `npm test` — full suite (1065 passing locally)

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -331,7 +331,8 @@ program
   )
   .option(
     "--only-dir <path>",
-    "only index files under this repo-relative path — repeatable; persisted, so a later build " +
+    "only index files under this repo-relative path — repeatable; the wiring walk and the --deep " +
+      "concept pass both honor it. Recorded in the graph fingerprint so a later build " +
       "(and the hooks/refresh path) walks the same set; everything outside the list is skipped",
     (val: string, prev: string[]) => [...prev, val],
     [] as string[],
@@ -462,6 +463,7 @@ program
     if (deep) {
       const c = await engine.init(dir, {
         extensions: opts.extensions,
+        onlyDirs,
         onProgress: ({ phase, index, total, file }) =>
           process.stderr.write(
             `\r${phase === "summarize" ? "reading" : "writing"} concepts ${index + 1}/${total}: ${file.slice(0, 40).padEnd(40)}`,
--- a/src/context/build.ts
+++ b/src/context/build.ts
@@ -14,6 +14,8 @@
 import { readFileSync, writeFileSync, renameSync, mkdirSync, existsSync } from "node:fs";
 import { join, resolve } from "node:path";
 import { walkDir } from "../ingest/fs.js";
+import { filterByOnlyDirs } from "../graph/source-files.js";
+import { readFingerprint } from "../graph/fingerprint.js";
 import { contentHash } from "../util/id.js";
 import { relPosix } from "../util/paths.js";
 import { readSourceFile } from "../util/source.js";
@@ -60,6 +62,10 @@ export interface BuildOptions {
   contextDir?: string;
   /** Extensions to treat as code. Default: {@link CODE_EXTENSIONS}. */
   extensions?: string[];
+  /** Repo-relative directory prefixes to limit the concept pass (`--only-dir`).
+   * Same prefix semantics as the wiring walk. When omitted, falls back to the
+   * whitelist recorded in the graph fingerprint (mirrors `checkGraph`). */
+  onlyDirs?: string[];
   /** Human label for the model, recorded in the manifest (e.g. "openrouter:openai/gpt-4o-mini"). */
   model: string;
   summarizer: Summarizer;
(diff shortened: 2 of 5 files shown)


shhdwi:
Verified locally on top of current `main` (clean merge incl. #254; build clean).

Reproduced #252 with an offline recording summarizer: `graft build --deep --only-dir src` on base main summarized the WHOLE repo (`docs/c.ts`, `src/a.ts`, `tools/b.ts`) — out-of-scope work + full re-synthesis. With this PR only `src/a.ts` is summarized/synthesized. Nice that `listContextFiles`/`filterByOnlyDirs` reads the CLI value else the graph fingerprint (mirroring `probeDrift`) and is shared by `checkContext`, so build and check can't disagree. `context-only-dir` 6/6, `graph-only-dir` 2/2, `context`+`check` 30/30. Fixes #252. Merging.

PR #263, opened 2026-08-29
fix(check): root source file cards must not collide with concept slugs (#261)

### LINKED ISSUE #261: A source file at the repository root makes its card collide with concept nodes — permanent "node file not in manifest"
**Version:** 0.13.0 · node v24.19.0 · Linux · repo: PHP/Laravel with `server.php` at the repo root

## Symptom

After a fully successful `graft build --deep` (meaning tier 100%, 5997/5997 nodes ready),
`graft check` still reports the context as stale, on every run, forever:

```
graft check: STALE

index mismatch (1):
  ! server: node file not in manifest

Run `graft build --deep` to regenerate, then commit graft/.

graph check: OK — the wiring graph is in sync with the code.
```

Rebuilding does not clear it, because the build is what creates the file.

## Root cause

Per-file cards mirror the source tree under the context dir, so a source file at the
**repository root** — here `server.php` — writes its card to `graft/server.md`, which is
also where concept nodes live. `readNodes` (`dist/context/node-file.js:207`) treats every
top-level `*.md` except `INDEX.md` as a concept node, and invents a slug from the filename
when the frontmatter has none (`dist/context/node-file.js:216`) — so the card silently becomes
a "concept" that can never appear in `manifest.nodes`, and `check` flags it at
`dist/context/check.js:86`.

In this repo: 143 top-level `*.md` files, of which exactly 142 carry a `slug:` in their
frontmatter — those 142 are the real concepts and match `manifest.json`. The one without is
the file card, whose frontmatter is only `covers: []`.

`readNodes` has three other callers, so the effect is wider than the exit code:

- `dist/graph/cards.js:178` — the card ends up listed under `## Concepts` in `INDEX.md`
- `dist/graph/cards.js:44` — it participates in the `covers` pass
- `dist/blast/modules.js:24` — `blast` can name it as an affected area

Worth noting for anyone reading the symptom: this does **not** drive the sync loop, though it
looks like it might. `indexDrift` is not read outside `check.ts`, and `checkStaleCount`
(`dist/claude/hooks.js:159`) counts only `graph.changed/added/removed`. The cost is a
permanently non-zero `graft check` (so it cannot be used as a CI gate) plus the pollution
above.

## Reproduction

Any repo with an indexed source file directly at its root. The card is written by a plain
build; the mismatch surfaces once a manifest exists, i.e. after a `--deep` run:

```
$ echo 'function hi(){}' > root-thing.js
$ echo 'export function deep(){}' > lib/deep.js
$ graft build .
$ ls graft/*.md
graft/INDEX.md   graft/root-thing.md      # <- the root file's card, beside the concepts

$ graft build --deep .
$ graft check
index mismatch (1):
  ! root-thing: node file not in manifest
```

(The first three steps are verified on a throwaway repo with 0.13.0; the `check` output above
is the one this repo produced for its own root-level `ws-server.php`.)

## Suggested fix

Distinguish concept nodes from file cards rather than assuming every top-level `.md` is a
concept. Skipping entries whose frontmatter carries no `slug` was a one-line fix here and
cleared `check`, `INDEX.md` and `blast` at once, while leaving the card itself in place and
regenerating normally. Writing root-level file cards into a subdirectory instead would also
work, at the cost of moving existing files.

Frankie-Xu:
## Summary
- #215 / #213 taught `readNodes()` to skip a top-level `graft/<stem>.md` when that stem is a **root source in `manifest.files`** and the file has no concept `slug`. That is the common case (`main.ts` → `graft/main.md`) and is unchanged.
- #261 is the other side of the same namespace collision, which #215's skip is too narrow to cover:
  1. **Stem vs concept slug.** A synthesized concept named `Server` occupies `graft/server.md`. The wiring card for root `server.php` used to overwrite it; `writeCovers` then stamped `covers: []` onto the remnant, and check reported a missing node (or, on 0.13.0, `server: node file not in manifest`). The card now relocates to `graft/_root/server.md` only when that filename is already a concept — default layout is unchanged, existing `graft/` trees keep working, and a real `server` concept is not deleted.
  2. **Skip keyed only on `manifest.files`.** The graph can still write a root card for a source the concept pipeline never recorded (e.g. `.lua` vs context `CODE_EXTENSIONS`). `readNodes()` now also recognizes a file card by wiring-card heading (`# server.php`) or `covers:`-only frontmatter, so those extras are not treated as concept nodes. A hand-dropped `graft/notes.md` still fails `indexDrift`.
- `writeCovers` no longer stamps `covers: []` onto slug-less top-level files (that was what made a file card look like a concept in the reporter's 0.13.0 dump).

`--deep` was not run against a live LLM (no key in this environment). Reproduction is the same wiring/cards + fake-synthesizer path #215 used: `buildContext` then `buildGraph`, then `checkContext`.

Closes #261

## Test plan
- [x] Root `server.php` + nested `lib/deep.php`: `checkContext` is clean (`indexDrift: []`)
- [x] Concept slug `server` + root `server.php`: `graft/server.md` still has `slug: server`; file card is `graft/_root/server.md`; check is clean
- [x] Root file card whose source is **not** in `manifest.files` (`.lua`): check is clean
- [x] Hand-dropped `graft/notes.md` is still `indexDrift` (collision case and #213 case)
- [x] Nested cards stay in subdirs; #213 `main.ts` fixture still green
- [x] `writeCovers` does not add `covers` to a root file card
- [x] `npm run build && npm test` — 1063 pass, 0 fail

### DIFF
--- a/src/context/node-file.ts
+++ b/src/context/node-file.ts
@@ -266,6 +266,10 @@ export interface ParsedNode {
  * for source files in the repo root. Those cards share a directory with concept
  * nodes (`graft/<stem>.md`) but are recorded in `manifest.files`, not
  * `manifest.nodes`. Nested file cards live in subdirs and are never scanned here.
+ *
+ * Incomplete on its own (#261): the graph can write a root card for a source
+ * that the concept pipeline never recorded (different extension lists), so the
+ * stem is absent from `manifest.files`. Combined with isRootFileCard().
  */
 function rootFileCardStems(dir: string): Set<string> {
   const manifest = readManifest(dir);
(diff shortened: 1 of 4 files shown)


shhdwi:
Verified locally on top of current `main` (clean; build clean). Confirmed this is **distinct from #215/#213**, not a duplicate — it closes two residual holes:
1. **Slug collision:** a concept named `Server` owns `graft/server.md`, and a root `server.php` card used to clobber it; `cardPathFor` now relocates the card to `graft/_root/server.md` when the stem is already a concept.
2. **Card for a source not in `manifest.files`** (e.g. a graph-indexed `.lua`): `readNodes` now recognizes a file card by `covers:`-only frontmatter or a `# path.ext` heading, and `writeCovers` no longer stamps `covers: []` on slug-less top-level files.

Before/after: against base main the `#215` case was already clean, but the slug-collision test overwrote the concept and the `.lua` test threw `indexDrift: ["server: node file not in manifest"]` — exactly #261. All 3 new tests pass on this PR; `context` 33/33, `covers` 4/4. Fixes #261. Merging — thanks for chasing the residual cases!

PR #265, opened 2026-08-29
feat(metrics): score Cursor sessions like Claude Code

afeddersen:
Add per-session usage accounting for Cursor — graft vs source reads and token savings — via repo-local .cursor hooks, a `graft stats` readout, and host-attributed `session_summary` telemetry.
- consolidate the Codex/Cursor/MCP config writers onto a shared ConfigWrite type and a readJsonObject load-or-skip helper (src/hosts/config-write.ts)
- force-close the just-ended conversation on Cursor's sessionEnd instead of waiting for the idle sweep, and stamp the recording host on each session
- anchor graft MCP-tool detection to the real tool-name list rather than a loose "graft" substring (src/mcp/tool-names.ts)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ transcripts/
 # and never part of a PR diff (the hook helpers/skills alongside them stay tracked).
 .claude/settings.json
 .claude/settings.local.json
+
+# Cursor project config — written per machine by `graft init` (hooks, MCP, rule),
+# a personal setup like the Claude settings above, not a shared artifact.
+.cursor/
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -30,9 +30,23 @@ if (files.length === 0) {
   process.exit(1);
 }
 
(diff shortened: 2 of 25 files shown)


shhdwi:
Verified locally on top of current `main` (clean merge; build clean; 95/95 across session-metrics, cursor-hooks, claude-hooks, telemetry-sessions).

Real session scoring for Cursor, mirroring Claude Code: `cursor-hooks.ts` writes `.cursor/hooks.json` wiring `postToolUse`/`afterMCPExecution`/`sessionEnd` to shared classify+record in `session-metrics.ts` (previously `graftReads`/`sourceReads` were never incremented, so Cursor `session_summary` shipped 0/0). Live demo through the hook: before → "no session recorded"; after a Cursor Read + a graft MCP call + a `graft ask`: `graft reads 2, source reads 1, 67% graft, ~6,600 saved`, attributed to host `cursor`. MCP double-count is correctly guarded, and the new `graft stats` reads the local mix. Nice, complete piece of work — thanks! Merging.

PR #187, opened 2026-08-24
feat(graph): add Lua language support

CroissantEdit:
## Summary

- Add Lua (`.lua`) to the breadth-tier generic extractor using the existing `tree-sitter-wasm` Lua grammar
- Register Lua in `GENERIC_LANGS`
- Add `lua.scm` queries for function and method definitions and call references
- Cover local functions, assigned functions, function-valued table fields, and `:` method syntax
- Update the supported-language documentation

## Test results

`npm run build` passes.

`npm test`: 832 passed, 5 skipped, 0 failed.

PR #266, opened 2026-08-29
feat(graph): add dedicated Luau support

ansien:
## Summary

- route `.luau` through the dedicated `tree-sitter-luau` grammar while keeping `.lua` on the Lua grammar
- extract Luau type aliases, functions, methods, and calls without relying on Lua parser error recovery
- ship the reviewed 94 KB Luau grammar WASM with its source hash and MIT license, without adding a runtime dependency
- cover casts, generics, compound assignment, if expressions, string interpolation, and `continue`

## Testing

- clean `npm ci`
- `npm run build`
- focused extension/parser suite: 24 passed
- full suite: 1,056 passed, 5 skipped
- `npm pack --dry-run` includes the Luau grammar, query, provenance, and license


shhdwi:
Really nice work — verified locally on top of current `main` and the extraction is solid:
- Breadth-tier `.luau` with a dedicated `luau.scm` that correctly adds Luau's `type_definition` handling (distinct from the existing Lua #187, not a duplicate).
- Behavioral: a real `.luau` file → functions + a type + a method + resolved call edges (`magnitude→add`, `move→add`), no garbage; `skeleton` shows the Luau type syntax. Tests 24/24 (generic-extract + supported-extensions), build clean.
- The vendoring is done carefully: MIT LICENSE committed alongside, `grammars/README.md` records provenance (tree-sitter-luau 1.2.0) + SHA-256, which I re-verified matches.

The one thing holding a merge is a **maintainer policy call, not your code**: this would be the repo's *first committed `.wasm` grammar* — every other language loads from the `tree-sitter-wasm` npm bundle, and graft has deliberately kept binaries out of the tree. Since `tree-sitter-wasm` doesn't ship Luau, vendoring is the only path today, so this is a real precedent to accept consciously.

@shrishdwi — flagging for your call: accept the vendored-`.wasm` precedent (the PR is otherwise merge-ready), or ask for the grammar to be published to npm / added to the `tree-sitter-wasm` bundle first? Either way the extraction work here is great and I'd love to see Luau land.

ansien:
Merged latest `main` in and resolved the conflict in `src/graph/generic.ts` (kept the new `swapGrammarForTest` seam alongside the Luau loader). PR is conflict-free again; generic-extract + supported-extensions pass (26/26).

@shhdwi @anirudhkumar-nanonets — could one of you take another look? Per the earlier review, the only open item is the policy call on committing the vendored `tree-sitter-luau.wasm` (the earlier tag `@shrishdwi` doesn't resolve to a GitHub user, so I don't think it reached anyone). Happy to go either way: keep it vendored with the LICENSE + SHA-256 provenance, or hold until the grammar is available from npm.

PR #220, opened 2026-08-24
feat: add LiteLLM provider with /v1/models discovery

prodmanpd:
## Summary

Adds **LiteLLM** as a first class provider (`provider: "litellm"`), so one endpoint reaches **100+ models** (OpenAI, Anthropic, Gemini, Bedrock, Vertex, Azure, Groq, and more) through a self hosted LiteLLM proxy.

A LiteLLM proxy speaks the OpenAI compatible wire format, so, per this repo's own design ("adding a wire format is one new adapter"), the adapter **reuses `OpenAIChatModel`'s translation** and only changes the defaults. The value over `provider: "openai"` plus a manual `baseUrl` is a self describing provider:
* a sensible default proxy address (`http://localhost:4000`), and
* **`/v1/models` auto discovery** (`listLiteLLMModels`) so callers can enumerate whatever models the proxy is configured to serve instead of hardcoding a model id.

## Changes

* `src/ai/llm/litellm.ts` (new): `LiteLLMChatModel extends OpenAIChatModel` (default proxy base URL and `litellm:<model>` label) and `listLiteLLMModels()` which calls `GET /v1/models` and returns the available model ids.
* `src/ai/llm/factory.ts`: `ProviderKind` gains `"litellm"`; `createChatModel` builds a `LiteLLMChatModel`.
* `src/ai/providers.ts`: default model for the `litellm` provider.
* `test/litellm-adapter.test.ts`

## Tests

**Types**: `npx tsc -p tsconfig.json --noEmit` is clean.

**Tests**: `npm test` (repo runner): **840 passed, 0 failed** (4 new litellm tests plus no regressions).
```
test/litellm-adapter.test.ts
  ok litellm: reuses OpenAI-compatible translation and labels as litellm
  ok litellm: /v1/models auto-discovery returns the proxy's model ids
  ok litellm: factory builds a LiteLLMChatModel for provider 'litellm'
  ok litellm: exposes a default proxy base URL
```
The new tests assert: the adapter forwards the provider prefixed model and temperature and reuses the OpenAI translation; `listLiteLLMModels` maps `/v1/models` to ids; and the factory wires `provider: "litellm"`.

## Usage

```ts
import { createChatModel } from "./ai/llm/factory.js";
import { listLiteLLMModels } from "./ai/llm/litellm.js";

// Chat through a LiteLLM proxy (defaults to http://localhost:4000)
const model = createChatModel({
  provider: "litellm",
  apiKey: process.env.LITELLM_API_KEY ?? "",
[...]

PR #267, opened 2026-08-30
feat: add OrcaRouter as a first-class LLM provider

lovejones2914-spec:
## What

Adds **OrcaRouter** as a first-class `GRAFT_PROVIDER` in Graft, mirroring the existing `litellm` wiring exactly: a named provider over the OpenAI-compatible wire format, pointed at the OrcaRouter AI gateway by default, with `ORCAROUTER_API_KEY` / `ORCAROUTER_BASE_URL` / `ORCAROUTER_MODEL` env fallbacks and `/v1/models` auto-discovery.

Today Graft users who want OrcaRouter set `GRAFT_PROVIDER=openai`, `GRAFT_BASE_URL=https://api.orcarouter.ai/v1`, and their key — treating the gateway as an anonymous custom base URL. This change lets them pick `orcarouter` directly, the same way `litellm` already lets them pick a LiteLLM proxy.

[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding orcarouter as a first-class provider means this project's users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL.

It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

## Why this shape

Graft's `provider` names the *wire format*, and `litellm` already proves the convenience-provider pattern: an OpenAI-compatible gateway gets a named entry point, a default base URL, `GRAFT_PROVIDER`-driven routing, and its own env fallbacks. OrcaRouter is the same shape, so it gets the same treatment.

## File-by-file (mirrors the `litellm` wiring)

| File | What it does, by analogy |
|---|---|
| `src/ai/llm/orcarouter.ts` | **new** — `OrcaRouterChatModel extends OpenAIChatModel`, defaulting `baseUrl` to `https://api.orcarouter.ai/v1` and labeling as `orcarouter:<model>`; `listOrcaRouterModels()` for `GET /v1/models` catalog discovery. Parallel to `src/ai/llm/litellm.ts`. |
| `src/ai/llm/factory.ts` | `ProviderKind` gains `"orcarouter"`; `createChatModel` routes it to `OrcaRouterChatModel`. Parallel to the `litellm` case. |
| `src/ai/providers.ts` | `resolveConfig` honors `ORCAROUTER_API_KEY` / `ORCAROUTER_BASE_URL` / `ORCAROUTER_MODEL` as env fallbacks, and defaults the `orcarouter` provider to the gateway base URL, with a provider-prefixed default model (`openai/gpt-4o-mini`). Parallel to the `litellm` entries. |
| `src/cli.ts` | `--provider` help text now lists `orcarouter`. |
| `src/index.ts` | Exports `OrcaRouterChatModel`, `listOrcaRouterModels`, `DEFAULT_ORCAROUTER_BASE_URL` (and, while here, the previously-unexported `litellm` equivalents). |
| `test/orcarouter-adapter.test.ts` | **new** — network-free tests mirroring `test/litellm-adapter.test.ts`: translation reuse, `orcarouter:` label, factory wiring, default base URL, model discovery. |
| `.env.example` / `README.md` | Document the `orcarouter` provider and its env vars, using Graft's existing provider vocabulary. |

## Verification

- `tsc -p tsconfig.json` passes.
- Network-free adapter tests pass (`test/orcarouter-adapter.test.ts` + the existing `litellm` / `openai` / `anthropic` adapter suites).
- **Live L3 test**: `createChatModel({ provider: "orcarouter", apiKey, model: "openai/gpt-4o-mini" })` hit the real gateway at `https://api.orcarouter.ai/v1` and returned a completion (`L3-OK`) over the new code path. With `GRAFT_PROVIDER=orcarouter`, `resolveConfig` resolves the gateway base URL, `ORCAROUTER_API_KEY`, and the default model, and a factory-built `OrcaRouterChatModel` completes end-to-end.

Note: I could not run the full test suite in this container — the tree-sitter grammars are native addons and this environment has no compiler to build them (`npm ci`'s node-gyp step fails on `make`), so all tree-sitter-dependent tests crash at load. All non-tree-sitter tests pass. CI's Ubuntu/Windows legs will exercise the full suite.

> Full-suite caveat from the container, for reviewers: `test/` files that import `tree-sitter` crash with `No native build was found` here because the grammars can't compile without a C toolchain. This is an environment limitation, not a test regression from this diff.

## Contact

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.


shhdwi:
Verified locally on top of current `main` (clean merge; build clean). `OrcaRouterChatModel extends OpenAIChatModel` (OpenAI-wire-compatible), overriding only the default base URL + label — the same clean pattern as the merged LiteLLM provider (#220). Registration confirmed end-to-end: `ProviderKind` extended, `createChatModel` case added, `DEFAULT_MODELS` stays a total record (exhaustive `never` still compiles), and `resolveConfig` selects it via `GRAFT_PROVIDER=orcarouter` + `ORCAROUTER_API_KEY/MODEL/BASE_URL`; openai/anthropic/litellm resolution unaffected. orcarouter 4/4, litellm 4/4, llm-adapters 12/12. Thanks for another provider! Merging.

PR #268, opened 2026-08-30
fix(windows): hide detached background processes

MarceloCaporale:
## Summary

On Windows, detached background Graft processes can open a visible, empty console window when `windowsHide` is omitted.

This is especially noticeable when Graft is invoked from Codex hooks: after an edit marks the index dirty, the Stop hook launches the background sync (`sync-run.js` -> `graft build .`), which can show a blank `node.exe` console.

## Changes

Add `windowsHide: true` to the three detached background spawns in:

- `src/claude/hooks.ts` — background index sync
- `src/telemetry/flush.ts` — telemetry flush
- `src/upkeep.ts` — daily update check

The explicit browser opener in `dist/cli.js` remains unchanged. `dist/` is generated and was not edited.

`windowsHide` is ignored on non-Windows platforms.

## Validation

- `npm run build` passes.
- Relevant hook/upkeep/telemetry tests: 53 passed; 1 existing failure in the savings-format assertion, unrelated to this change.
- Full `npm test`: 1,047 passed, 7 existing unrelated failures, 5 skipped.

### DIFF
--- a/src/claude/hooks.ts
+++ b/src/claude/hooks.ts
@@ -296,7 +296,7 @@ function handleStop(input: any, dir: string): void {
   const stats = readStats(dir);
   if (stats?.dirty && acquireLock(dir)) {
     patchStats(dir, { syncing: true });
-    const child = spawn(process.execPath, [syncRun, dir], { detached: true, stdio: 'ignore' });
+    const child = spawn(process.execPath, [syncRun, dir], { detached: true, stdio: 'ignore', windowsHide: true });
     child.unref();
   }
 }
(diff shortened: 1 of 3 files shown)


shhdwi:
Verified locally on top of current `main` (clean merge; build clean). Adds `windowsHide: true` to the three silent background spawns (Stop-hook async rebuild in `hooks.ts`, telemetry flush in `flush.ts`, update-check in `upkeep.ts`), keeping `detached: true`. `windowsHide` is a no-op off Windows, so zero behavior change here. The one untouched detached spawn (the user-initiated browser opener) is correctly left visible. 63/63 tests (claude-hooks, upkeep, telemetry). Nice, tightly-scoped fix — thanks! Merging.

Issue #253, opened 2026-08-28
recover-tool: accept a bare item array and repair truncated JSON (small local models via Ollama drop the tool envelope / hit max_tokens); add an opt-in dump of unparsed content


yelaman-toktarov:
## Summary

Follow-up to #129 / #172 (and likely the same family as #235). `recoverToolArgsFromContent` (`dist/ai/llm/recover-tool.js`) rescues a forced-tool payload that a gateway put into `content`, but it rejects two shapes that small local models produce routinely. With `qwen2.5-coder:7b` behind Ollama's OpenAI endpoint this made the meaning tier fail on **every** file (`⚠ crux: model did not honor tool_choice and content is not parseable tool-call JSON`) and left one synthesis batch empty — with no way to see what the model actually returned.

**Version:** graft 0.15.0, provider `openai` → Ollama `/v1` (`qwen2.5-coder:7b`, `num_ctx 24576`, `num_predict 8192`), Node 23.7.

## What the model returns (captured by dumping `res.text`)

1. **Bare item array, no wrapper.** On longer outputs the model drops the `{ "symbols": [...] }` / `{ "name": "record_symbols", "arguments": {...} }` envelope and returns just the items — for crux:
   ```json
   ```json
   [
     {"id": "…/FreshUserFactory.java#createAndLogin", "summary": "…", "crux_start": 38, "crux_end": 41},
     …
   ]
   ```
   ```
   and for synth the same with concept-node objects (`{"name": …, "type": …, "summary": …, "sources": […]}`). On a 5-symbol file the envelope is kept and recovery works; from ~15 symbols up it is dropped and `asPayload` returns `undefined` (an array item that carries `name` is treated as a possible tool wrapper and, lacking `parameters|arguments|args`, rejected). Result: 12 warnings in the first 10 files, i.e. ~100 % crux failure.

2. **Truncated JSON at `max_tokens`.** A synth batch of ~56 files (~14.8k prompt tokens) or a crux request for a 154-symbol file produces a *valid prefix* that ends mid-item. Recovery tries `JSON.parse` on the whole string and on the `{…}` / `[…]` slices, all fail → batch/file empty. The crux path already has a 2-attempt retry over `missing` ids that would happily finish the remainder — if the first pass returned the 106 complete items instead of nothing.

## What worked for us (local patch on `dist/`, happy to send a PR)

In `recoverToolArgsFromContent`:

```js
// (a) bare array of payload items → treat as the payload
if (Array.isArray(value)) {
[...]

PR #269, opened 2026-08-30
fix(llm): recover bare arrays and truncated JSON from gateway content (#253)

Frankie-Xu:
## Summary
- Follow-up to #129 / #216. `recoverToolArgsFromContent` still rejected two shapes that small local models (Ollama `qwen2.5-coder:7b`) emit routinely: a **bare item array** with no `{name, parameters}` envelope, and **truncated JSON** when `max_tokens` / `num_predict` cuts the last item. Meaning then failed on every file (`content is not parseable tool-call JSON`).
- **Bare array:** after trying wrapper items, an array of objects that do not carry `parameters|arguments|args` is the payload. Synth nodes legitimately have `name`; that is no longer treated as a tool wrapper.
- **Truncation:** after whole-string / `{…}` / `[…]` parse fail, a linear scan (no regex) cuts at the last `}` that is not inside a string and appends only unmatched `]` / `}` (cap: 64 cuts, 8 closers). Incomplete last items are dropped, not guessed. Unrepairable prefixes still return `undefined` and warn.
- **Not in this PR:** opt-in dump of unparsed content (`GRAFT_DUMP_UNPARSED_DIR` / `--debug-unparsed` in the issue). That is a separate diagnostic surface; leaving it out keeps this to the two recovery shapes.

## Overlap with #254
Does not share files with #254 (`src/ai/crux.ts`, `src/ai/failure.ts`, `src/graph/enrich.ts`). This PR only touches `src/ai/llm/recover-tool.ts` and `test/llm-ops.test.ts`, and does not change `LlmFailureGate`. Either PR can land first; no rebase needed unless #254 later expands into `recover-tool.ts`.

## Test plan
- [x] Bare item array (synth nodes with `name`, crux `{id,…}`, fenced) → payload
- [x] Truncated JSON at a complete object (bare array, `{nodes:[…]}` envelope, nested `}`) → complete items only
- [x] Unrepairable truncation (no complete object, `}` only inside an open string) → `undefined`, no guess
- [x] 8k spaces around a truncated prefix stays linear (CodeQL `js/polynomial-redos` regression; no new regex)
- [x] Real `toolCalls` still preferred over a bare content array
- [x] `ChatSynthesizer` / `ChatCruxSummarizer` recover through `clean()` / `parseResults()`
- [x] Existing #129 shapes still parse; wrong-tool wrapper still refused
- [x] `npm run build` and `npx tsc --noEmit` clean; `node --import tsx --test test/llm-ops.test.ts` — 16 pass

Addresses #253

### DIFF
--- a/src/ai/llm/recover-tool.ts
+++ b/src/ai/llm/recover-tool.ts
@@ -1,13 +1,16 @@
 /**
  * Recover a forced-tool payload from assistant `content` when a gateway ignored
- * `tool_choice` and wrote JSON text instead of `tool_calls` (#129).
+ * `tool_choice` and wrote JSON text instead of `tool_calls` (#129, #253).
  *
  * Conservative: only succeeds when `text` parses as JSON matching one of:
  *   - `[{ name, parameters|arguments|args }]`
  *   - `{ name, parameters|arguments|args }`
  *   - a fenced ```json block wrapping either
  *   - a bare object whose `payloadKey` is an array
- * Anything else (prose, truncated JSON, a guessed shape) returns undefined.
+ *   - a bare array of payload items (no tool envelope) — items may carry `name`
+ * Truncated JSON is repaired only by closing unmatched `[`/`{` after the last
+ * complete object (linear scan, bounded closers). Anything else (prose,
+ * unrepairable truncation, a guessed shape) returns undefined.
  * Callers MUST run the same schema validation they use for real tool calls.
  */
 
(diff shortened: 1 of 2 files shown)


Frankie-Xu:
Kept this as Draft while the review queue was full.

Rebased onto `main` at aa1e2bb (`release: 0.16.0`). Follow-up to #216: recover a bare item array and truncated JSON from gateway content (#253), linear scan only (no backtracking regex). Real `toolCalls` still win; empty recoveries are not treated as a successful cache (#177).

Closes #253.

Frankie-Xu:
Rebased onto main after #277 so the stale not-ok-197 checks could re-run — all checks green now.

Frankie-Xu:

Issue #260, opened 2026-08-28
`--deep`: one crux call per file cannot describe a file with more than ~125 symbols — the truncated tool response is silently dropped


vladig-kz:
**Title:** `--deep`: one crux call per file cannot describe a file with more than ~125 symbols — the truncated tool response is silently dropped

**Version:** 0.13.0 · node v24.19.0 · Linux · repo: PHP/Laravel, 644 files, 5997 nodes
**Provider:** OpenAI-compatible, `--base-url https://api.deepseek.com --model deepseek-chat`

## Symptom

After every other file succeeded, four files failed no matter how often `--deep` was re-run:

```
✗ app/Http/Controllers/RequestController.php: model returned no usable symbol summaries
✗ app/Http/Controllers/Controller.php: model returned no usable symbol summaries
✗ app/Http/utils.php: model returned no usable symbol summaries
✗ public/js/app.js: model returned no usable symbol summaries
```

The discriminator is the number of targets, not file size:

| file | bytes | symbols | result |
|---|---|---|---|
| `public/js/sceditor/sceditor.js` | 251 KB | 109 | ok |
| `app/Models/Model.php` | 185 KB | 122 | ok — largest that worked |
| `RequestController.php` | 93 KB | 129 | fails |
| `Controller.php` | 79 KB | 144 | fails |
| `app/Http/utils.php` | 140 KB | 187 | fails |
| `public/js/app.js` | 337 KB | 698 | fails |

## Root cause

`ChatCruxSummarizer.describeFile` (`dist/ai/crux.js:131`) sends every target of a file in a
single request under `maxTokens: 8192` (`dist/ai/crux.js:136`). At ~90 completion tokens per
entry that ceiling is reached around 125 entries. Raw response for the 129-target file:

```
  request max_tokens: 8192
  finish_reason: length
  usage: {"prompt_tokens":13023,"completion_tokens":8192,"total_tokens":21215}
  tool_calls returned: 1
  arguments string: 31521 chars
  JSON.parse(arguments) FAILED: Unterminated string in JSON at position 31521
  ...tail: "...\"crux_start\": 1785, \"crux_end\": 1787}, {\"id\": \"app/Http/Controllers/
             RequestController.php#RequestController.getNotify |"
```

The model was answering correctly and simply ran out of budget two-thirds of the way through.
graft cannot parse the truncated arguments, `argsFromResponse` falls through to
`recoverArgsFromContent` (the content is empty — it was all in the tool call), `parseResults`
[...]

PR #270, opened 2026-08-30
fix(deep): recover crux payloads when the model echoes the source line (#259)

### LINKED ISSUE #259: `--deep`: a model that echoes the whole target line makes every file fail with "no usable symbol summaries"
**Title:** `--deep`: a model that echoes the whole target line makes every file fail with "no usable symbol summaries"

**Version:** 0.13.0 · node v24.19.0 · Linux · repo: PHP/Laravel, 644 files, 5997 nodes
**Provider:** OpenAI-compatible, `--base-url https://api.deepseek.com --model deepseek-chat`

## Symptom

The concept pass succeeds, the crux pass fails on every single file:

```
✗ app/Collection.php: model returned no usable symbol summaries
✗ app/User.php: model returned no usable symbol summaries
✗ app/Database/EloquentBuilder.php: model returned no usable symbol summaries
...
✗ the deep pass did not complete — the meaning tier is incomplete.
  summaries: 6 files in a row failed, so the pass stopped rather than keep calling.
  meaning coverage: 47/5997 symbols (1%).
```

## Root cause

`userContent` (`ai/crux.ts` → `dist/ai/crux.js:39`) renders each target as one line whose
`id=` prefix is followed by three more `|`-separated fields:

```
- id=app/User.php#User.__construct | method | lines L40-L58 | public function __construct($object = null)
```

The system prompt then asks for "that id verbatim". The model reads `id=` as introducing the
whole line and returns it intact. Captured raw tool call:

```json
{
  "id": "app/User.php#User.__construct | method | lines L40-L58 | public function __construct($object = null)",
  "summary": "Builds an User from a source object, copying matching public properties and wrapping each Properties entry into an UserProperties instance.",
  "crux_start": 47, "crux_end": 57
}
```

Everything except the `id` is correct: one entry per target, no blank summaries, crux ranges
inside their spans, `0/0` where the prompt asks for it. But `parseResults`
(`dist/ai/crux.js:56`) stores `id: s.id` unchanged, `enrichGraph` matches with
`results.get(node.id)` (`dist/graph/enrich.js:109`), nothing matches, `applied === 0`, and the
file is reported as a failure (`dist/graph/enrich.js:130`). After 5 such files
`MAX_CONSECUTIVE_FAILURES` (`dist/ai/failure.js:17`) stops the whole pass.

So a fully correct answer is discarded over a formatting detail in the id.

## Reproduction

Any provider that echoes the target line back. Reproduced on every `--deep` run against
`deepseek-chat`, and on single-file probes issued directly to the same endpoint. Minimal check:
run `--deep` on one small file and log `parseResults()` output next to the requested node ids.

## Suggested fix

Tolerant matching is the smaller change and fixes existing prompts in the wild — normalising
the returned id by cutting at the first `" | "` was enough here to go from 0/5 to 5/5 matched
on the same request (no node id in a real graph contains that sequence). Falling back to a
suffix/prefix match against the requested ids would be more robust still.

Worth doing on top: make the target line unambiguous, e.g. put the id last, quote it, or send
targets as JSON. As a side benefit the echoed suffix costs ~120 characters of output per
entry — roughly a third of the response budget on a large file.

## Note on the provider

`deepseek-chat` is currently served by DeepSeek as `deepseek-v4-flash` (the response `model`
field says so). Its sibling alias `deepseek-v4-pro` rejects graft's forced `tool_choice`
outright with `400 Thinking mode does not support this tool_choice`, so the object-form
fallback at `dist/ai/llm/openai.js:124` retries with `"required"` and gets the same 400. Same
model family, different alias, different capability — may be worth a line in the docs about
which DeepSeek alias works.

Frankie-Xu:
## Summary
- `deepseek-chat` copies the whole TARGET line into `record_symbols.id` (`id | kind | lines Lx-Ly | signature`). `summary` and crux ranges are fine; `enrichGraph` looks up `results.get(node.id)`, matches nothing, and every file dies as "no usable symbol summaries".
- `parseResults` now peels that TARGET-line suffix onto a **requested** id. A lone `|`, a truncated line, or an id that peels to something we did not ask for is left unchanged — rescue failure is still a failure, and blank summaries stay `pending` (#177). No prompt change (0 extra tokens); no #260 chunking.

Closes #259

## Payload shape (issue sample)

**Before** (what the model returned — `id` is the whole TARGET line):

```json
{
  "id": "app/User.php#User.__construct | method | lines L40-L58 | public function __construct($object = null)",
  "summary": "Builds an User from a source object, copying matching public properties and wrapping each Properties entry into an UserProperties instance.",
  "crux_start": 47,
  "crux_end": 57
}
```

**After peel** (what `parseResults` stores — lookup key matches `node.id`):

```json
{
  "id": "app/User.php#User.__construct",
  "summary": "Builds an User from a source object, copying matching public properties and wrapping each Properties entry into an UserProperties instance.",
  "crux_start": 47,
  "crux_end": 57
}
```

Misses are named, not lumped into "no usable":
- echo that does not map to a requested id → `model echoed the target line into each symbol id`
- empty/`{}` tool args (truncated JSON from the OpenAI adapter) → `empty or truncated tool response`
- blank summaries → `no usable symbol summaries` (unchanged, still not cached)

## Test plan
- [x] `node --import tsx --test test/crux-echo.test.ts` — issue sample rescue, garbage not rewritten, blank summaries stay pending, truncation vs echo, clean `toolCalls` regression
- [x] `node --import tsx --test test/llm-ops.test.ts test/graph-enrich-pending.test.ts test/graph-enrich-failures.test.ts` — normal toolCalls / #172 / #127 still green
- [x] `npm test` — 1070 passing locally

### DIFF
--- a/src/ai/crux.ts
+++ b/src/ai/crux.ts
@@ -134,15 +134,95 @@ function userContent(input: FileCruxInput): string {
   return `FILE: ${input.path}\n\n${numberLines(input.source)}\n\nTARGETS (${n} — return all ${n}, one entry per id):\n${targets}`;
 }
 
+/**
+ * Strip a TARGET-line echo from a crux `id` (#259).
+ *
+ * {@link userContent} renders each target as
+ * `id=${id} | ${kind} | lines L${start}-L${end} | ${signature}`. Some models
+ * (deepseek-chat) copy that whole line into the tool's `id`. Real node ids are
+ * `path#Name` and do not contain ` | `.
+ *
+ * Only rewrite when the suffix is that TARGET shape AND the peeled token is one
+ * of the ids this call asked for (or no expected list was given). A lone ` | `,
+ * a truncated line, or an id that peels to something we did not ask for is left
+ * unchanged so the caller can fail it with a specific reason instead of caching
+ * a guess (#177).
+ */
+export function peelEchoedCruxId(raw: string, expectedIds: readonly string[] = []): string {
+  const trimmed = stripIdWrapper(raw);
+  if (expectedIds.includes(trimmed)) return trimmed;
+
+  const peeled = peelTargetLineSuffix(trimmed);
+  if (peeled !== null && (expectedIds.length === 0 || expectedIds.includes(peeled))) {
+    return peeled;
+  }
+
+  if (expectedIds.length > 0) {
+    const hit = longestPrefixId(trimmed, expectedIds);
+    if (hit) return hit;
+  }
+  return trimmed;
+}
+
+/** True when `id` still looks like the TARGET line, not a node id. */
+export function looksLikeEchoedTargetLine(id: string): boolean {
+  return peelTargetLineSuffix(stripIdWrapper(id)) !== null;
+}
+
+function stripIdWrapper(raw: string): string {
+  let s = raw.trim();
+  if (
+    (s.startsWith('"') && s.endsWith('"') && s.length >= 2) ||
+    (s.startsWith("'") && s.endsWith("'") && s.length >= 2)
+  ) {
+    s = s.slice(1, -1).trim();
+  }
+  return s.startsWith("id=") ? s.slice(3) : s;
+}
+
+/**
+ * If `id` is `realId | kind | lines Lstart-Lend [| signature]`, return `realId`.
+ * Linear scan — the distinctive marker is ` | lines L`, which we emit and no
+ * node id contains. Kind must be a single token (the {@link Kind} vocabulary).
+ */
+function peelTargetLineSuffix(id: string): string | null {
+  const marker = " | lines L";
+  const linesAt = id.indexOf(marker);
+  if (linesAt < 0) return null;
+  const before = id.slice(0, linesAt);
+  const sep = before.lastIndexOf(" | ");
+  if (sep < 0) return null;
+  const kind = before.slice(sep + 3).trim();
+  if (!kind || /\s/.test(kind)) return null;
+  const after = id.slice(linesAt + marker.length);
+  if (!/^\d+-L\d+(?:\s+\|.*)?$/.test(after)) return null;
+  const peeled = before.slice(0, sep).trim();
+  return peeled || null;
+}
+
+/** Longest requested id that `raw` equals or continues as `id | …` (echoed rest). */
+function longestPrefixId(raw: string, expectedIds: readonly string[]): string | undefined {
+  let best: string | undefined;
+  for (const id of expectedIds) {
+    if (raw === id || raw.startsWith(`${id} | `)) {
+      if (!best || id.length > best.length) best = id;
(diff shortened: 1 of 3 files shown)


Frankie-Xu:
Rebased onto current `main` (`aa1e2bb` — release 0.16.0). Still Draft; still closes #259. Waiting on #269 before marking Ready.

**Conflicts** (both files also touched by #254's miss-class / circuit-breaker split):
- `src/ai/crux.ts` — kept #259's `peelEchoedCruxId` + `expectedIds` into `parseResults`, and #254's `classifyCruxMiss` / `lastMiss` on `describeFile`.
- `src/graph/enrich.ts` — empty/`{}` replies stay on #254's miss class + `quality: true` (do not trip `LlmFailureGate`). Applied-zero replies keep #259's named reasons (echo vs unmatched ids); blank summaries still go through `formatCruxMiss` (`empty-parsed`). Rescue still requires a requested id; garbage is not rewritten.

**Tests** (Node v26.5.0):
- targeted crux / enrich / llm-ops: **54/54**
- `npm test`: **1174/1174**
- `npm run build`: green

No `gh pr ready`. No #260 chunking.

Frankie-Xu:
Cross-check with #332 (L4XB, still open, Fixes #327).

**Overlap:** both PRs rewrite parseResults() in src/ai/crux.ts so an echoed TARGET line becomes the bare node id.

- #332 
ormalizeTargetId: trim, cut at the first  | , then drop a leading id=. The #259 sample (pp/User.php#User.__construct | method | lines L40-L58 | public function __construct(...)) peels correctly under that rule, as does Grok's id=<id> | kind | lines … shape. So the parseResults half of this draft is already covered if #332 lands.
- This PR's peelEchoedCruxId is stricter (requires a  | lines L marker and an id from expectedIds) and also changes src/graph/enrich.ts (unusableSymbolsReason / looksLikeEchoedTargetLine) so an echo miss is not reported as blank-summary / #177.

**Suggestion:** keep this Draft. Wait for #332 to land for the parseResults peel, then either close the crux.ts overlap or narrow this PR to the enrich-only miss taxonomy (echo vs truncated vs unmatched ids). Do not Ready both — they will conflict in parseResults.

Leaving Ready to the open-PR window; not marking this ready from here.

Frankie-Xu:
Superseded by https://github.com/trailhq/Graft/pull/332.

Both PRs rewrite parseResults so an echoed TARGET line (id | kind | lines Lx-Ly | ...) becomes the bare node id. #332 normalizeTargetId already peels the #259 deepseek sample; after c057407 it only peels when the result is a requested id, which was the conservatism this draft was keeping.

Closing rather than marking Ready. The leftover enrich.ts miss-reason naming in this branch is not worth a third overlapping PR.

Leaving issue #259 open until #332 lands.

PR #272, opened 2026-08-31
Astro client <script> blocks: a container language can carry several embedded shapes

ilativity:
> Stacked on #271 — the container row for Astro. Please merge that first; this branch contains its commit and the diff here will shrink to one commit once it lands.

## What

The fence was half the file. An Astro page's *client* behaviour lives in `<script>` tags inside the markup — in the repo this was verified against, one page keeps a **900-line block** there and a layout has four of them at **depth 3** — and none of it was reachable, because `blocks()` walked the root's named children looking for a single `block`/`body` pair.

`ContainerLang` now carries `embeds: EmbeddedBlock[]`, so a language can declare several shapes:

```js
{
  name: "astro", exts: [".astro"], wasm: "astro",
  embeds: [
    { block: "frontmatter",    body: "frontmatter_js_block" },
    { block: "script_element", body: "raw_text", nested: true, accept: isJavaScriptScript },
  ],
  inner: "typescript",
}
```

Vue becomes `embeds: [{ block: "script_element", body: "raw_text" }]` — same behaviour, one indirection.

## Three decisions worth reviewing

**`nested` is opt-in per shape, not a change to the walk.** A Vue SFC's `<script>` is always top-level. Descending for it would newly index a `<script>` written inside a `<template>` — a `.vue` behaviour change smuggled in with the Astro work. There's a test pinning that Vue still ignores it.

**`accept` exists because not every `<script>` holds code.** `<script type="application/ld+json">` is structured data, and its body — `{"@type": "Organization"}` — parses as a perfectly valid TypeScript block, so without a check the graph mints nodes out of a JSON-LD blob and nothing downstream complains. `isJavaScriptScript` reads the `type` attribute: absent or valueless is JavaScript per the HTML spec; `module` / `text/javascript` / `text/typescript` and friends are code; `ld+json`, `importmap`, `speculationrules` are not. A `type={dynamic}` that can't be read statically is treated as data — one un-indexed block beats guessing at content we cannot see.

**Blocks now come back in document order across shapes**, not registry order. Not cosmetic: `extractContainer` mints ids in arrival order, so a name defined in both the fence and a script would otherwise take its `~2` suffix from however the registry happens to be written.

## Verification

Six new tests (22 in the file, `npm test` = **1170 tests, 0 failures**): a script three levels down with a non-zero, non-frontmatter offset; document order and per-block offsets across a fence and two scripts; duplicate names across shapes; the JSON-LD/`importmap`/dynamic-type matrix; and the Vue non-descent regression.

Spans were also checked by hand against real source rather than only against fixtures:

| symbol | reported | actual |
|---|---|---|
| `reclassifyHikes` | `[lang]/hikes/index.astro:L826-L845` | line 826 of a 2,114-line page ✓ |
| `loadMoreActivity` | `admin/UserProfileModal.astro:L2165-L2201` | line 2165 of a 2,596-line component ✓ |

On that repo, over #271: **2,808 → 3,120 nodes, 6,561 → 7,262 edges**, functions 1,641 → 1,906, methods 64 → 95. `reclassifyHikes` answered "no hits in 486 indexed files" before this work started.

## Still out

Component usage in the markup (`<Card />`) is not a call in any script block, so it still produces no edge — "who renders this component" remains unanswerable for Astro. That needs the markup itself modelled, which is a different tier of work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,30 @@
 
 ### Added
 
+- **Astro (`.astro`) joins the container tier**, the second row after Vue and the
+  one its registry comment was holding open. The `---` frontmatter is the same
+  shape as a Vue SFC's `<script>` with a different fence: `tree-sitter-astro`
+  (already in the bundled `tree-sitter-wasms`) parses the file to a top-level
+  `frontmatter` node whose only named child, `frontmatter_js_block`, holds
+  TypeScript — so the row is `block: "frontmatter", body: "frontmatter_js_block",
+  inner: "typescript"` and no extraction code changes. The span arithmetic
+  carries over unchanged and was the thing verified rather than assumed:
+  `frontmatter_js_block` starts on the opening `---` row exactly as `raw_text`
+  starts on its tag row, so slice line N lands on file line N +
+  `startPosition.row`. A fixture whose fence is pushed down the file by a leading
+  HTML comment pins it, because the usual case — a fence on line 1 — has offset 0
+  and passes even when the shift is dropped.
+
+  Frontmatter is where an Astro page's imports and server-side code live, so
+  pages stop being invisible: a `lib` function's `callers` now names the pages
+  that import it, and `graft grep` reaches `.astro` at all. Client `<script>`
+  blocks are deliberately still out — they nest inside the markup, and `blocks()`
+  walks the root's named children only, so reaching them needs a recursive walk
+  and a `block` that can name two node types. A file with only a client script
+  degrades to a file node; it never reports a wrong line. Svelte stays out of the
+  registry for the reason Astro just left it: nobody has verified its `body` node
+  against a real repo.
+
 - **`graft init --no-statusline`** (and `GRAFT_NO_STATUSLINE=1`) skips writing
   Claude Code's `statusLine` / `subagentStatusLine`. A custom bar — in the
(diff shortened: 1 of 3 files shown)


ilativity:
Following the bot's **0 of 4 reached** on Container Parsing — worth explaining rather than leaving to be guessed at.

Three of the four (`attrValue`, `descendants`, `blocks`) are private and reached only through `extractContainer`, which is what the 22 tests in `container-extract.test.ts` drive. That is deliberate: what matters at that level is which spans come back, not which helper saw which node, and exporting internals to satisfy a reachability heuristic would be the tail wagging the dog. So this is the undercount the bot's own note describes — "read a low ratio as 'look here', never as a coverage gate".

The fourth is different, and the bot was right to point at it. `isJavaScriptScript` is exported and carries a truth table, so I've added a direct test of the predicate itself (`3dc9a31`): 19 tags covering absent / bare / empty / quoted / unquoted / upper-case / padded values, the JSON family, `importmap`, `speculationrules`, and the interpolated `type={expr}` case.

**Writing it found a real defect**, which is the argument for the test. `<script type>` — the attribute present with no value — was being classified as data and skipped. HTML reads a valueless attribute as the empty string, and an empty `type` is a classic script, so it is JavaScript. `attrValue` was conflating "no value" with "not a literal"; only the second may return null now. Fixed in the same commit.

Suite is green: 1171 tests, 0 failures.

PR #274, opened 2026-08-31
fix(deep): chunk crux calls for symbol-dense files (#260)

### LINKED ISSUE #260: `--deep`: one crux call per file cannot describe a file with more than ~125 symbols — the truncated tool response is silently dropped

Frankie-Xu:
## Summary
- One crux call per file hits `maxTokens: 8192` around ~125 symbols. The OpenAI adapter's `JSON.parse` of the truncated tool arguments fails, `parseResults` returns `[]`, and the whole file is dropped. Re-running `--deep` hits the same ceiling, so those files stay failed forever.
- **A (this PR):** chunk targets at `CRUX_CHUNK_SIZE = 80` (pinned by test; ~90 completion tokens/entry stays under 8192). Dense files take several sequential calls; results are merged. Small files still cost one call. `finish_reason=length` is logged instead of swallowed. Empty chunks stay `pending` and are not cached as `ready` (#177).
- **B (not this PR):** bump `maxTokens` and retry once on `finish_reason=length`. Rejected: `public/js/app.js` has 698 targets (~65k completion tokens); there is no ceiling that covers that in one call.

Orthogonal to the #254 circuit breaker (quality misses vs quota) and Draft #270 (echoed TARGET-line ids). Does not touch `--only-dir` (#262), `recover-tool` (#269), or the echo peel.

Closes #260

## Rebase vs Draft #270
Both PRs edit `src/ai/crux.ts`. This one only changes packing (`describeFile` splits + merges; `parseResults` untouched). #270 peels echoed ids inside `parseResults`. They compose: chunk, then peel per chunk.

Suggested order: land this, rebase #270 onto it (conflict should be `describeFile` vs `parseResults`, not overlapping hunks). If #270 lands first, rebase this onto it — the `describeFile` rewrite is the larger hunk.

## Test plan
- [x] Mock ceiling at 125 symbols (the observed cliff, **test-only** — not a production constant). 200 targets in one call → 0 summaries; after chunking at 80 → 3 calls (80/80/40), 200 summaries. No DeepSeek key, no live API.
- [x] Small file (5 targets) is still one `model.create`
- [x] `finish_reason=length` with empty `{}` tool args logs `truncated` / `finish_reason=length` (the silent-drop shape)
- [x] Empty chunks: 200 nodes stay `pending`, not cached `ready` (#177)
- [x] A later empty chunk keeps the first chunk's summaries; missed ids stay `pending`
- [x] `node --import tsx --test test/crux-chunk.test.ts test/llm-ops.test.ts test/graph-enrich-pending.test.ts test/graph-enrich-quality.test.ts test/graph-enrich-failures.test.ts`
- [x] `npm run build`
- [ ] No DeepSeek/OpenAI key in this window — did not replay a live dense PHP/JS file

### DIFF
--- a/src/ai/crux.ts
+++ b/src/ai/crux.ts
@@ -1,8 +1,10 @@
 /**
- * Tier-2 "meaning" call for the code graph — batched one request per file.
+ * Tier-2 "meaning" call for the code graph — batched per file, chunked by symbol
+ * count so a dense file cannot blow `maxTokens` and drop the whole reply (#260).
  *
  * Given a source file (with 1-based line numbers) and the list of definitions in
- * it, one call returns, for each definition:
+ * it, one call (or several chunks of {@link CRUX_CHUNK_SIZE}) returns, for each
+ * definition:
  *   1. `summary` — one plain-English sentence: what the symbol is *for*, at the
  *      business-logic level, not a restatement of its signature.
  *   2. `crux_start`/`crux_end` — the smallest contiguous range of FILE line
@@ -10,9 +12,10 @@
  *      the decision or rule the code encodes. `0/0` means there is no single
  *      crux (a trivial getter, a plain data holder).
  *
- * Batching per file means N definitions cost one request, not N — and the model
- * sees each symbol's neighbours, which sharpens the summaries. Line numbers are
- * consumed once, at write time, to slice the crux text verbatim from source.
+ * Small files still cost one request. Files denser than {@link CRUX_CHUNK_SIZE}
+ * are split, the chunks merged, and a `finish_reason=length` stop is logged
+ * instead of silently discarding every target. Line numbers are consumed once,
+ * at write time, to slice the crux text verbatim from source.
  */
 import type { ChatModel, ChatResponse } from "./llm/types.js";
 import { recoverToolArgsFromContent, warnToolChoiceIgnored } from "./llm/recover-tool.js";
@@ -78,6 +81,19 @@ export function formatCruxMiss(kind: CruxMissKind, finishReason: string | null):
   return `model returned no usable symbol summaries [${kind}, finish_reason=${fr}]`;
 }
 
+/** One stderr line when a crux call stopped because the completion hit maxTokens (#260). */
+function warnCruxTruncated(
+  path: string,
+  requested: number,
+  parsed: number,
+  finishReason: string | null,
+): void {
+  const fr = finishReason == null || finishReason === "" ? "null" : finishReason;
+  console.error(
+    `⚠ crux: ${path}: truncated tool response [finish_reason=${fr}] for ${requested} targets (${parsed} parsed) — keeping any other chunks instead of dropping the file`,
+  );
+}
+
 const SYSTEM_PROMPT = `You explain code definitions for a code graph that helps engineers navigate a codebase.
 
 You are given ONE source file with 1-based line numbers, and a list of TARGET definitions in it. Describe EVERY target via the record_symbols tool.
@@ -109,6 +125,21 @@ const SYMBOLS_SCHEMA = {
   },
   required: ["symbols"],
 } as const;
+
+/**
+ * Max targets packed into one crux LLM call (#260). At ~90 completion tokens per
+ * entry, 80 × 90 stays under `maxTokens: 8192`. A file with more symbols is split
+ * and the chunks merged; raising maxTokens cannot cover a 698-symbol file.
+ * Pinned by `test/crux-chunk.test.ts` — do not raise it silently.
+ */
+export const CRUX_CHUNK_SIZE = 80;
+
+/** Split `nodes` into batches of at most `size` (default {@link CRUX_CHUNK_SIZE}). */
+export function chunkCruxTargets<T>(nodes: readonly T[], size = CRUX_CHUNK_SIZE): T[][] {
+  const out: T[][] = [];
+  for (let i = 0; i < nodes.length; i += size) out.push(nodes.slice(i, i + size));
+  return out;
+}
 
 /** Cap the file text sent per request so one huge file can't blow the context. */
 const MAX_CODE_CHARS = 18_000;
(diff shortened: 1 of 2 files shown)


Frankie-Xu:
Rebased onto current main. CI is green. Ready for review.

SeatownSin:
Chunking is the right call over raising `maxTokens` — I reached the same conclusion independently
on a 42 MB repo, for the same reason you give in the "B (not this PR)" note. One thing I think
this misses, though, and I left the measurement on #260 before noticing this PR was open.

The chunk loop spreads the file's source unchanged into every chunk:

```ts
for (const nodes of chunkCruxTargets(input.nodes)) {
  const { parsed, res } = await this.describeChunk({ ...input, nodes });
```

`describeChunk` builds its prompt with `userContent()`, which still calls
`numberLines(input.source)` — and `numberLines` clips at `MAX_CODE_CHARS` (18k) **from the start
of the file**. So chunk 5 of a 7,838-line file is shown the same first ~450 lines that chunk 1
was, and is asked about symbols at line 7,000 that appear nowhere in its prompt.

The output-budget problem is genuinely fixed by this PR. What's left is that the binding
constraint moves: those later chunks now return *successfully*, describing symbols from their
signature alone. That is worse than the truncation it replaces, because truncation was at least
loud — `finish_reason=length`, which this PR now logs properly. A confident summary for a
function the model never saw looks exactly like a good one.

The fix composes with what's already here — window the **source** alongside the targets. Split
on the chunk's real source span rather than a fixed target count, send that slice, and keep the
line numbers file-absolute (`buildCrux` slices the file by whatever the model reports back, and
#260 flags the same constraint).

Measured on a synthetic 500 KB / 8,000-line / 200-symbol file with a stub model:

```
requests made    : 29
symbols returned : 200 / 200
targets whose start line was NOT in the shown source : 0
```

That last line is the assertion worth adding to `test/crux-chunk.test.ts`. The current dense-file
test uses `source: "export const x = 1;\n"` for every node, so the source window is never
exercised — a chunker that shows the wrong 18k passes it. Something like: build the fixture with
real distinct line ranges, then assert every target's `crux_start` falls inside the source that
chunk's prompt actually contained.

On a real 14-file / 1,447-symbol directory, source-span windowing took coverage 6% → 96%, and
the remaining 4% was the decorated-id bug (#327 / #332). Happy to share the windowing patch if
it's useful here — it's about 60 lines and sits in the same `describeFile` you've already
rewritten, so it would be a rebase on this rather than a competing PR.

Frankie-Xu:
Thanks @SeatownSin — you are right. Chunking fixed the output-budget miss, but each chunk still sent the whole `input.source` through `numberLines`, which always clips 18k chars from the **start of the file**. Later TARGETS then sat at L7000 while the prompt showed the first ~450 lines, so the model could write a confident summary of a function it never saw.

Pushed `03655c2` on this branch (not a competing PR):

- `userContent` now slices source to that chunk's real `startLine`–`endLine` span.
- Numbered lines in the prompt stay **file-absolute**, so `buildCrux` can still cut by the model's returned range.
- `test/crux-chunk.test.ts` adds a fixture the one-line dense-file test could not: 80 short functions, then padding past 18k, then a late symbol. Each chunk's prompt must contain its targets' absolute lines.

Did not take a pasted patch. Did not touch `parseResults` / id peel (#332 / #377) or `maxTokens`.

PR #276, opened 2026-09-01
Install Claude Code wiring in ~ too, so worktrees keep graft

anirudhkumar-nanonets:
Fixes the worktree report from sj: graft not running in `.claude/worktrees/<branch>` after `.claude` came out of `.gitignore`.

## Root cause

His `.gitignore` line 1 is a blanket `*.json`. That ignores **both** `.mcp.json` and `.claude/settings.json`. `git worktree add` checks out tracked files only, so the worktree gets graft's `.cjs` shims and neither of the two files that *point at* them — no SessionStart hook, no MCP server. graft is absent in a tree that looks correctly wired.

graft already has the repair — `reconcileWiring` rewrites both files — and it can never run here:

- `runUpkeep` is called only from `claude/hooks.ts` and `mcp/server.ts`, the two things that are missing.
- `seedGraph` can't help either: it runs on the query path, which needs the MCP server.
- The wiring stamp lives in `graft/.cache/`, gitignored and not in `seedable()`'s allowlist, so a worktree never has one.

The fix cannot live in the repo, because a `.gitignore` can erase anything there.

## Change

`graft init` also writes a user-level copy, which no `.gitignore` reaches and Claude Code reads for every project:

```
~/.claude/helpers/graft-hooks.cjs   the shim, addressed absolutely
~/.claude/settings.json             the 4 hook entries
~/.claude.json                      mcpServers.graft (user MCP scope)
```

Codex has been installed this way from the start (`hosts/codex-hooks.ts`, whose own note says its entries "fire in every repo opened with Codex"). Claude Code was the one host graft wired repo-only. `hosts/claude-global.ts` is a deliberate mirror of that file.

Repo-level writes are unchanged — this is the floor underneath, not a replacement.

### Scoped deliberately narrow

- **Hooks only in `~`.** No statusline (a session allows exactly one; taking it machine-wide would silently outrank the user's own), no footer regex, no Bash allowlist. Those stay per-repo, where the user opted in.
- **`tools/list` advertises nothing** in a repo with no graph *and* no built parent checkout — a user-scope registration starts the server in every project, and six schemas per session is real context for tools that can only answer "run graft build". The parent-checkout clause is load-bearing: gating on the current tree alone would hide graft in exactly the fresh worktree this PR exists to fix.
- **`--no-global` now reaches the claude layer** from both the CLI and the wiring replay. It previously meant "no out-of-repo writes, except these three".

## Two reversed invariants

Both were explicit and are now stated the other way in the tests:

- `test/hosts-plan.test.ts` — *"claude writes are all repo-scoped — the claude layer never touches ~"*. It does now, by design.
- `test/mcp-server.test.ts` — an unbuilt repo used to advertise all six tools. It now advertises none. Still callable, so the `graft build` soft error is unchanged.

Also fixed: 8 test call sites passed no `home`, so with this change the suite would have written into the developer's real `~/.claude`. All now use a scratch home.

## Verification

`test/hosts-claude-global.test.ts` drives real `git init` / `git worktree add` with `*.json` ignored, asserting the worktree loses both repo files and the user-level copy covers it. Plus idempotency, foreign-key preservation, unparseable-file refusal, `--no-global`, and retract symmetry.

Live check against a scratch `HOME`:

| repo | tools advertised |
|---|---|
| built parent | 6 |
| fresh worktree, no graph of its own | 6 |
| unrelated repo, never invited graft | 0 |

Full suite: 1069/1071. The one failure (`instructions must stay under 1000 chars, got 1055`) also fails on `main` and is unrelated.

Closes PLG-1072.

### DIFF
--- a/src/claude/init.ts
+++ b/src/claude/init.ts
@@ -1,6 +1,8 @@
 import { mkdirSync, writeFileSync, readFileSync, chmodSync } from 'node:fs';
 import { join, dirname } from 'node:path';
+import { homedir } from 'node:os';
 import { execFileSync } from 'node:child_process';
+import { installClaudeGlobal, type GlobalWrite } from '../hosts/claude-global.js';
 import { mergeGraftSettings } from './settings-merge.js';
 import { statuslineShim, hooksShim } from './shim-template.js';
 import { skillTemplate } from './skill-template.js';
@@ -52,11 +54,16 @@ export interface InitResult {
   skill: string;
   /** the `.mcp.json` write registering the graft MCP server for Claude Code. */
   mcp: McpWrite;
+  /** the user-level writes under `~/.claude`, empty when `global: false`. */
+  global: GlobalWrite[];
   warnings: string[];
   built: boolean;
 }
 
-export function runInit(dir: string, opts: { build?: boolean; cliPath?: string; statusline?: boolean } = {}): InitResult {
+export function runInit(
+  dir: string,
+  opts: { build?: boolean; cliPath?: string; statusline?: boolean; global?: boolean; home?: string } = {},
+): InitResult {
   // Same list `--dry-run` and the picker report, so the two can't drift apart.
   const [settings, statusline, hooks, skill, mcpTarget] = claudeTargets(dir).map((t) => t.path);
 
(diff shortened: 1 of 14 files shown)


anirudhkumar-nanonets:
Rebased onto `main` (0.16.0) — the branch was cut from a 10-commit-stale local `main`, which is why GitHub couldn't compute a merge ref and the blast-pages `publish` job failed at checkout before any build ran. Merged rather than overwrote the 0.16.0 changes it collided with:

- `--statusline` / `statuslineWanted` now threads through `runInit` alongside `--global`.
- `mergeGraftHooks` uses the renamed `isGraftEntry`.
- `claude-global.ts` now reuses `hosts/config-write.ts` (`writeOwned`, `readJsonObject`, `ConfigWrite`) instead of the local copies I'd written — 0.16.0 extracted exactly those.

Two follow-up fixes CI caught that local runs couldn't:

- The worktree tests needed a git identity in the env: a runner has no `user.email`, and blanking `GIT_CONFIG_GLOBAL` removes any it had, so `git commit` exited 128.
- On Windows the installed hook command read `C:\Users\…\helpers/graft-hooks.cjs` — `join` gives backslashes, the template appends `/graft-hooks.cjs`. The path now goes in posix form throughout (Node accepts `/` on Windows).

**Remaining red is pre-existing on `main`, not from this branch:**

| check | cause |
|---|---|
| `test` (both platforms) | `not ok 197 — cursor-session-end force-closes THIS conversation`. Same test, same number, fails on `main`'s own CI for the last 5 commits including `release: 0.16.0`. |
| `analyze` | `Loaded a configuration file for version '4.37.8', but running version '4.37.7'` — `codeql.yml` has `init` bumped by #240 and `analyze` still on the older pin. This branch touches no workflow files. |

Everything else green: `Node 24 WASM regression`, `blast`, `publish`. Local suite 1174 pass / 0 fail / 1 skipped.

PR #296, opened 2026-09-03
feat(lsp): definition and references walks for servers without call hierarchy

hiovi:
## Why

`graft build --lsp` only speaks call hierarchy. A server that offers definition and references but no call hierarchy (several do) answers the readiness probe empty for 90 s and the pass adds nothing, silently. This teaches the pass two more instruments and lets the server's advertised capabilities pick one.

## What

- **Definition walk.** `resolveEdges(nodes, raw, { unresolvedCalls })` now reports the calls it dropped as ambiguous, each with the callee token's position (breadth-tier raw calls carry `pos`; depth-tier calls are unchanged). When the server has `definitionProvider` but no `callHierarchyProvider`, the pass asks `textDocument/definition` at each dropped call site inside a callable and maps the answer to a node. Only the edges the static pass could not settle are asked about.
- **References walk.** With nothing to ask about, each in-repo reference to a function that sits inside another callable becomes a caller edge.
- A server advertising `callHierarchyProvider` keeps the existing walk unchanged (`nodeAtUri`/`link` are shared helpers now). `LspClient` gains `supports()`, `definition()`, `references()`; the readiness probe follows the chosen instrument.
- `enrichWithLsp` takes an optional `server` (test seam) and `callSites`.

## Tests

`test/lsp-fake-server-probe.ts` is a stand-in language server spawned over stdio that answers from the fixture's text and advertises either shape (`FAKE_LSP_MODE`). Three end-to-end tests on a Rust fixture where `compute` is defined in two files (so the breadth tier drops the call): the definition walk settles it, the references walk finds the caller, and a call-hierarchy server still takes the old path. `node --import tsx --test test/lsp-enrich.test.ts test/generic-extract.test.ts test/graph-resolve-typed.test.ts test/graph-references.test.ts test/graph-cross-language.test.ts` is green; `tsc --noEmit` clean.

Split out of #294; no new registry rows here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,9 @@ compiler-grade layer — all `$0` and deterministic (no model, no key):
   `lsp_resolved` call edges (member calls the static pass can't type) when a
   language server is on your `PATH`: **rust-analyzer** (Rust), **clangd** (C/C++),
   **gopls** (Go), **pyright** (Python), **typescript-language-server** (TS/JS).
+  A server without call hierarchy still contributes: graft asks it for the
+  definition of each call the static pass had to drop as ambiguous, or walks its
+  references when there is nothing to ask about.
   It's best-effort — with no server installed the graph is unchanged.
 
 Twenty-three languages in total. A file whose language isn't listed is skipped, not
--- a/src/graph/build.ts
+++ b/src/graph/build.ts
@@ -288,7 +288,11 @@ export async function buildGraph(
     files: entries,
   });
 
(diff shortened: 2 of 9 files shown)


TheRottingTomatoes:
Do you even review the vibecode? https://github.com/trailhq/Graft/issues/230 Please have your agent take a look at this.

hiovi:
Apologies @TheRottingTomatoes My bad. I pushed this PR by mistake thinking I was creating it on my own fork rather than on the original source. Anyway I can have a look at it myself. I'll convert it to draft in the meantime.

Issue #293, opened 2026-09-03
Graft does not index Godot .gd or GLSL .glsl files


rikiyanai:
## Summary

Graft currently reports no indexed definitions for Godot GDScript (`.gd`) and GLSL shader (`.glsl`) files. In a Godot-heavy repository, this makes `graft skeleton` / MCP `graft_file_api` fail on core renderer files even though the files are tracked and present.

## Reproduction

In a repo containing tracked files such as:

- `godot_project/scripts/render/PerspectiveProbePipeline.gd`
- `godot_project/shaders/visible_surface_builder.glsl`
- `godot_project/shaders/perspective_probe_march_pack.glsl`
- `godot_project/shaders/cell_owner_product.glsl`

Run:

```bash
graft skeleton godot_project/shaders/cell_owner_product.glsl .
```

Observed:

```text
graft skeleton — godot_project/shaders/cell_owner_product.glsl

no definitions indexed for this file
```

The files are present in git. The issue is that `src/graph/source-files.ts` only enumerates files claimed by `languageOf`, `genericLangOf`, or `containerLangOf`; neither `.gd` nor `.glsl` is claimed by the current depth/generic/container registries.

Local verification against the current registry:

```text
unsupported=.gd,.glsl
supported_has_gd=false
supported_has_glsl=false
```

## Expected behavior

At minimum, `.gd` and `.glsl` files should be source-indexed with file nodes and useful definition nodes so `skeleton`, `grep`, `ask`, and MCP `graft_file_api` can route agents to real code instead of reporting an empty API surface.

A reasonable first pass would support:

- `.gd`: `class_name`, `func`, `static func`, `class`, `signal`, `enum`, `const`, `var`, `@export var`
- `.glsl`: `struct`, function definitions, `layout/uniform` declarations, `const`, preprocessor defines

## Why this matters

In Godot projects, renderer ownership often lives in `.gd` orchestration files plus `.glsl` shader producers/consumers. Without these extensions, Graft can project failure-log/code references that mention the files, but the wiring graph has no definitions for the actual source files, so structural code tools are blind on the most important surfaces.

Frankie-Xu:
`tree-sitter-wasm@1.1.6` already includes `gdscript`, `glsl`, and `gdshader`.

`.gd` is already being added in #299 (breadth-tier row + tags query + `preload`/`load` imports). I am not opening a second GDScript PR.

[...]

PR #299, opened 2026-09-04
GDScript support, plus two fixes: Windows blast radius and sticky --only-dir

wesleytakatsu:
Four changes, found while wiring graft into a 1,619-file Godot + Go game repo. Two are bug fixes that stand on their own; two add GDScript. Happy to split this into separate PRs if you'd rather review them apart — the files don't overlap.

Every claim below is measured on that repo (1,273 files indexed after scoping, 846 of them `.gd`).

---

### 1. `fix(claude): match the edited file on Windows, so blast radius fires` (c4f39b9)

The post-edit hook forwards `tool_input.file_path` exactly as the host wrote it. On Windows that is a backslash path, while every `node.path` in the graph is posix by construction. `nodeIdsInFile` compared them with `filePath.endsWith("/" + n.path)`, which cannot match across separators — so the lookup returned an empty set and `formatBlastRadius` returned `null`.

**On every edit, in every language, with no error anywhere.** Editing a file with eight dependents cost the hook's full ~9s and printed nothing. Normalizing the incoming path turns that same edit into the eight-line radius the hook was written to produce.

This one is independent of the GDScript work and affects every Windows user today.

### 2. `fix(build): make --only-dir survive a no-flag rebuild` (4408ba9)

`--only-dir`'s whitelist was recorded only in the graph fingerprint, and the fingerprint is rewritten by every build. `sync-run.ts` shells a bare `graft build .` that never sees a CLI flag, so the first post-edit rebuild after a `--only-dir` build silently re-indexed the whole repo.

Where the excluded directories are vendored reference code, that is not cosmetic: on this repo an unscoped index put four of six hits for "player movement" in sample FPS templates and none in the actual game. One file edit undid the scope, with nothing in the output saying so.

The list now persists in `graft/.cache/only-dirs.json` and is read back when no flag is passed. Keeping it under `graft/` rather than in `.graft/config.json` preserves the property `test/graph-only-dir.test.ts` guards — a `--only-dir` build still leaves no trace in the repo being indexed, and that test passes unchanged. `--all-dirs` clears it, since a sticky whitelist otherwise has no escape hatch short of deleting the graph.

Measured: `--only-dir` build indexes 1,273 files; a subsequent bare `graft build .` used to return to 1,619 and now stays at 1,273.

### 3. `feat(graph): add GDScript to the breadth tier` (c06d6ad)

Godot projects are a large unserved corpus — a mid-size game is a four-figure count of `.gd` files, and today every one is skipped, so graft indexes the C#/C++/Go edges of a Godot repo and none of the game logic. `tree-sitter-wasm` already ships the gdscript grammar, so this is one registry row plus a tags query.

The query is deliberately narrow on variables. A bare `(variable_statement)` also matches every local inside a function body — on this repo that was ~36k extra nodes (66.5k vs 30.5k total), which buries the symbols an agent is searching for. Scoping const/var to `(source ...)` and `(class_body ...)` keeps the class-level API and drops the locals, at the cost of ~1.4k local-to-local call edges.

Signals get a name of their own so `graft callers <signal>` finds every emit and connect site — that is the wiring an agent most needs in a Godot codebase, and the thing grep is worst at.

### 4. `feat(graph): GDScript preload/load become import edges` (d1cb17c)

GDScript has no import statement; a script reaches another through `preload("res://...")` or `load(...)`. Without those, `.gd` files had symbol-level call edges and no file-level edges at all, so anything reading incoming edges on a file — blast radius, `graft callers <file>` — had nothing to report for an entire language.

`res://` is the Godot project root, which sits at an unknown depth under the repo root and may appear more than once, so the specifier is matched as a path suffix over an index of every directory boundary — the same approach already used for Java, C and PHP. Only a unique hit resolves; an ambiguous or absent one keeps the raw specifier rather than pointing the edge at a same-named file in a different project.

`@reference.import` is handled generically rather than as a GDScript special case: any breadth-tier grammar can now mark a specifier and get file-level import edges, and the edge is attributed to the file, not to whatever definition encloses the call.

Measured: 3,301 import edges, 3,252 resolved to a file (the 49 left are runtime `load()` of computed paths); total edges 29,166 → 31,204.

---

### Tests

New regression tests in `test/generic-extract.test.ts` (GDScript symbols, and preload/load import edges including the dedupe and the dangling-specifier case), `test/graph-only-dir.test.ts` (whitelist survives a no-flag rebuild; `--all-dirs` clears it), and `test/claude-format.test.ts` (a Windows path resolves to the same nodes).

`npx tsx --test` over `test/graph-*.test.ts` plus `generic-extract`, `covers`, `regression-single-repo`, `supported-extensions`, `util-state`, `context-only-dir`: **377 pass, 0 fail, 5 skipped.**

Three tests in `claude-format` and four in `claude-hooks`/`session-metrics` fail on my machine, all in the token/dollar accounting paths. They fail identically on a clean checkout of `main` — I verified with `git stash` — so they are not from this branch.

### Docs

`README.md`'s breadth-tier list gains GDScript (23 → 24 languages).

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ compiler-grade layer — all `$0` and deterministic (no model, no key):
 - **Broad** — symbols (functions, classes, methods, types, …) plus name-resolved
   call edges via a generic tree-sitter extractor, one grammar per language:
   **Rust, C, C++, C#, Ruby, Scala, Elixir, Solidity,
-  OCaml, Zig, Dart, Clojure, Nix, Lua**.
+  OCaml, Zig, Dart, Clojure, Nix, Lua, GDScript**.
 
 - **Compiler-grade edges (opt-in)** — `graft build --lsp` adds precise
   `lsp_resolved` call edges (member calls the static pass can't type) when a
@@ -212,7 +212,7 @@ compiler-grade layer — all `$0` and deterministic (no model, no key):
   **gopls** (Go), **pyright** (Python), **typescript-language-server** (TS/JS).
   It's best-effort — with no server installed the graph is unchanged.
 
-Twenty-three languages in total. A file whose language isn't listed is skipped, not
+Twenty-four languages in total. A file whose language isn't listed is skipped, not
 indexed. Adding a broad-tier language is a small contribution — see
 [CREDITS.md](CREDITS.md) for the folks who added the current set.
 
--- a/src/claude/format.ts
+++ b/src/claude/format.ts
@@ -1,4 +1,5 @@
 import { basename } from 'node:path';
+import { toPosixPath } from '../util/paths.js';
 import type { Stats, SessionState } from './state.js';
 import type { GraphV1, EdgeV1 } from '../graph/types.js';
 // The injection gate reuses the federation floors rather than inventing its own:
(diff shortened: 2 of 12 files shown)


Frankie-Xu:
`tree-sitter-wasm@1.1.6` does ship `gdscript` (and `glsl` / `gdshader`). This PR is the right place for the `.gd` half of #293 — I will not open a parallel GDScript registry-row PR.

#293 also asks for `.glsl`. That grammar is in the same wasm bundle and is not in this diff. Happy to take that as a follow-up so this PR does not grow.

The Windows blast-radius path match and sticky `--only-dir` look independently useful; splitting them as you offered would make review easier, but not blocking on that from here.

PR #300, opened 2026-09-04
feat(ingest): allow --include-dir to opt into named dot-directories (#221)

### LINKED ISSUE #221: Feature request: opt-in inclusion of hidden directories in the repo-root walk
**Summary.** `graft build` at a repo root silently skips hidden directories, so codebases whose primary source lives under a dot-directory get a near-empty graph. Passing the hidden directory *explicitly* works (`graft build .kb` parses everything), which shows the parser handles it fine — the exclusion is purely in the walk. Requesting an opt-in include mechanism so a single repo-root graph can cover chosen dot-directories.

**Environment.** graft 0.13.0 (npm), Node 24, macOS.

**Reproduction.** Our repo ([amiable-dev/chancery](https://github.com/amiable-dev/chancery)) keeps its engine — a ~2,400-line CLI, libraries, and 20 test suites, all JavaScript — under `.kb/`, with only auxiliary scripts outside it:

```
$ graft build
✓ wiring: 123 nodes (96 function, 21 method, 5 file, 1 class), 266 edges, 5 cards
  parsed: 5 of 5 files
```

Five files — all incidental site/eval scripts. Roughly 90% of the repo's code is invisible to the graph. Copy the engine to a visible path (or pass the hidden dir explicitly) and the same content parses fully:

```
$ graft build .kb
✓ wiring: 311 nodes (265 function, 46 file), 946 edges, 46 cards
  parsed: 46 of 46 files
```

**Why per-directory builds aren't a workaround.** `graft build .kb` produces a *separate, disjoint* graph: its own `graft/` cache inside the target (which for us would leak into our npm tarball — `.kb` is our package allowlist), no cross-links to the root graph, and none of the `graft init` wiring (the MCP server, skill, and hooks all point at the repo-root graph). Agents get either the empty root graph or an unwired side graph.

**Suggested shapes** (any of these would solve it):

1. `graft build --include <paths...>` — explicit extra roots merged into the repo-root walk (mirrors your existing `--extensions` opt-in style).
2. A config key (e.g. in the repo's graft config) listing hidden paths to traverse, so `graft build` and the post-edit hooks agree without repeating flags.
3. Honor a `!`-negation for hidden paths if the walker already consults ignore files.

**Prior art for the trap:** GitHub CodeQL's JS extractor also excludes dot-directories by default and offers `paths:` config to opt back in — engines-under-dot-dirs is a real layout (ours governs the repo it lives in, hence the hidden prefix).

Happy to test a branch. Thanks for Graft — the tier-1 determinism and "lexical proposes, graph disposes" retrieval are exactly the right instincts, which is why we'd like the graph to see our code.

Frankie-Xu:
## Summary
- Default is unchanged: every dot-directory (`.kb`, `.github`, `.vscode`, …) is still skipped unless named.
- `graft build --include-dir .kb` is the same persisted switch as `--include-dir build`. The walk descends into that directory and its nodes land on the **repo-root graph** the MCP server and hooks already point at — not a disjoint `graft build .kb` side graph.
- `.git` is never overridable. `--include-dir .git` is rejected with a non-zero exit and is not persisted.
- Still a bare directory name (no `/` or `\`). Git's ignore rules stay authoritative; the flag only lifts graft's own skip.

Closes #221

## Test plan
- [x] No flag / no persisted `includeDirs`: `.kb`, `.github`, `.vscode` stay skipped
- [x] `graft build --include-dir .kb`: `.kb` nodes on the same root graph as `main.ts`; `.github` still skipped
- [x] Follow-up no-flag `graft build` still includes `.kb` (`.graft/config.json` persist); fingerprint probe is clean
- [x] `--include-dir .git` exits 1 and does not persist
- [x] `--include-dir` still rejects paths (`foo/bar`, `foo\bar`)
- [x] Git-visible `.kb` is walked once included; a gitignored `.kb/` stays excluded
- [x] `npx tsc -p tsconfig.json --noEmit`
- [x] `npm test` — 1224 pass, 0 fail

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- **`graft build --include-dir` can opt a named hidden directory into the
+  repo-root walk** (#221). Default is unchanged: `.github`, `.vscode`, `.kb`
+  and every other dot-directory stay skipped. Pass `--include-dir .kb` (a
+  bare name, same switch as `build/` / `vendor/`) and the walk descends into
+  that directory so its nodes land on the same root graph the MCP server and
+  hooks already point at. The choice persists in `.graft/config.json`, so a
+  later no-flag build still includes it. `.git` is never overridable — the
+  CLI rejects `--include-dir .git` with a non-zero exit. Git's ignore rules
+  stay authoritative; the flag only lifts graft's own skip.
+
 ## 0.18.0
 
 ### Added
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -44,6 +44,7 @@ import { formatNonInteractiveHelp, formatPlan, runPicker } from "./cli-picker.js
 import { homedir } from "node:os";
 import { formatUpgradeReport, formatVersionReport, getNpmViewVersion, readCurrentVersion, runUpgrade } from "./cli-meta.js";
 import { patchBuildConfig, type BuildConfig } from "./util/state.js";
+import { NEVER_INCLUDE_DIRS } from "./ingest/fs.js";
 import { normalizePathPrefix } from "./util/paths.js";
 import { latestSession, formatSessionStats, sessionInputRate } from "./claude/session-metrics.js";
 import { setInputRate } from "./context/savings.js";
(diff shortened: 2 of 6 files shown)


Frankie-Xu:
Based on current main. CI is green. Ready for review.

Issue #298, opened 2026-09-04
graft build --deep: hardcoded maxTokens=8192 truncates tool-call JSON, and file-node summaries discarded due to id-decoration mismatch


krajtar:
## Summary

Running `graft build --deep` (v0.16.0) with `GRAFT_PROVIDER=litellm` against an OpenAI-compatible proxy (LiteLLM fronting a vLLM backend) fails to complete the meaning tier for a subset of files, deterministically across repeated runs. Two distinct bugs were found in the deep pass; both are in code paths shared with the `openai` provider.

## Bug 1: Hardcoded `maxTokens: 8192` truncates large tool-call payloads

`dist/ai/crux.js` (`describeFile`) and `dist/ai/synthesize.js` hardcode `maxTokens: 8192`. Files with many symbols — and large synthesis batches — need more output tokens than that for the tool-call JSON. The backend cuts the generation off at the cap, the tool-call arguments come back unparseable (or literally `{}`), and `finish_reason` is `length`. Graft then reports:

```
✗ path/to/big_file.py: model returned no usable symbol summaries [truncated, finish_reason=length]
```

and synthesis batches silently come back `0 nodes, 0 links`.

The backend in question accepts much larger output caps, and raising the hardcoded value to 32768 fixed both symptom classes in our testing.

**Suggestion:** make the per-call output-token cap configurable (env var, provider option, or auto-detection), or retry with a higher cap / a smaller target batch when a response arrives with `finish_reason=length`.

## Bug 2: File-node summaries silently discarded due to id-decoration mismatch

In the crux pass, `userContent()` renders each target as:

```
- id=${n.id} | ${n.kind} | lines L${startLine}-L${endLine}
```

and the system prompt instructs the model to return each entry "using that id verbatim". For **file-kind nodes**, `node.id` is the bare path, but models that follow the instruction echo the full decorated line:

```json
{"symbols": [{"id": "src/some/dir/file.py | file | lines L1-L243", "summary": "…", "crux_start": 158, "crux_end": 184}]}
```

`collectFileCrux` / the enrich loop then match results with `results.get(node.id)` — the **bare** path — which never matches the decorated echo. The perfectly valid entry is discarded, `applied === 0`, and the file is reported with the *fallback* miss label:

```
✗ src/some/dir/file.py: model returned no usable symbol summaries [empty-parsed, finish_reason=null]
```

That error message is misleading in two ways:

[...]

PR #301, opened 2026-09-05
fix(deep): match file-node crux ids when the model echoes the decorated target (#298)

Frankie-Xu:
## Summary
- Bug 2 of #298: `userContent` paints each target as `id=${n.id} | kind | lines Lx-Ly`. File-node ids are the bare path, so models that echo that decoration never hit `results.get(node.id)`, `applied === 0`, and a valid summary is reported as `empty-parsed`.
- Strip on `" | "` (node ids never contain it) in `parseResults`, and again in `collectFileCrux` against the requested id set. Unmatched entries are discarded — never applied by list order. Empty / mismatched replies stay `pending` (#177).
- When the model returned summaries that match no requested id, the error is now `id-mismatch` instead of the `empty-parsed` / `finish_reason=null` fallback.
- Does **not** bump `maxTokens` from 8192 to 32768.

Partially addresses #298 (bug 2). Bug 1 (tool-call JSON truncated at 8192) stays with #269 (recover truncated JSON) and #274 (chunk symbol-dense crux calls).

## Overlap with other Drafts
- Draft #270 peels echoed TARGET-line ids in `parseResults` (#259, including function ids with signatures). This PR is the narrower file-node decoration case plus wanted-set matching in `collectFileCrux`. Not merged together.
- Draft #274 chunks `describeFile`; it does not touch `parseResults`. This PR does not chunk.

Suggested order: this and #274 have independent hunks. #270 also edits `parseResults` — rebase whichever lands second.

## Test plan
- [x] Decorated file-node id `path | file | lines L1-L243` applies to the bare path
- [x] `path#Symbol` still matches when echoed bare
- [x] Unmatched ids are not applied by order; nodes stay `pending`, not cached `ready`
- [x] Error names `id-mismatch` / "matched no requested id"
- [x] `maxTokens` remains 8192 (asserted on the crux request)

### DIFF
--- a/src/ai/crux.ts
+++ b/src/ai/crux.ts
@@ -47,7 +47,12 @@ export interface CruxSummarizer {
 }
 
 /** Why a crux call produced no usable summaries (#235). */
-export type CruxMissKind = "empty-toolCalls" | "unparseable" | "truncated" | "empty-parsed";
+export type CruxMissKind =
+  | "empty-toolCalls"
+  | "unparseable"
+  | "truncated"
+  | "empty-parsed"
+  | "id-mismatch";
 
 export interface CruxMiss {
   kind: CruxMissKind;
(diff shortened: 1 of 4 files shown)


Frankie-Xu:
Superseded by #332 for the decorated-id miss (#298 bug 2). Leaving #270 open — that is echoed source-line payloads, not target ids.

PR #332, opened 2026-09-09
Tolerate a model echoing the whole target line as the symbol id

### LINKED ISSUE #327: Graft build --deep fails with Grok (with fix)
# `graft build --deep` silently drops valid summaries when the model echoes the target-list suffix into the returned `id`

## Summary

When the LLM used for the deep/meaning pass returns a symbol's `id` field with extra text appended (e.g. `"<id> | file | lines L1-L139"` instead of the bare `"<id>"`), `parseResults()` in `src/ai/crux.js` (compiled from `src/ai/crux.ts`) stores the result under the wrong key, so `enrich.js`'s node lookup (`results.get(node.id)`) never finds it. The file is then reported as a total miss —

```
✗ <path>: model returned no usable symbol summaries [empty-parsed, finish_reason=null]
```

— even though the model produced a complete, well-formed summary. The error message is misleading: nothing was empty. A good summary was generated and then silently discarded because of a strict-equality id mismatch.

## Environment

- `@nanonets/graft` 0.16.0
- Node v24.14.0, Windows 10 (build 26200)
- `GRAFT_PROVIDER=openai` (OpenAI-compatible wire format)
- `GRAFT_BASE_URL=https://api.x.ai/v1`
- `GRAFT_MODEL=grok-4.6` (reproduced on both `grok-4.6` and an earlier Grok model — model choice was not the variable)

## Reproduction

Run `graft build --deep` on a repo containing a Python file whose only extracted target is the whole-file node (e.g. a short procedural script with no top-level functions Graft's extractor treats as separate symbols). Example failure, reproduced consistently across 5 files in one project:

```
✗ society/engineering/house-standard-calc-scripts/calc_v3.py: model returned no usable symbol summaries [empty-parsed, finish_reason=null]
✗ society/engineering/house-standard-calc-scripts/summary_gen_v3b.py: model returned no usable symbol summaries [empty-parsed, finish_reason=null]
✗ society/engineering/house-standard-calc-scripts/mage_production_calc.py: model returned no usable symbol summaries [empty-parsed, finish_reason=null]
✗ society/engineering/house-standard-calc-scripts/report_gen_v3.py: model returned no usable symbol summaries [empty-parsed, finish_reason=null]
✗ society/engineering/house-standard-calc-scripts/summary_gen_v3.py: model returned no usable symbol summaries [empty-parsed, finish_reason=null]
```

Failure did not correlate with file size (files ranged 6,078–99,865 chars) or with the model version, which ruled out both the `MAX_CODE_CHARS` truncation path and a reasoning-token-budget theory before the actual cause was found via xAI's request/response logging.

## Root cause (confirmed via raw provider logs)

xAI's request logging shows the model receiving Graft's target line exactly as built in `crux.ts`:

```
TARGETS (1 — return all 1, one entry per id):
- id=society/engineering/house-standard-calc-scripts/summary_gen_v3b.py | file | lines L1-L139
```

and returning a `record_symbols` tool call with a **complete, correct summary**, but with the full target line — including the `| file | lines L1-L139` suffix — copied into the `id` field:

```json
{
  "symbols": [
    {
      "id": "society/engineering/house-standard-calc-scripts/summary_gen_v3b.py | file | lines L1-L139",
      "summary": "Builds a markdown roll-up of heaviest-scenario material-category totals (CSEB grades, mortar, infill, rods, glass, superwood, roman concrete, gravel, shingles, wool) per module size and climate from results_v3.json.",
      "crux_start": 0,
      "crux_end": 0
    }
  ]
}
```

The prompt instruction — *"Return EXACTLY ONE entry for EVERY target id, using that id verbatim"* — is ambiguous about where the id substring ends within the target line (`id=<VALUE> | <kind> | lines L<start>-L<end>`). Grok apparently parses "verbatim" as "the whole displayed line," rather than stopping at the first ` | `.

Two places in Graft then compound this into a silent data loss instead of a visible mismatch:

1. **`parseResults()`** (`src/ai/crux.ts`, compiled to `dist/ai/crux.js:83`) takes `s.id` as-is with no validation against the known target ids:
   ```ts
   .map((s) => ({
     id: s.id,
     summary: typeof s.summary === "string" ? s.summary.trim() : "",
     ...
   }));
   ```

2. **`collectFileCrux()` / `enrich.ts`** matches results back to nodes with a strict-equality map lookup:
   ```ts
   for (const r of list) if (!results.has(r.id)) results.set(r.id, r);
   ...
   const r = results.size > 0 ? results.get(node.id) : undefined;
   if (!r || !r.summary.trim()) { /* treated identically to a real miss */ }
   ```

Since `results` has one entry keyed by the mangled id, `results.get(node.id)` (the real, bare id) returns `undefined`. `applied` stays `0` even though a usable summary exists in `results`. `enrich.ts` then hits the "got entries but nothing applied" branch and reports it via `cruxMissMessage(summarizer, "empty-parsed")` — a label that specifically means "the model gave us nothing," which isn't what happened here. The `--deep` build correctly refuses to exit 0 on this (per #127), so the failure is at least loud — but it points at the wrong cause and the underlying summary is thrown away rather than salvaged.

## Suggested fix

At minimum, `parseResults()` should tolerate the model echoing the full target-line format instead of the bare id, by truncating at the first ` | `:

```ts
.map((s) => ({
  id: s.id.includes(" | ") ? s.id.split(" | ")[0] : s.id,
  summary: typeof s.summary === "string" ? s.summary.trim() : "",
  crux_start: num(s.crux_start),
  crux_end: num(s.crux_end),
}));
```

This is safe because a real Graft node id (a file path, or `path::symbol`) never legitimately contains the literal substring `" | "` — that sequence only ever comes from the target-line's own `| <kind> | lines ...` formatting, so stripping from the first occurrence recovers exactly the intended id and nothing else.

Applying this one-line patch locally (to the compiled `dist/ai/crux.js`) resolved all 5 failures in our repro — the previously-"empty" summaries all matched their nodes and applied correctly on the next `--deep` run, with no other changes.

A more robust long-term fix might also:
- Make the TARGETS line format less ambiguous to a model reading it as "the id" (e.g. put the id on its own line, or wrap it in an unambiguous delimiter like backticks, rather than a `key=value | field | field` line where "verbatim" is easy to over-scope).
- Have `collectFileCrux` fall back to a prefix/fuzzy match (or Levenshtein-lite) against the known target ids when an exact key isn't found, logging a warning instead of silently treating a near-miss as a total miss — this would make the failure mode "degraded but recovered" instead of "discarded," and would likely help with other providers that have similar quirks, not just this one.
- Since this was very hard to diagnose from the CLI's `empty-parsed` message alone (it reads as "the model produced nothing," when in this case it produced something valid under the wrong key), consider distinguishing "matched 0 of N results to known ids" (a matching-layer problem) from "the model's tool call had 0 entries" (a generation problem) as separate `CruxMissKind`s — they have very different causes and very different fixes.

## Notes

- Reproduced with `GRAFT_PROVIDER=openai` pointed at xAI's OpenAI-compatible endpoint (`https://api.x.ai/v1`) using `grok-4.6`. We did not have a chance to test whether OpenAI/Anthropic models exhibit the same "echo the whole target line" behavior under load — it may be Grok-specific, but the underlying matching code has no defense against it regardless of which provider triggers it.
- Happy to share the full raw xAI request/response log (redacted of API key IDs) if useful for a regression test fixture.

L4XB:
Fixes #327

`graft build --deep` lists each target as `- id=<id> | <kind> | lines L<a>-L<b>` and asks the model to return the id "verbatim". Grok (reproduced by the reporter on `grok-4.6` via the OpenAI-compatible endpoint, with raw provider logs) copies the whole displayed line into the returned `id`. `parseResults()` stored the result under that mangled key, `enrich`'s `results.get(node.id)` missed, and a complete, well-formed summary was discarded and reported as `model returned no usable symbol summaries [empty-parsed, ...]`, which points at the wrong cause.

Change: `parseResults()` now runs the returned id through `normalizeTargetId()`, which cuts at the first ` | ` and drops a leading `id=`. A bare id is unchanged, so nothing changes for models that follow the instruction. The prompt wording is left as is.

Tests (`test/llm-ops.test.ts`):
- `ChatCruxSummarizer` with symbols returned as `a.ts | file | lines L1-L5`, `id=sym1` and `  sym2  ` yields results keyed `a.ts`, `sym1`, `sym2` (fails on `main`, passes with the change).
- `normalizeTargetId` unit cases, including an id that itself contains `#`.

`node --import tsx --test test/llm-ops.test.ts`: 11 pass.

### DIFF
--- a/src/ai/crux.ts
+++ b/src/ai/crux.ts
@@ -135,6 +135,21 @@ function userContent(input: FileCruxInput): string {
 }
 
 /** Normalize the tool's parsed argument object into a {@link NodeCrux} list. */
+/**
+ * The prompt lists targets as `- id=<id> | <kind> | lines L<a>-L<b>` and asks for
+ * the id "verbatim". Some models (Grok, #327) copy the whole displayed line, or
+ * the `id=` prefix, into the returned `id`; enrich then looks the result up by
+ * the bare node id, misses, and reports a complete summary as `empty-parsed`.
+ * Strip the echoed decoration so the summary lands on its node.
+ */
+export function normalizeTargetId(raw: string): string {
+  let id = raw.trim();
+  const sep = id.indexOf(" | ");
+  if (sep !== -1) id = id.slice(0, sep).trimEnd();
+  if (id.startsWith("id=")) id = id.slice(3).trim();
+  return id;
+}
+
 function parseResults(obj: { symbols?: unknown } | undefined): NodeCrux[] {
   if (!obj || !Array.isArray(obj.symbols)) return [];
   const num = (v: unknown) => (typeof v === "number" && Number.isFinite(v) ? Math.trunc(v) : 0);
@@ -142,7 +157,7 @@ function parseResults(obj: { symbols?: unknown } | undefined): NodeCrux[] {
     .map((s) => s as Record<string, unknown>)
     .filter((s) => typeof s.id === "string")
     .map((s) => ({
-      id: s.id as string,
+      id: normalizeTargetId(s.id as string),
       summary: typeof s.summary === "string" ? s.summary.trim() : "",
       crux_start: num(s.crux_start),
       crux_end: num(s.crux_end),
--- a/src/ai/crux.ts
+++ b/src/ai/crux.ts
@@ -134,30 +134,40 @@ function userContent(input: FileCruxInput): string {
   return `FILE: ${input.path}\n\n${numberLines(input.source)}\n\nTARGETS (${n} — return all ${n}, one entry per id):\n${targets}`;
 }
 
-/** Normalize the tool's parsed argument object into a {@link NodeCrux} list. */
 /**
  * The prompt lists targets as `- id=<id> | <kind> | lines L<a>-L<b>` and asks for
- * the id "verbatim". Some models (Grok, #327) copy the whole displayed line, or
- * the `id=` prefix, into the returned `id`; enrich then looks the result up by
- * the bare node id, misses, and reports a complete summary as `empty-parsed`.
- * Strip the echoed decoration so the summary lands on its node.
+ * the id "verbatim". Some models (Grok, #327; deepseek, #259) copy the whole
+ * displayed line, or the `id=` prefix, into the returned `id`; enrich then looks
+ * the result up by the bare node id, misses, and reports a complete summary as
+ * `empty-parsed`. Strip the echoed decoration so the summary lands on its node.
+ *
+ * With `expectedIds` (the ids we asked for) the peel only applies when it lands
+ * on one of them. An id we did not request keeps its raw text even if it happens
+ * to contain ` | `, so a genuine miss is still reported as a miss.
  */
-export function normalizeTargetId(raw: string): string {
-  let id = raw.trim();
-  const sep = id.indexOf(" | ");
-  if (sep !== -1) id = id.slice(0, sep).trimEnd();
-  if (id.startsWith("id=")) id = id.slice(3).trim();
-  return id;
+export function normalizeTargetId(raw: string, expectedIds?: ReadonlySet<string>): string {
+  const id = raw.trim();
+  if (expectedIds?.has(id)) return id;
+  let peeled = id;
+  const sep = peeled.indexOf(" | ");
+  if (sep !== -1) peeled = peeled.slice(0, sep).trimEnd();
+  if (peeled.startsWith("id=")) peeled = peeled.slice(3).trim();
+  if (expectedIds && !expectedIds.has(peeled)) return id;
+  return peeled;
 }
 
-function parseResults(obj: { symbols?: unknown } | undefined): NodeCrux[] {
+/** Normalize the tool's parsed argument object into a {@link NodeCrux} list. */
+function parseResults(
+  obj: { symbols?: unknown } | undefined,
+  expectedIds: ReadonlySet<string>,
+): NodeCrux[] {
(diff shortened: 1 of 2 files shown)


Frankie-Xu:
Heads-up from the #259 side: this 
ormalizeTargetId cut-at-first- |  also peels the deepseek TARGET-line shape that draft #270 is fixing (id | kind | lines Lx-Ly | signature). If this lands, #270 can drop its parseResults rewrite and keep only the enrich.ts miss naming (unusableSymbolsReason).

One conservatism gap vs #270: a garbage id that merely contains  |  (no lines L marker, not a requested id) is truncated here. Real graft ids are path#Name and do not contain  | , so it should be fine in practice — flagging in case you want an expectedIds guard later.

L4XB:
Thanks, that gap is closed in c057407: `parseResults` now passes the set of requested node ids into `normalizeTargetId`, and the peel only applies when it lands on one of them. A requested id is returned verbatim even if it contains ` | `, and an unrequested id that merely contains ` | ` (or an `id=` prefix that peels to something we never asked for) keeps its raw text, so a genuine miss is still reported as a miss. The bare `normalizeTargetId(raw)` form without the set keeps the unconditional peel for callers that have no target list.

Added the deepseek shape from #259 (`id | kind | lines Lx-Ly | signature`, with the signature itself containing ` | `) to the unit test, plus a garbage entry in the summarizer test that must survive untouched. `node --import tsx --test test/llm-ops.test.ts`: 12 passed; `tsc --noEmit` clean.

PR #335, opened 2026-09-09
Resolve TypeScript calls through named imports inside the imported module

### LINKED ISSUE #330: TypeScript external named-import calls resolve to unrelated local same-name functions
## Summary

For TypeScript/TSX, a call through a named import from an external package can be resolved to an unrelated same-named function inside the repository.

This produces a false `calls` edge rather than an undercount. `graft callers`, repo-map hotspots, and transitive blast-radius walks can therefore report that production code depends on a test helper or mock that it never imports.

Because traversal does not expose inferred confidence, these false edges are indistinguishable from extracted ones. The graph therefore cannot be trusted for repository orientation or change-impact decisions without manually verifying the underlying imports and source.

## Environment

- `@nanonets/graft` 0.16.0
- macOS arm64
- Node.js 24.17.0
- Also confirmed in current `main` at `3a85a2f40e283ac0c9b617d9a4bead18d73c3c70`

## Minimal reproduction

Create two TypeScript files:

```ts
// src/component.ts
import { useRouter } from "next/navigation";

export function Component() {
  return useRouter();
}
```

```ts
// test/next-navigation-mock.ts
export function useRouter() {
  return { push() {} };
}
```

Then run:

```sh
graft build .
graft callers useRouter --json --no-refresh
```

### Actual behavior

The local `test/next-navigation-mock.ts#useRouter` symbol receives an incoming `calls` edge from `src/component.ts#Component`, even though that file imports `useRouter` from the external `next/navigation` package.

The same pattern occurs with test-framework imports: calls to `expect()` imported from `vitest` can be attached to an unrelated unique local helper named `expect`.

### Expected behavior

The external call must not resolve to the local mock/helper. If the imported module cannot be resolved to a file in the repository, the call should remain external/unresolved or be omitted.

## Root cause

The extractor already collects TypeScript named-import bindings, including the local name, exported name, and module specifier:

- [`collectImportedSymbols()`](https://github.com/trailhq/Graft/blob/3a85a2f40e283ac0c9b617d9a4bead18d73c3c70/src/graph/extract.ts#L850-L876)
- [`collectTsImportBindings()`](https://github.com/trailhq/Graft/blob/3a85a2f40e283ac0c9b617d9a4bead18d73c3c70/src/graph/extract.ts#L1012-L1024)

That provenance is used for non-call `references`, but it is not attached when a call edge is captured. The raw call keeps only the bare callee name, member flag, and source file:

- [call-edge capture](https://github.com/trailhq/Graft/blob/3a85a2f40e283ac0c9b617d9a4bead18d73c3c70/src/graph/extract.ts#L701-L770)

The resolver then sends the bare name through `resolveName()`. If there is exactly one same-family function with that name anywhere in the repository, it becomes the target with `confidence: "inferred"`:

- [bare-call resolution](https://github.com/trailhq/Graft/blob/3a85a2f40e283ac0c9b617d9a4bead18d73c3c70/src/graph/resolve.ts#L294-L334)
- [`resolveName()` unique-global fallback](https://github.com/trailhq/Graft/blob/3a85a2f40e283ac0c9b617d9a4bead18d73c3c70/src/graph/resolve.ts#L353-L379)

This is especially misleading because traversal drops edge confidence. `EdgeHit` and the JSON/human `callers` output do not expose that the edge was inferred:

- [`EdgeHit` construction](https://github.com/trailhq/Graft/blob/3a85a2f40e283ac0c9b617d9a4bead18d73c3c70/src/graph/traverse.ts#L114-L142)
- [`callers --json` hit shape](https://github.com/trailhq/Graft/blob/3a85a2f40e283ac0c9b617d9a4bead18d73c3c70/src/graph/traverse-cli.ts#L113-L150)

## Why this matters

- A unique collision is treated as a positive match; adding a second unrelated local definition would instead make the edge disappear as ambiguous.
- Test helpers and mocks can become false production hubs.
- `--depth all` can propagate the false edge into a much larger reported blast radius.
- The normal and JSON outputs look as authoritative as extracted edges, so callers cannot filter or audit inferred results.

## Suggested fix

1. Preserve import provenance on bare call edges. If the callee identifier is a named import, carry its exported name and module specifier into `RawEdge`.
2. Resolve that call only within the resolved in-repo target module. If the specifier is external or unresolved, do not fall back to the repo-wide name index.
3. Respect lexical shadowing: a same-file local binding should win when it actually shadows the import.
4. Preserve `confidence` through `EdgeHit`, CLI JSON/human output, MCP results, and transitive walks so any remaining inferred edges are visible and filterable.

Suggested regression cases:

- external named import colliding with one local function
- in-repo named import with another same-named function elsewhere
- aliased named import
- local lexical binding shadowing an import
- `callers` output preserving `extracted` versus `inferred` confidence

## Related issue

#68 discusses opt-in inferred calls for C/C++ and explicitly requires confidence labeling. It does not cover JS/TS named-import provenance or this external-import false-positive behavior.

L4XB:
Fixes #330

## Problem

For TypeScript/TSX, a bare call to a symbol imported by name from an external package (`import { useRouter } from "next/navigation"; … useRouter()`) went through `resolveName()`'s unique-global fallback: if exactly one same-family function with that name existed anywhere in the repo (a test mock, a helper named `expect`), the call was bound to it with `confidence: "inferred"`. `graft callers`, hotspots and blast-radius walks then reported production code depending on a mock it never imports, and the edge looked exactly like an extracted one.

The extractor already collected named-import bindings (`collectImportedSymbols` / `collectTsImportBindings`, with lexical shadowing handled by `withoutShadowedImports`) and used them for `references`, but not for `calls`.

## Change

- `src/graph/extract.ts`: a bare TypeScript/TSX call whose callee is a (non-shadowed) named import is emitted with the import's module `specifier` and its **exported** name (aliases are unwrapped). Member calls and other languages are untouched.
- `src/graph/resolve.ts`: a call edge carrying a `specifier` is resolved inside that module only:
  - external or unresolved module → the edge is dropped (no repo-wide guess);
  - in-repo module that defines the name → that definition, `extracted`;
  - in-repo module without the definition (a barrel re-export) → the existing name-based fallback, so barrel imports keep their recall.
- The `ambiguousRepo()` fixture in `test/graph-traverse-cli.test.ts` now imports `shared` through a barrel: a direct `import { shared } from "./a.js"` is no longer ambiguous, it correctly resolves to `a.ts#shared` — which is the point of this change. The A6 zero-hit assertions are unchanged.

Not included: surfacing `confidence` in `EdgeHit` / `callers --json` / MCP output (point 4 of the issue) — that is output plumbing rather than resolution and would be a separate change.

## Tests

`test/graph-resolve-imported-calls.test.ts`:
- extractor: a call through `import { useRouter } from "next/navigation"` and an aliased `import { helper as h }` carry `{ name, specifier }`; a local `const useRouter = …` shadowing the import keeps the call a bare name;
- resolver: external specifier → no edge (fails on `main`, the mock got the edge); no provenance → unchanged unique-name fallback; in-repo module → `extracted` edge to that module even though a same-named mock exists; in-repo barrel without the name → name-based fallback.

`node --import tsx --test test/graph-traverse-cli.test.ts test/graph-resolve-imported-calls.test.ts test/graph-resolve-typed.test.ts`: 35 pass; `tsc --noEmit` clean. (The `savings` / `session-metrics` / `claude-format` suites fail identically on `main` in my environment — number formatting under a non-English locale — and are unrelated.)

### DIFF
--- a/src/graph/extract.ts
+++ b/src/graph/extract.ts
@@ -92,7 +92,11 @@ export interface RawEdge {
   relation: Relation;
   file: string; // the file this edge originates in (scopes name resolution)
   targetId?: string; // already-resolved target (contains)
-  specifier?: string; // module path to resolve (imports / imported-symbol references)
+  /** module path to resolve (imports / imported-symbol references). On a
+   * `calls` edge (TypeScript): the callee is a named import from this module,
+   * and `name` is its EXPORTED name; resolve.ts then confines resolution to
+   * that module instead of guessing from a unique repo-wide name match (#330). */
+  specifier?: string;
   name?: string; // symbol name to resolve (extends/implements/calls)
   viaMember?: boolean; // calls: was it `obj.foo()` (→ prefer method targets)?
   /** calls with viaMember: the receiver's resolved type name (from bindings /
@@ -722,12 +726,23 @@ function walk(node: Parser.SyntaxNode, ctx: WalkCtx, out: NodeV1[], edges: RawEd
       consumedCallee === "R6Class" || (consumedCallee === "list" && rIsMixinContainer(node));
     const callee = isConsumedRClassCall ? null : calleeName(node, ctx.lang);
     if (callee) {
+      // A bare TypeScript call through a named import carries where the callee
+      // comes from. `ctx.importedSymbols` already excludes bindings shadowed by
+      // a local declaration in scope, so `useRouter()` under a local
+      // `function useRouter` stays a bare name. Without this, resolve.ts's
+      // unique-name fallback bound a call to `useRouter` from "next/navigation"
+      // to an unrelated test mock of the same name (#330).
+      const imported =
+        !callee.viaMember && (ctx.lang === "typescript" || ctx.lang === "tsx")
+          ? ctx.importedSymbols.get(callee.name)
+          : undefined;
       const callEdge: RawEdge = {
         source: ctx.parentId,
         relation: "calls",
-        name: callee.name,
+        name: imported ? imported.name : callee.name,
         viaMember: callee.viaMember,
         file: ctx.rel,
+        ...(imported ? { specifier: imported.specifier } : {}),
         ...(callee.kinds ? { kinds: callee.kinds } : {}),
       };
       // Overloading languages: the call site's argument count, to pick the right
--- a/src/graph/resolve.ts
+++ b/src/graph/resolve.ts
@@ -317,6 +317,25 @@ export function resolveEdges(
           : e.file.endsWith(".java")
             ? ["class", "struct", "enum", "interface"]
             : ["function"]);
+      // A call through a named import (TypeScript, extract.ts) names its module.
+      // An external or unresolved module means the callee is not in this repo:
+      // drop the edge rather than let the unique-name fallback bind it to an
+      // unrelated same-named local function (a test mock, a helper named
+      // `expect`) and report it as a production dependency (#330). An in-repo
+      // module that defines the name resolves to that definition alone; one
+      // that does not (a barrel re-export) keeps the name-based fallback.
+      if (e.specifier) {
+        const targetFile = resolveImport(e.specifier, e.file, byId);
+        if (!byId.has(targetFile)) continue;
+        const inModule = (perFileName.get(targetFile)?.get(e.name!) ?? []).filter((n) =>
+          callKinds.includes(n.kind),
+        );
+        if (inModule.length === 1) {
+          add(e.source, inModule[0].id, "calls", "extracted");
+          continue;
+        }
+        if (inModule.length > 1) continue;
+      }
       let hit = resolveName(e.name!, e.file, callKinds, perFileName, globalName);
       // Python is the Java case without the `new` to mark it: `Widget()` is an
       // ordinary call node, so a constructor edge dies against the function-only
--- a/test/graph-resolve-imported-calls.test.ts
+++ b/test/graph-resolve-imported-calls.test.ts
@@ -1,0 +1,97 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { extractFile } from "../src/graph/extract.js";
+import { resolveEdges } from "../src/graph/resolve.js";
+import type { NodeV1 } from "../src/graph/types.js";
+
+// #330: a TypeScript call through a named import must not resolve to an
+// unrelated same-named local function when the import is external or points at
+// a different in-repo module.
+
+function n(id: string, kind: NodeV1["kind"]): NodeV1 {
+  const post = id.includes("#") ? id.split("#")[1] : id;
+  const name = post.split(".").pop()!;
+  return { id, name, kind, path: id.split("#")[0], span: "L1-L1", signature: null,
+    exported: true, origin: "ast", body_hash: "h", summary_state: "pending", summary: null, crux: null } as NodeV1;
+}
+
+test("extract: a bare call through a named import carries the module and exported name", () => {
+  const { rawEdges: edges } = extractFile(
+    "src/component.ts",
+    [
+      'import { useRouter } from "next/navigation";',
+      'import { helper as h } from "./lib/util";',
+      "export function Component() {",
+      "  h();",
+      "  return useRouter();",
+      "}",
+    ].join("\n"),
+    "typescript",
+  );
+  const calls = edges.filter((e) => e.relation === "calls" && e.source === "src/component.ts#Component");
+  assert.deepEqual(
+    calls.map((e) => ({ name: e.name, specifier: e.specifier })).sort((a, b) => a.name!.localeCompare(b.name!)),
+    [
+      { name: "helper", specifier: "./lib/util" },
+      { name: "useRouter", specifier: "next/navigation" },
+    ],
+  );
+});
+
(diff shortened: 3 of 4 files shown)


0xSamuraiDev:
Validated head `1e62f87a018c9fecbe331791e1f1cdc9f2450380` against base `3a85a2f40e283ac0c9b617d9a4bead18d73c3c70` on Node 24.19.0, macOS arm64.

The fix restores callers through an aliased named import after a cold graph build, including when another module exports a function with the same name and arity. The real `graft_trace_calls` MCP handler reports the caller for the imported module and does not attach it to the unrelated module.

- Base: **3/3 new integration cases fail**, each with `no indexed callers` after a successful fresh build.
- PR head: **3/3 pass** (JavaScript, TypeScript and TSX).
- The PR's three targeted test files: **35 pass**.
- Adjacent reference, binding and MCP-tool tests: **42 pass**.
- `tsc --noEmit -p tsconfig.json`: passes.

I think the small test below adds useful coverage: the new unit tests check extraction and resolution separately, while this exercises file-type dispatch, extraction, module resolution, graph persistence and the MCP tool response together. It uses real source files and a cold build, with no mocked graph nodes or edges. This is tool-handler integration coverage, not a test of the stdio server lifecycle.

<details>
<summary>Suggested regression test: test/graph-imported-call-mcp.test.ts</summary>

```ts
import { test } from 'node:test';
import assert from 'node:assert/strict';
import { mkdirSync, rmSync, writeFileSync } from 'node:fs';
import { join } from 'node:path';
import { buildGraph } from '../src/graph/build.js';
import { callTool } from '../src/mcp/tools.js';
import { tmpRepo } from './helpers.js';

for (const ext of ['js', 'ts', 'tsx']) {
  test(`${ext}: a persisted aliased call reaches the MCP caller tool in the correct module`, async (t) => {
    const root = tmpRepo('imported-call-mcp');
    t.after(() => rmSync(root, { recursive: true, force: true }));
    mkdirSync(join(root, 'src'));
    mkdirSync(join(root, 'test'));
    const target = `src/limits.${ext}`;
    const collision = `test/mock.${ext}`;
    writeFileSync(join(root, target),
      'export function withinLimit(total, limit) { return total <= limit; }\n');
    writeFileSync(join(root, collision),
      'export function withinLimit(total, limit) { return false; }\n');
    writeFileSync(join(root, `src/consumer.${ext}`),
      "import { withinLimit as budgetAllows } from './limits.js';\n" +
      'export function checkout() { return budgetAllows(1, 2); }\n');

    // Exercise extraction, resolution, persistence and the real MCP tool handler;
    // do not construct RawEdge/NodeV1 fixtures or reuse an existing graph.
    const built = await buildGraph(root, { reuse: false });
    assert.deepEqual(built.errors, []);
    const callers = await callTool(root, 'graft_trace_calls', {
      symbol: 'withinLimit', in: target,
    });
    assert.equal(callers.isError, false, callers.text);
    assert.ok(callers.text.includes(`calls ← checkout (src/consumer.${ext}:`), callers.text);

    // The same exported name in another module must not acquire the caller.
    const unrelated = await callTool(root, 'graft_trace_calls', {
      symbol: 'withinLimit', in: collision,
    });
    assert.equal(unrelated.isError, false, unrelated.text);
    assert.doesNotMatch(unrelated.text, /calls ← checkout/);
  });
}
```

Run with:

```sh
node --import tsx --test test/graph-imported-call-mcp.test.ts
```

</details>

Existing checks run:

```sh
node --import tsx --test test/graph-traverse-cli.test.ts test/graph-resolve-imported-calls.test.ts test/graph-resolve-typed.test.ts
node --import tsx --test test/graph-references.test.ts test/graph-bindings.test.ts test/mcp-tools.test.ts
```

L4XB:
Thanks for running it independently — and for the suggested test. Added as `test/graph-imported-call-mcp.test.ts` in 5639219.

I kept your fixture idea (aliased named import plus a collision module exporting the same name and arity) and the three file types, and adapted the mechanics to this file's neighbours: the cold build goes through `execFileSync(process.execPath, ['--import','tsx','src/cli.ts','build', dir])`, the way `builtRepo()` in `test/mcp-tools.test.ts` does, rather than importing `buildGraph` — same coldness, one less way for the test to diverge from what a user runs.

Verified both directions:

- **Red on `3a85a2f`** (your base): all three fail, each with `no indexed callers — the graph has no incoming call/reference edges for this symbol as written`. Exactly the reported symptom, so the cell fails for the right reason rather than on a rendering detail.
- **Green on `5639219`**: 3/3.
- Your adjacent set plus the new file: `graph-traverse-cli`, `graph-resolve-imported-calls`, `graph-resolve-typed`, `mcp-tools`, `graph-imported-call-mcp` — **62 pass**.
- `tsc --noEmit -p tsconfig.json`: clean.

The collision module is the part worth keeping, and you were right to build the suggestion around it: with a single candidate, a resolver that ignores the import specifier still lands on the right symbol by luck, so the test would pass on the broken code.

One deliberate difference from your draft: I assert on the *caller's* name in the `graft_trace_calls` output and then assert the unrelated module's response does **not** contain it, rather than matching `/calls ← checkout/`. The arrow rendering is the handler's presentation and could reasonably change; the two properties I care about are that the edge exists and that it is attached to the imported module, not the same-named one.

Issue #323, opened 2026-09-09
0.13.0+ won't start on Windows without a C toolchain: tree-sitter-kotlin@0.3.8 ships no prebuilds, so require() throws before any command runs


back1ply:
## What happens

`graft` 0.13.0 and later cannot start on a Windows machine without a C toolchain. It fails at module load, before any command runs — so `init`, `build`, `--help` and even `--dry-run` all die the same way.

```
Error: No native build was found for platform=win32 arch=x64 runtime=node abi=137 uv=1 libc=glibc node=24.18.0
    loaded from: ...\node_modules\@nanonets\graft\node_modules\tree-sitter-kotlin

    at load.resolve.load.path (node-gyp-build\node-gyp-build.js:60:9)
    at load (node-gyp-build\node-gyp-build.js:22:30)
    at Object.<anonymous> (tree-sitter-kotlin\bindings\node\index.js:3:43)
```

## Repro

```
npx -y @nanonets/graft@latest init --dry-run --agents claude
```

Exit code 1, nothing written.

## Environment

| | |
|---|---|
| OS | Windows 11 Pro 10.0.26200 (x64) |
| Node | v24.18.0 (ABI 137), via nvm-windows |
| npm | 12.0.2 |
| graft | 0.16.0 (also 0.15.0, 0.13.0) |
| C toolchain | none (no Visual Studio Build Tools) |

## Cause

`tree-sitter-kotlin@0.3.8` ships **no prebuilt binaries at all**. The published tarball contains only sources:

```
LICENSE  README.md  binding.gyp  bindings/  grammar.js  package.json  queries/  src/
```

No `prebuilds/`, no `build/`. So `node-gyp-build` has nothing to load and must compile at install time. On a machine with no compiler that compile does not happen, and the failure only surfaces later at `require()`.

This contrasts with the other grammars — the CHANGELOG notes for `@davisvaughan/tree-sitter-r` that it "ships prebuilds for every supported platform", which is why those are fine.

Not strictly Windows-specific: the package ships no prebuilds for any platform. It just doesn't bite on Linux/macOS, where a compiler is usually present.

## Bisect

`tree-sitter-kotlin` enters the dependency tree in 0.13.0, matching the CHANGELOG entry "Kotlin moves to full-fidelity extraction":

| version | `tree-sitter-kotlin` | starts on Windows |
|---|---|---|
| 0.10.1 | no | yes |
| 0.12.1 | no | **yes** |
| 0.13.0 | yes | no |
[...]

PR #337, opened 2026-09-10
fix: one unloadable grammar no longer takes down the whole CLI (#323)

Ercaner1988:
Implements suggestion (3) from #323, plus (2) for Kotlin.

## What happens today

`src/graph/extract.ts` imports all nine depth-tier grammars at the top of the module. They are native (node-gyp) modules, so any one of them failing to load takes the CLI down at import time, before argv is read — with a `node-gyp-build` stack trace that never says "graft".

On current `main`, with `tree-sitter-kotlin` made unloadable:

```console
$ graft --version
node:internal/modules/run_main:107
Error: No native build was found for platform=win32 arch=x64 runtime=node abi=137 uv=1 node=24.16.0
    loaded from: tree-sitter-kotlin

$ graft --help                          # same
$ graft init --dry-run --agents claude   # same — the repro in the issue
$ graft build <repo with no .kt in it>   # same
```

I have hit this twice from unrelated causes: Kotlin, as in this issue, and `tree-sitter-r` under Bun, where the binding asks for a prebuild filename that the scoped package does not ship (r-lib/tree-sitter-r#205, one-line fix in r-lib/tree-sitter-r#206). Both were one line to fix once found, and neither was findable from the error, which is the part fix (3) changes.

## The isolation (fix 3)

`createRequire` plus a try/catch per grammar, keeping whatever loads. The extension table is filtered by that, which is all the rest of the pipeline needs: `entryFor` stops claiming the language's extensions, and its files then take the paths a language graft has no grammar for takes today — the breadth tier where a generic row claims the extension, unindexed where none does. `listSourceFiles` never enumerates them, so `build.ts`'s `generic!` stays safe.

The warning is once per language, from `entryFor` — the first time a file that language would have claimed comes past. A repo with no Kotlin in it stays quiet; a Kotlin repo never indexes short in silence:

```
graft: .kt/.kts files are not parsed with their own grammar — tree-sitter-kotlin failed to load (No native build was found for platform=win32 arch=x64 runtime=node abi=137 uv=1 node=24.16.0). Reinstalling graft rebuilds it; every other language indexes as usual.
```

This is the contract your two WASM tiers already state for themselves — "an unavailable grammar is skipped rather than fatal" (container.ts), "degrades to a file node only (never throws)" (generic.ts). The native tier was the one that could not say it.

`tree-sitter` itself stays a static import: it is the shared dependency of all nine, and there is no depth tier without it.

## Kotlin's fallback row (fix 2)

If you would rather take the isolation without this, it is the one `GENERIC_LANGS` row in `generic.ts` plus the two warm-call filters in `build.ts` and `check.ts` — say so and I will split it back out.

With the isolation alone that machine starts but indexes no Kotlin at all, which on a Kotlin repo is still a bad day. `tree-sitter-wasm` ships a kotlin grammar and `src/graph/queries/kotlin.scm` has been in the tree unused since Kotlin was promoted to the depth tier, so the fallback costs one registry row.

A row whose extension the depth tier also claims is unreachable while the native grammar loads — `.java` has been exactly that all along. Both `warmGenericGrammars` calls now apply the tier precedence their own parse loops apply, so a fallback row cannot warm a WASM grammar the run will never call (already true of `.java`; a second row would have doubled it).

Measured on a three-file repo (`.ts`, `.kt`, `.rs`):

| `tree-sitter-kotlin` | result |
| --- | --- |
| loads (today) | `11 nodes, 8 edges [kotlin, rust, typescript]` |
| broken, `main` | the CLI does not start |
| broken, isolation only | `6 nodes, 4 edges [rust, typescript]`, plus the warning |
| broken, this PR | `11 nodes, 5 edges [kotlin, rust, typescript]` |

On a healthy install the graph is identical with and without the fallback row — same nodes, same edges, compared field by field. The row really is unreachable.

`graft check` agrees with the build it follows on the fallback path (`graph check: OK`, with `Repo`, `describe`, `reload` and `openRepo` all present), so this does not reopen the tier-mismatch shape of #236.

## Tests

`test/grammar-unavailable.test.ts` drives the real CLI in a child process, with `test/break-grammar-preload.cjs` standing in for the missing native build — so it runs on a runner that has a compiler, which is why CI never caught the original. The preload breaks both ways a module can be reached, `import` and `require`, so the test still fails if the static imports ever come back. Each of the three fails without the part of the change it covers.

`npm test` on this branch: 1223 tests, 1210 pass, 8 fail, 5 skipped — the same 8 that fail on `upstream/main` before it, and no others. They are `toLocaleString()` assertions that assume a `,` thousands separator, where this `tr-TR` machine renders `~100.000`. Nothing to do with this PR: they are #338, fixed separately in #345, and this branch is deliberately independent of that one.

The three tests this PR adds pass; each fails without the part of the change it covers.

One thing you may see and should not blame on this branch: `test/mcp-server.test.ts` timed out on two of my five full-suite runs (its `rpc()` helper waits a fixed 15s, then reads `undefined.result`). It is a load-sensitive deadline — those tests need ~2s to get a first response when run alone, and I measured MCP server startup identical on this branch and on `main` (1.7–2.0s across five spawns each), so nothing here made the server slower. But this PR does add a test file that spawns three short-lived CLI processes, so it adds parallel load, and that deadline has little headroom on a busy machine.

Environment: Node v24.16.0, Windows 11 x64, rebased onto `4f39d19` (0.18.0 plus the brain commits after it).

## No CHANGELOG entry

I dropped the one I had. 0.18.0 is already released, so there is no unreleased section to add to, and `de8456e` shows the release commit is what writes them. Happy to add a bullet wherever you want it.

## Not in this PR

Loading the nine grammars costs ~220 ms, paid by every command including `--version`. `GRAMMAR_MODULES` makes lazy per-language loading a small follow-up, but that is a performance change and does not belong in a fix.

### DIFF
--- a/src/graph/build.ts
+++ b/src/graph/build.ts
@@ -194,7 +194,15 @@ export async function buildGraph(
   // needs ONCE here (buildGraph is async) before the synchronous parse loop below
   // can call extractGeneric. Depth-tier (native) grammars need no warmup.
   await warmGenericGrammars(
-    new Set(files.map((f) => genericLangOf(f.abs)?.name).filter((n): n is string => !!n)),
+    // Tier precedence, same as the parse loop below: the breadth tier is reached
+    // only for files no depth grammar claims, so a fallback row (`.java`, `.kt`)
+    // must not warm a WASM grammar this build will never call.
+    new Set(
+      files
+        .filter((f) => !languageOf(f.abs))
+        .map((f) => genericLangOf(f.abs)?.name)
+        .filter((n): n is string => !!n),
+    ),
   );
   // Container tier (.vue and friends) loads its wrapper grammars the same way,
   // for the same reason: extractContainer runs inside the sync loop below.
--- a/src/graph/check.ts
+++ b/src/graph/check.ts
@@ -88,7 +88,13 @@ export async function checkGraph(
   const onlyDirs = fpOnlyDirs && fpOnlyDirs.length > 0 ? new Set(fpOnlyDirs) : undefined;
   const sourceFiles = listSourceFiles(root, outDir, undefined, onlyDirs);
   await warmGenericGrammars(
-    new Set(sourceFiles.map((f) => genericLangOf(f)?.name).filter((n): n is string => !!n)),
+    // Tier precedence, as in buildGraph and in the loop below.
+    new Set(
+      sourceFiles
+        .filter((f) => !languageOf(f))
+        .map((f) => genericLangOf(f)?.name)
+        .filter((n): n is string => !!n),
+    ),
   );
   // Container-tier grammars need the same warmup as the generic ones, for the same
   // reason: extraction below is synchronous. Missing this is what made `graft
--- a/src/graph/extract.ts
+++ b/src/graph/extract.ts
@@ -7,20 +7,80 @@
  * resolved against the whole-repo node index later, in build.ts.
  */
 import Parser from "tree-sitter";
-import TypeScript from "tree-sitter-typescript";
-import Python from "tree-sitter-python";
-import Go from "tree-sitter-go";
-import R from "tree-sitter-r";
-import Java from "tree-sitter-java";
-import Kotlin from "tree-sitter-kotlin";
-import Swift from "tree-sitter-swift";
-import PHP from "tree-sitter-php";
+import { createRequire } from "node:module";
 import { basename } from "node:path";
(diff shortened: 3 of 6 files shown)


Ercaner1988 (commented):
well, I rechecked our implementation and couldn't find any blatant error or anything.

PR #339, opened 2026-09-10
fix: include PHP type usages in reference edges

### LINKED ISSUE #324: PHP emits no type-reference edges, so `callers` reports nothing for any class or enum
## Summary

`graft callers` returns "no indexed callers" for every PHP class and enum, including ones that
are instantiated, type-hinted, or referenced by constant. The equivalent construct resolves in
Java.

## Reproduction

Three files:

```php
// src/Color.php
<?php
namespace App;

enum Color: string {
    case RED = 'red';
    case BLUE = 'blue';

    public static function warm(): array { return [self::RED]; }
}
```

```php
// src/Painter.php
<?php
namespace App;

use App\Color;

class Painter {
    public ?Color $favourite = null;

    public function paint(Color $c): string {
        return $c->value . implode('', array_map(fn($x) => $x->value, Color::warm()));
    }
}
```

```php
// src/Decorator.php
<?php
namespace App;

use App\Color as C;

class Decorator {
    public function decorate(): string { return C::BLUE->value; }
}
```

```
$ graft build .
✓ wiring: 10 nodes (3 file, 3 method, 2 class, 1 enum, 1 function), 10 edges

$ graft callers Color
Color · enum · src/Color.php:L4-L9
  no indexed callers
```

`graft callers warm` resolves the call site correctly, so the files are indexed and method call
edges work. Only the type itself has no incoming edge.

A plain class behaves the same way. Given `class Widget` in one file and `$w = new Widget();` in
another, the graph holds `calls` and `imports` edges but no `references` edge for `Widget`.

## Expected

`callers --help` says "Who calls/references a symbol". A property type, a parameter type, a
return type, an `instanceof` check and a `new` expression all reference the symbol.

## Environment

`graft 0.16.0`, installed globally from npm. Node 22, linux/arm64. Structural build,
no `--deep`, no `--lsp`.

## Where it comes from

`GENERIC_LANGS` in `dist/graph/generic.js` lists rust, java, c, cpp, ruby, scala, elixir,
solidity, ocaml, zig, dart, clojure, nix and lua. PHP is not there, so PHP runs through the depth
tier in `extract.ts`.

`dist/graph/queries/php.scm` still ships and still contains:

```scheme
(object_creation_expression
  [
    (qualified_name (name) @name)
    (variable_name (name) @name)
  ]) @reference.class
```

That query is never loaded for PHP. I confirmed this by patching the pattern and rebuilding: the
output does not change.

Java is in `GENERIC_LANGS`, so `java.scm` fires `@reference.class` on `object_creation_expression`
and `generic.js` turns it into a `references` edge. That is why an instantiated Java class
resolves and an instantiated PHP class does not.

#142 removed the stale `.php` entry from `GENERIC_LANGS` once the depth tier handled PHP. The
depth tier emits `calls`, `imports`, `contains`, `extends` and `implements`, but no type
references, so the captures in `php.scm` had no replacement.

For enums the effect is total. PHP forbids `new` on an enum, so even a working `@reference.class`
on `object_creation_expression` could never fire for one. A PHP enum cannot gain an incoming edge
at all.

## Two possible fixes, and a scope question

1. Emit `references` edges from the depth tier for PHP type positions: property types, parameter
   types, return types, `instanceof`, and `class_constant_access_expression`. Precise, and it is
   the only option that fixes enums.
2. Let `callers` fall back to `imports` edges when a symbol has no call edges. That data already
   exists and already handles aliases: `phpUseNames` strips `use X as Y` and keeps the qualified
   name, and `resolve.ts` settles it to the defining file. On a large PHP codebase I see hundreds
   of `imports` edges pointing at the file of a type whose `callers` output is empty.

Option 2 is cheaper but coarser, since `imports` is file to file rather than symbol to symbol,
and it would change `callers` behaviour for every language at once.

Following the approach in #89, I would rather ask which shape you want before opening a PR.

## Related

#34 fixed this for functions referenced without being called, but types were out of scope.
#117 reports that `ask` misses files its own graph holds one edge away, which is the same
observation about `imports` edges being underused, applied to ranking rather than to `callers`.
#144 covers other PHP constructs that are not extracted.

## Note on how this was produced

For transparency: I investigated this and wrote it up with the assistance of a coding agent.
Everything above is reproducible rather than inferred. The three files are the actual repro, and
the root cause was traced by reading the shipped `dist/`. One limit worth stating plainly: I
confirmed that editing `php.scm` changes nothing against the installed npm package, not against a
source build.

cookerpapa:
Fixes #324.

Emit references for PHP type annotations, construction, instanceof and static class accesses. Reuse existing PHP import resolution for aliases and qualified names. Exclude dynamic expressions, declarations and primitive types.

Add 13 regressions, including the reported enum case through buildGraph and callersOf.

Validation: 12 new cases fail before the fix. Afterwards, the graph suite has 342 passes, 4 skips and zero failures. Typecheck and git diff --check pass.

### DIFF
--- a/src/graph/extract.ts
+++ b/src/graph/extract.ts
@@ -699,6 +699,12 @@ function walk(node: Parser.SyntaxNode, ctx: WalkCtx, out: NodeV1[], edges: RawEd
   }
 
   // not a definition — capture calls/imports/references, then descend with the same context
+  if (ctx.lang === "php") {
+    const ref = phpTypeReference(node, ctx);
+    if (ref) {
+      edges.push({ source: ctx.parentId, relation: "references", ...ref, file: ctx.rel });
+    }
+  }
   // R's `call` node is also its ONLY vehicle for library()/require()/source() —
   // there's no separate import-statement grammar construct to key off, so isImport
   // must be checked before the generic calls path or every import call would be
@@ -968,7 +974,36 @@ function phpAttributeClassRef(
   const nameNode =
     attr.childForFieldName("name") ??
     attr.namedChildren.find((c) => c.type === "name" || c.type === "qualified_name");
-  if (!nameNode) return null;
+  return phpClassRef(nameNode, ctx);
+}
+
+/** Only class-name positions count: a variable, member name or declaration is not a type use. */
+function phpTypeReference(node: Parser.SyntaxNode, ctx: WalkCtx): { name: string; specifier?: string } | null {
+  let nameNode: Parser.SyntaxNode | null | undefined;
+  switch (node.type) {
+    case "named_type":
+    case "object_creation_expression":
+    case "class_constant_access_expression":
+      nameNode = node.namedChildren[0];
+      break;
+    case "scoped_call_expression":
+    case "scoped_property_access_expression":
+      nameNode = node.childForFieldName("scope");
+      break;
+    case "binary_expression":
+      if (node.childForFieldName("operator")?.text === "instanceof") nameNode = node.childForFieldName("right");
+      break;
+  }
+  // This grammar also parses object and relative class keywords as named_type.
+  if (nameNode?.type === "name" && /^(object|self|parent|static)$/i.test(nameNode.text)) return null;
+  return phpClassRef(nameNode, ctx);
+}
+
+function phpClassRef(
+  nameNode: Parser.SyntaxNode | null | undefined,
+  ctx: WalkCtx,
+): { name: string; specifier?: string } | null {
+  if (!nameNode || (nameNode.type !== "name" && nameNode.type !== "qualified_name")) return null;
   if (nameNode.type === "qualified_name") {
     const fqn = nameNode.text.replace(/^\\/, "");
     return { name: fqn.replace(/^.*\\/, ""), specifier: fqn };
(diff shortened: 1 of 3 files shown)


Frankie-Xu (commented):
Reviewing as someone who landed the PHP attribute (#155) and anonymous-class (#175) edges this is continuing.

**What looks right**
- Type positions are the ones that matter for #324: 
amed_type, 
ew, instanceof, scoped access (::class / static method / static property). The dedicated extract test excludes declarations, primitives, self/parent/static, and dynamic expressions (
ew $Shade(), $value->Shade(), $Shade::RED).
- phpAttributeClassRef now delegates to phpClassRef, so use aliases (including grouped use App\Types\{Color as Shade, …}) stay on the existing PHP import map. Attribute name nodes are 
ame / qualified_name, which phpTypeReference's switch does not match, so this should not double-emit #155 edges.
- Body is Fixes #324 only. It should **not** close #144. 
esolve.ts still does if (!byId.has(targetFile)) continue for a specifier whose module is not in the graph — that is exactly why vendor #[Route] / #[Deprecated] still vanish. The comment tweak on the no-specifier PHP branch is fine; it does not keep those inferred edges.

**Ask**
- statusCheckRollup on this PR only shows Blast viewer publish (green). The Node test matrix (ubuntu/windows + WASM regression) has not appeared. If that is waiting on first-time-contributor workflow approval, the extract/resolve change is still reviewable; I would not merge until 	est/graph-php.test.ts (existing in-repo attribute fixture) and the new graph-php-references.test.ts have actually run on CI.

Not requesting changes on the type-reference work itself. Vendor attributes remain a separate slice on #144.

PR #340, opened 2026-09-10
Fix/wasm parser lifetime

csm-kb:
Was working with `graft` on a large codebase (~65k files) to stress-test, noticed some failure doing so -> attempted a fix pass to move it towards functional. Feel free to leave inputs or modify/reject!

I have a subsequent PR that improves the `graft build` memory tolerance + performance using a streaming worker pool (~2.5x speedup on that codebase) that I can open afterwards.

---
#### Problem
`extractGeneric` and `parseWasm` created a `Parser` per file and never deleted it or the `Tree`. Both live in the WASM heap (2 GB cap, no finalizer), so a 65k-file C repo aborted after ~12k files and every subsequent parse failed with `Aborted()` after a few minutes.

#### Solution
Reuse one parser per grammar and delete every tree. After a real abort, fail every later parse with one message instead of one per file. `parseWasm` now returns the tree for the container tier to free, and rethrows instead of caching a symbol-less file as a success.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,23 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- **`graft build` no longer aborts partway through very large repos.** The
+  breadth-tier extractor created a web-tree-sitter parser per file and never
+  freed it or the syntax tree; those live in the WASM heap, which is capped at
+  2 GB, so on a C source base of ~65,000 files every parse after ~12,000 failed
+  with `Aborted()`. One parser per grammar is now reused and every tree is
+  deleted after extraction, in the container (`.vue`) tier as well as the
+  breadth tier. If the runtime does abort, the build reports it once and says
+  to restart, instead of once per remaining file.
+- **A parse failure is retried on the next `graft build`** instead of being
+  replayed from the extract cache until the file's bytes change, so a one-off
+  failure no longer needs a manual cache delete to clear.
+- **A `.vue` wrapper grammar that throws is now a build error** for that file,
+  not a silently symbol-less file cached as a success.
+
 ## 0.17.0
 
 ### Added
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -526,7 +526,11 @@ program
       },
       { repo: buildRoot },
     );
-    for (const e of g.errors) console.error(`✗ ${e}`);
+    // A poisoned WASM runtime or a broken grammar can fail tens of thousands of
+    // files identically; the first 20 say everything the rest would.
+    const MAX_ERRORS_SHOWN = 20;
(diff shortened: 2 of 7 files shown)


pavelsem:
Confirming this on a second language, and offering a measurement.

**Env:** graft 0.18.0, Node v24.9.0, Windows 11. Private C# monorepo, 5361 parseable files. `Aborted()` first appears at file ~4843 of 5361, and all 519 remaining `.cs` files then fail — the same cascade you describe. Each failing file parses fine in isolation, and the 662-file subtree they sit in parses 662/662 on its own, so it is aggregate growth rather than content.

**Your diagnosis reproduces exactly.** I arrived at `generic.js` independently: there is not one `.delete()` call in any of the 26 `dist/graph/*.js` modules. C# has no native binding in 0.18.0 — it loads `tree-sitter-c_sharp.wasm` — so it runs entirely on the leaking WASM path.

**The `extractGeneric` half looks separable from the rest of this PR.** `parseWasm` genuinely needs the signature change, since it returns `tree.rootNode` and callers cannot free the tree. But in `extractGeneric` the tree is function-local: last used at line 274, returned at 275. Pooling one parser per grammar and deleting the tree before the return is +8/−3:

```diff
+const parserPool = new Map();
+function pooledParser(langName, language) {
+    let p = parserPool.get(langName);
+    if (!p) { p = new tsMod.Parser(); p.setLanguage(language); parserPool.set(langName, p); }
+    return p;
+}
 export function extractGeneric(rel, source, langName) {
@@
     let tree;
     try {
-        const parser = new tsMod.Parser();
-        parser.setLanguage(entry.language);
-        tree = parser.parse((i) => source.slice(i, i + PARSE_CHUNK));
+        tree = pooledParser(langName, entry.language).parse((i) => source.slice(i, i + PARSE_CHUNK));
     }
@@
     else if (langName === "php")
         extractPhpUses(tree.rootNode, rel, rawEdges);
+    tree.delete();
     return { nodes, rawEdges };
 }
```

Results on the repo above, same commit, `--no-reuse` both times:

| | stock 0.18.0 | patched |
|---|---|---|
| parse failures | 519 | 0 |
| files indexed | 4842 / 5361 | 5361 / 5361 |
| nodes | 46 406 | 54 733 |
| edges | 11 765 | 18 318 |

Edges recover +56% against +18% for nodes — the dropped files took every edge crossing into them with them. This path is shared by every WASM-parsed language, so it is not C#-specific.

Two things it deliberately does not do, both of which your PR does: it leaves `parseWasm` alone, and it has no post-abort short-circuit, so if an abort still happened the pooled parser would be reused in a dead state. Offered only in case a smaller conflict-free slice is useful while this PR has conflicts against `main`; the `parseWasm` rework here is the more complete fix.

Caveat on pooling: it assumes single-threaded parsing, so it would need revisiting alongside the streaming worker pool you mention.

PR #345, opened 2026-09-10
fix: pin rendered numbers to one locale so graft can read its own savings footer (#338)

### LINKED ISSUE #338: Savings are under-counted ~1000x outside en-US: the footer is written with toLocaleString() and read back with a comma-only regex
graft writes its savings footer with the machine's locale and reads it back with a regex that only knows `,`. On any locale that groups thousands with something else, the number it recovers is the first group — `1.990` comes back as `1`.

`src/context/savings.ts`, fifteen lines apart:

```ts
// writer, line 121
`[graft] tokens saved ≈ ${delta.toLocaleString()} (${pct}%) — this output ≈ ` +

// reader, line 137
for (const m of text.matchAll(/\[graft\] tokens saved ≈ ([\d,]+)/g)) {
  total += Number(m[1].replace(/,/g, '')) || 0;
```

The comment above them says they live next to each other "so the two can't drift". They have not drifted from each other — they have drifted from `Intl`.

## Repro

A round trip through the pair, on a `tr-TR` machine, no arguments involved:

```console
$ node --import tsx probe.ts
default locale: tr-TR
footer written: [graft] tokens saved ≈ 1.990 (100%) — this out
read back     : 1 (should be 1990)
```

```ts
// probe.ts
import { savingsLine, sumSavingsFooters } from "./src/context/savings.js";
const line = savingsLine("x".repeat(40), { files: 2, baselineChars: 8000 })!;
console.log(sumSavingsFooters(line));   // 1, not 1990
```

And the reader against footers rendered in a few locales — this part is locale-independent, so it reproduces anywhere:

```console
en-US  "1,990"     -> 1990
tr-TR  "1.990"     -> 1
de-DE  "1.990"     -> 1
fr-FR  "1 990"     -> 1      # U+202F narrow no-break space
ru-RU  "1 990"     -> 1
hi-IN  "1,990"     -> 1990
```

So it is not a Turkish problem: de, es, it, nl, pt-BR, id, da, el group with `.`, and fr, ru, sv, pl, cs, fi group with a space. Any of those under-counts by roughly a factor of a thousand — and the smaller the true number, the closer the wrong one looks to plausible.

## What it costs

Everything downstream of `savedTokens` reads that recovered number, not the real one:

- the statusline's `~N tok saved` (`src/claude/format.ts:42`)
- its dollar value — the 0.17.0 headline (#282). A session that saved 40k tokens prices 40.
- `graft stats` (`src/claude/session-metrics.ts:202`)
- `saved_tokens_bucket` on the `session_summary` telemetry event (`src/telemetry/sessions.ts:67`), so every affected user reports a near-zero bucket and the aggregate under-states graft's payoff by however large that share of your users is.

The parser is the only one of its kind (`sumSavingsFooters` is the single `[\d,]` regex in `src/`), so the blast radius is exactly the savings accounting — nothing about the graph itself is affected.

## How it surfaces

`npm test` on such a machine fails 8 tests, on `main`, with no local changes:

| test | file |
| --- | --- |
| `tool-savings counts a REAL savings line (with the turn nudge) exactly once` | `test/claude-hooks.test.ts` |
| `formatRetrieval appends a tokens-saved line when ask reports a baseline` | `test/claude-format.test.ts` |
| `renderStatusline carries the dollar value once the session has been billed` | `test/claude-format.test.ts` |
| `renderStatusline shows tokens alone until a turn has been billed` | `test/claude-format.test.ts` |
| `savingsLine: reports saved tokens and percent when the output is smaller` | `test/savings.test.ts` |
| `formatSessionStats renders the mix, savings and last query` | `test/session-metrics.test.ts` |
| `formatSessionStats prices the saving once the session has been billed` | `test/session-metrics.test.ts` |
| `formatSessionStats omits the dollar line rather than claiming zero` | `test/session-metrics.test.ts` |

Seven of those are assertions that assume `,` — a rendering difference, not a defect. The eighth is the bug itself, and it is worth reading on its own:

```
✖ tool-savings counts a REAL savings line (with the turn nudge) exactly once
  AssertionError: Expected values to be strictly equal:
  1 !== 1990
  at test/claude-hooks.test.ts:464
```

That test feeds a real `savingsLine()` through the `tool-savings` hook and asserts the session recorded 1990. It is already the round-trip test for this; it just has no way to fail on a `,` locale.

CI cannot see any of it: GitHub runners are `en-US`, so the comma is always there.

To reproduce on Linux or macOS, `LC_ALL=de_DE.UTF-8 npm test` should do it — Node takes its default locale from the environment there. I have only verified it on Windows, where the locale comes from the OS regional setting instead and those variables have no effect.

## Fix

Your call which end to pin, but they have to agree:

1. Format the footer's number with a fixed locale. `src/cli-epilogue.ts:50` already does exactly this (`nodes.toLocaleString("en-US")`), so there is precedent in the tree, and it keeps the reader untouched.
2. Or emit the footer's number ungrouped and widen the regex to `(\d+)`. Least pretty, most robust — nothing to strip, nothing to get wrong.
3. Or keep the display locale-aware and strip every separator `Intl` might have used on the way in. This one is a trap: `1.990` is a valid en-US *decimal*, so the reader would be guessing, and it has to know about U+202F and U+00A0 too.

Whichever you pick, the seven rendering assertions want a fixed locale in the test rather than the machine's, or they will keep failing for contributors outside `en-US`.

Happy to send a patch for whichever you prefer.

## Environment

graft `main` (0.17.0, `3a85a2f`), Node v24.16.0 (ICU 78.3, full ICU), Windows 11 x64, OS locale `tr-TR`. Found while running the suite for #337; unrelated to that PR, which is why it is here and not in it.

Ercaner1988:
Fixes #338.

`savingsLine()` wrote the footer with the machine's locale; `sumSavingsFooters()` read it back with `([\d,]+)`. On a machine that groups thousands with anything but a comma, graft parsed its own footer as the first group — `1.990` came back as `1`, and every savings total downstream went with it.

## The change

One formatter, pinned, next to the reader that has to cope with its output:

```ts
export function groupDigits(n: number): string {
  return n.toLocaleString('en-US');
}
```

Five call sites now go through it. Three of them write the `[graft] tokens saved ≈ N` line that `sumSavingsFooters` parses — `context/savings.ts`, `ask/ask.ts`, `claude/format.ts`, each with its own wording of the same footer — and the other two are the displays that sit beside it, the statusline's `~N tok saved` and `graft stats`.

I pinned the two display sites as well, and that is the one judgement call in here worth flagging. They are never parsed, so locale-aware would be defensible — but graft's output is English throughout, a statusline reading `1.990` directly above a footer reading `1,990` would be its own small bug, and `src/cli-epilogue.ts:50` already pins the same way (`nodes.toLocaleString("en-US")`), which is where I took the precedent from. I left that line alone rather than reach across from an init banner into the savings module; say the word if you would rather have it routed through the same helper, or would rather keep the two displays locale-aware and pin only the parsed line.

The reader is untouched. Widening it instead was the option I would not take: `1.990` is a valid en-US *decimal*, so a reader that strips separators is guessing, and it would have to know about U+202F and U+00A0 too.

## Tests

`test/savings-locale.test.ts`. A process cannot change its own ambient locale — Node reads it from the environment on posix and from the OS on Windows — so the test moves the *default* instead: `toLocaleString()` with no argument answers as a `.`-grouping locale would, while calls that name a locale explicitly are left alone. That is exactly the difference the fix turns on, so the test fails without it and passes with it **on an en-US runner too**, which is the property CI was missing. It asserts the stand-in actually changes an unqualified call first, so the rest cannot pass for the wrong reason.

The third test renders the footer through `formatAsk()`. `ask/ask.ts` has its own wording of the parsed line, and no test rendered it at all — your blast-radius bot flagged it ("no test reaches Ask Token Savings"), and it was right: reverting just that file to `toLocaleString()` brought #338 back for every `graft ask` with the rest of the suite still green. Now it fails.

All three fail on `upstream/main`. The first two:

```
✖ #338: the savings footer round-trips on a machine that is not en-US
  actual: '[graft] tokens saved ≈ 1.990 (100%) — this output ≈ 10 tok …'
  expected: /tokens saved ≈ 1,990/
✖ #338: what graft displays groups the same way as what it parses
  actual: '… tokens saved:  ~12.345'
```

## Verified

The eight tests that fail on `upstream/main` on this machine — the ones I mentioned in #337 — are the eight this fixes. `npm test` on a `tr-TR` box:

| | tests | pass | fail |
| --- | --: | --: | --: |
| `upstream/main` | 1220 | 1207 | 8 |
| this branch | 1223 | 1218 | 0 |

(5 skipped either way.)

Seven of the eight were assertions that assumed `,`, and they now hold because the output does. The eighth, `tool-savings counts a REAL savings line`, was the round trip itself failing `1 !== 1990` — it is the test that should have caught this and could not, because the footer it feeds is rendered in the runner's own locale. It passes here now, which is the end-to-end proof: a real `savingsLine()` through the real hook, on a machine that groups with `.`, recording 1990.

One test needed the same treatment as the source. `test/savings.test.ts:52` built its expected value with `(…).toLocaleString()`, so once the footer was pinned it asserted `1.990` against a footer saying `1,990` — the mirror image of the bug. It now goes through `groupDigits` too.

Re-measured against `de8456e` (0.18.0) after rebasing onto it: same 8, same numbers.

## No CHANGELOG entry

I dropped the one I had. 0.18.0 is already released, so there is no unreleased section to add to, and `de8456e` shows the release commit is what writes them. Happy to add a bullet wherever you want it.

Environment: Node v24.16.0, Windows 11 x64, OS locale `tr-TR`.

### DIFF
--- a/src/ask/ask.ts
+++ b/src/ask/ask.ts
@@ -17,7 +17,7 @@ import { existsSync, readdirSync, readFileSync } from "node:fs";
 import { join, resolve } from "node:path";
 import matter from "gray-matter";
 import { contextDirFor } from "../context/node-file.js";
-import { withSavings, savingsFor, savingsTurnNudge, type Savings } from "../context/savings.js";
+import { withSavings, savingsFor, savingsTurnNudge, groupDigits, type Savings } from "../context/savings.js";
 import { loadGraphCached, loadAskIndexCached } from "../graph/load.js";
 import {
   assertPrefixIndexed,
@@ -1561,9 +1561,9 @@ function askSavingsLine(r: AskResult, body: string): string {
   const saved = base - pack;
   const pct = Math.round((saved / base) * 100);
   return (
-    `[graft] tokens saved ≈ ${saved.toLocaleString()} (${pct}%) — this pack ≈ ` +
-    `${pack.toLocaleString()} tok vs reading the ${r.saved.files} source file(s) whole ≈ ` +
-    `${base.toLocaleString()} tok. Estimate (baseline = those files read in full).` +
+    `[graft] tokens saved ≈ ${groupDigits(saved)} (${pct}%) — this pack ≈ ` +
+    `${groupDigits(pack)} tok vs reading the ${r.saved.files} source file(s) whole ≈ ` +
+    `${groupDigits(base)} tok. Estimate (baseline = those files read in full).` +
     savingsTurnNudge(saved)
   );
 }
--- a/src/claude/format.ts
+++ b/src/claude/format.ts
@@ -6,6 +6,7 @@ import type { GraphV1, EdgeV1 } from '../graph/types.js';
 // question in both places, and one set of calibrated numbers beats two.
 import { HIGH_FLOOR, STRONG_FLOOR } from '../ask/fuse.js';
 import { dollarsSaved, formatDollars } from '../context/price.js';
+import { groupDigits } from '../context/savings.js';
 
 const C = {
   indigo: (s: string) => `\x1b[38;2;84;111;255m${s}\x1b[0m`,
@@ -39,7 +40,7 @@ export function renderStatusline(
     // stands alone rather than carrying a rate nobody measured.
     const usd = dollarsSaved(saved, session?.inputCostMicros, session?.inputTokensBilled);
     const money = usd === null ? '' : ` · ~${formatDollars(usd)}`;
-    top.push(C.indigo(`~${saved.toLocaleString()} tok saved${money}`));
+    top.push(C.indigo(`~${groupDigits(saved)} tok saved${money}`));
   }
 
   const bottom: string[] = [];
@@ -133,9 +134,9 @@ export function formatRetrieval(ask: AskJson, cap = 5): string | null {
   const base = tokensOf(ask.saved!.baselineChars);
   const pct = Math.round((saved / base) * 100);
   return (
-    `${body}\n[graft] tokens saved ≈ ${saved.toLocaleString()} (${pct}%); this pack ≈ ` +
-    `${tokensOf(body.length).toLocaleString()} tok vs reading the ${ask.saved!.files} file(s) whole ≈ ` +
-    `${base.toLocaleString()} tok (estimate).`
+    `${body}\n[graft] tokens saved ≈ ${groupDigits(saved)} (${pct}%); this pack ≈ ` +
+    `${groupDigits(tokensOf(body.length))} tok vs reading the ${ask.saved!.files} file(s) whole ≈ ` +
+    `${groupDigits(base)} tok (estimate).`
   );
 }
 
--- a/src/claude/session-metrics.ts
+++ b/src/claude/session-metrics.ts
@@ -31,7 +31,7 @@
  */
 import { statSync } from 'node:fs';
 import { join } from 'node:path';
-import { sumSavingsFooters } from '../context/savings.js';
+import { groupDigits, sumSavingsFooters } from '../context/savings.js';
 import { dollarsSaved, formatDollars } from '../context/price.js';
 import { readSession, writeSession, sessionDir, listSessionIds, type SessionState } from './state.js';
 import type { AgentHost } from '../telemetry/contract.js';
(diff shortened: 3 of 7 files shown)


pavelsem:
Bumped into this on Windows with Czech locale (`cs-CZ`).

In a recent session across 5 retrieval calls, graft actually saved ~67k tokens (with our agent reporting `~24k tokens this turn` in a single turn from two calls). However, `graft stats` reported:
```
tokens saved: ~65
value saved:  ~<$0.01
```

Because `cs-CZ` uses non-breaking space (`\u00a0`) as the thousands separator:
- 4 824 was parsed as 4
- 10 674 was parsed as 10
- 27 086 was parsed as 27
- 5 109 was parsed as 5
- 19 173 was parsed as 19
Total: 4 + 10 + 27 + 5 + 19 = 65 tokens (~1000x under-reported).

Can confirm PR #345 fixes this exact issue. Would love to see this reviewed and merged!

PR #394, opened 2026-09-15
feat: enhance native parser loading and add tests for missing bindings

nsdevaraj:
Fix for native binding is missing for macOS ARM64 / Node 26.6.0.


Frankie-Xu (commented):
Lazy-loading native grammars so a missing `tree-sitter-kotlin` does not take down `graft --version` or a TypeScript-only build is the right direction, and the dlopen probe tests are useful.

The product question for a *required* missing binding is already answered on Ready #214 / #119: fall back to the WASM grammar rather than throw. This PR's `extractFile(Main.kt, …)` path still throws `No native build was found`, and the README documents that as intentional. That is the opposite of #214 — on Windows without MSVC, or Node 26 without a prebuild, Kotlin (and PHP, Swift, …) would still fail the whole file instead of using `tree-sitter-wasm`.

`src/graph/extract.ts` is also the file #214 rewrites for the WASM fallback and Swift completion. I would not layer another native-loading pass onto #214 unless a maintainer wants these two merged. If this lands first, the remaining work is still `throw` → WASM, not more native `require` plumbing.

Not opening a parallel extract.ts PR.

PR #396, opened 2026-09-15
fix(ingest): treat a cross-drive --dir as outside the repo (#384)

### LINKED ISSUE #384: `--dir` on a different Windows drive defeats the outside-the-repo guard: absolute paths accumulate in `.gitignore` and `.ignore`
## Summary

`ensureGitignored` already guards against a `--dir` outside the repo:

```js
const rel = relPosix(root, contextDir);
if (rel === "" || rel.startsWith(".."))
    return; // dir is at/above the repo root — nothing sane to ignore
```

and the doc comment states the intent exactly: *"No-ops when … the dir lives outside `root`
(a custom `--dir` elsewhere, which can't be expressed as a repo-relative ignore)."*

On Windows that guard does not hold when the context dir is on a **different drive**.
`path.win32.relative()` returns an absolute path — not a `../..` chain — when the two paths
have different drive letters, so nothing starts with `..`, the guard passes, and the drive
letter is written into the repo's `.gitignore` with a leading slash:

```
# graft's local graph cache — regenerable, not committed (run `graft build`).
/F:/graft-dir-repro-tmp/
```

The assumption is stated in the comment three lines below the guard — *"`rel` is always
repo-relative here, so a leading `/` is always safe"* — and it is the thing that isn't true
here.

Same-drive is handled correctly, which is why this is narrow:

| root | `--dir` | `relPosix` | guard skips? |
|---|---|---|---|
| `F:/AI/repo` | `F:/elsewhere/scratch` | `../../elsewhere/scratch` | ✅ yes |
| `C:/…/repo` | `F:/graft-dir-repro-tmp` | `F:/graft-dir-repro-tmp` | ❌ **no** |

`ensureSearchable` has the identical guard and the identical outcome in `.ignore`.

## Reproduction

graft 0.16.0, Node v22.14.0, Windows 11. `dist/context/node-file.js` is **byte-identical in
0.18.0**, so this is current.

```console
$ cd C:/…/dirrepro && git init -q . && echo "node_modules/" > .gitignore
$ graft build . --dir 'F:/graft-dir-repro-tmp'
✓ wiring: 2 nodes (1 file, 1 function), 1 edges, 1 cards [javascript]
  → F:/graft-dir-repro-tmp
  F:\graft-dir-repro-tmp/ is git-ignored (added automatically) — a local cache; …
```

`.gitignore` afterwards:

```
node_modules/

# graft's local graph cache — regenerable, not committed (run `graft build`).
/F:/graft-dir-repro-tmp/
```

`.ignore` afterwards:

```
# graft's cards are gitignored but should stay greppable: …
!F:/graft-dir-repro-tmp/
F:/graft-dir-repro-tmp/.cache/
F:/graft-dir-repro-tmp/.graph/
```

Three consequences:

1. **The entries are inert.** They cannot match anything, because the path they name is not in
   the repository at all:
   ```console
   $ git check-ignore -v F:/graft-dir-repro-tmp
   fatal: 'F:/graft-dir-repro-tmp' is outside repository
   ```
2. **They accumulate.** The `present` check matches the exact entry text, so re-running with the
   same `--dir` is correctly idempotent — but each *distinct* `--dir` appends another comment +
   entry block to both files. Two runs against two scratch dirs leave four dead entries.
3. **The build dirties a clean tree**, which is the part likely to surprise someone in CI or a
   pre-commit hook: `.gitignore` shows as modified and `.ignore` as a new untracked file, purely
   as a side effect of pointing `--dir` at scratch space on another drive.

## Suggested fix

Treat "absolute after `relative()`" as the same case the `..` check is already catching:

```js
if (rel === "" || rel.startsWith("..") || isAbsolute(rel))
    return;
```

Comparing the resolved roots (or `parse(root).root !== parse(contextDir).root`) would work
equally well and reads more explicitly as "different volume". Either way both
`ensureGitignored` and `ensureSearchable` need it, since they share the guard verbatim.

## Related, same `--dir` handling

While in here: one file is still written to the **default** context dir even when `--dir` points
elsewhere. After a build with `--dir F:/…`, the repo on `C:` contains:

```
graft/.cache/telemetry-repo-id.json
```

Reproduced by deleting `graft/` and re-running with `--dir` — it comes back. It is one small
file, but it lands as an untracked `graft/` directory inside the repo that the `--dir` ignore
entry does not cover (broken or not), so it compounds the dirty-tree effect above. Happy to
split this out if you'd rather keep the issues one-to-one.

Frankie-Xu:
## Summary
- `ensureGitignored` / `ensureSearchable` treated `rel.startsWith(..)` as `--dir` outside the repo. On Windows, `path.win32.relative` for a different drive returns an **absolute** path (`F:/tmp`), so the guard missed and `/F:/graft-dir-repro-tmp/` was appended to the repo `.gitignore` / `.ignore`.
- Treat an absolute `rel` (`path.isAbsolute` or `path.win32.isAbsolute`) as outside too. Same-drive in-repo `--dir` still writes `/graft/`.

## Not in this PR
- Does not change `GRAFT_NO_GITIGNORE` / `GRAFT_NO_IGNORE`.
- Does not index a cross-drive directory.
- Does not move `graft/.cache/telemetry-repo-id.json` (split out in the issue).

Closes #384

## Test plan
- [x] `root` on this volume + `F:/graft-dir-repro-tmp` → neither ignore file contains `F:`
- [x] Same-drive `<root>/graft` still gets `/graft/`
- [x] `node --import tsx --test --test-name-pattern ensureGitignored|ensureSearchable test/context.test.ts`


Frankie-Xu:
Follow-up: 919d38e stores node-file.ts and the test as LF to match main. The previous commit was CRLF-only noise. Behaviour is unchanged.

PR #436, opened 2026-09-21
fix(crux): recover the target id when a model echoes the whole row

UnkyBadTouch:
## Problem

The crux (meaning) pass matches each returned summary back to its node by **exact id** (`enrichGraph` → `results.get(node.id)`). The user turn labels targets as:

```
- id=<id> | <kind> | lines L<a>-L<b> | <signature>
```

Some models don't copy `<id>` verbatim — they echo the whole row into the `id` field. Confirmed with `deepseek-chat` (and `deepseek-reasoner`), on **both** DeepSeek's OpenAI- and Anthropic-compatible endpoints:

```
expected:  src/ChargeSizer.php#ChargeSizer.plan
returned:  src/ChargeSizer.php#ChargeSizer.plan | method | lines L49-L69 | public function plan(...)
```

Every summary then misses its node, so `enrichGraph` applies 0 for the file and reports the misleading quality miss:

```
✗ src/ChargeSizer.php: model returned no usable symbol summaries [empty-parsed, finish_reason=null]
```

(`finish_reason=null` because `summarizer.lastMiss` is actually `null` — the model answered fine; the fallback label is used when entries exist but nothing matched.) On a 233-file PHP repo this left meaning coverage at 34%.

## Fix

Node ids never contain ` | `, so `normalizeId` trims the echoed row back to its first segment. A faithfully-copied id is returned unchanged.

## Verification

Reproduced against the engine directly with a real `ChatCruxSummarizer` + `enrichGraph` on the failing files:

- before: `4 errors / 34 computed`
- after: `0 errors / 215 computed` (including a 101-symbol file)

Added a regression test that feeds a row-echoed id through `ChatCruxSummarizer.describeFile`. `test/llm-ops.test.ts` and the `graph-enrich-*` / `cli-deep-failure` suites pass.


UnkyBadTouch:
An odd bug found, tested against opencode go and DeepSeek API

